PR #3, opened 2015-06-06
fix issue with missing folders

jbinard:


Pierre-Gilles:
I agree with the signin link in the homepage, but maybe the two button "login" and "register" can be at the same line ? I think it can be more beautiful.

Otherwise thanks for the PR !

screenshot

PR #5, opened 2015-06-08
Fix MusicService when soundCapable is disabled.

Spope:
Hi, I started an instance of Gladys with soundCapable to false, and MusicService create an instance of node-dir only if soundCapable is true, but the this instance is used without any check (L-180). I have pasted the same check, you can also check for the dir variable.

### DIFF
--- a/api/services/MusicService.js
+++ b/api/services/MusicService.js
@@ -176,52 +176,54 @@ module.exports = {
 				callback(null);
 		}
 
-		// get recursively all the files in the folder
-		dir.files(sails.config.music.folder, function(err, files) {
-			if(err) return sails.log.warn(err);
-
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
You're right that's not normal ! 

Otherwise, another way to solve the problem in fact could be put the node-dir instance out of the first if statement.
Because in fact reading music file does not require to be "soundCapable", and if for example we want to stream music right into the browser of the user, even if the server is not capable of playing sound,  he can import music from a directory.

So I would prefer to just simply put `var dir = require('node-dir');` out of the first if.

Thanks for the pull request ;)

Spope:
Great, it was a quick fix, I was eager to test Gladys!

PR #24, opened 2015-06-23
pm2 / rename app to gladys

VonOx:
Yes i know this is a detail but this is her name no ? :smiley_cat:


Pierre-Gilles:
You're right, and that's not a detail, it's important :) Thx !

PR #29, opened 2015-07-01
Adding links to Github and Gladys Website

VonOx:
They are in the header beside the user menu

Screenshot

### DIFF
--- a/config/locales/en.json
+++ b/config/locales/en.json
@@ -19,5 +19,7 @@
   "pagename-sensor":"Sensors",
   "pagename-alarm":"Alarm",
   "pagename-me":"Me",
-  "pagename-music":"Music"
+  "pagename-music":"Music",
+  "GithubLinkTitle":"View on Github",
+  "GladysWebsiteTitle":"Go to Gladys website"
 }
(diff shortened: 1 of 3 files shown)


VonOx:
Maybe translation is not necessary ?
Or if i understand the mechanics ==>  "View on github":"Voir sur Github" is good enough for fr and en ?

Pierre-Gilles:
I don't know if putting github and website links in the navbar is a good ideas, it's maybe too visible !

I would prefer like in the footer or something like that.

For the translation that's perfect, you have understood :) The way you've done it is the best practice. 

I know sometimes I just write  "English-text":"french-text" and nothing in the en file, but that's not a good practice, I need to change that

VonOx:
Yes, i agree
The problem, this version of admin Lte have no footer / Need update or customize.

Pierre-Gilles:
Ok I think updating the theme can be the best solution, but it will take some time. I'll make some tests at home!

VonOx:
I have started the update yersteday.

The blue light skin have white sidebar

VonOx:
Can you close this PR ?

PR #31, opened 2015-07-09
Fix bug callback undefined

lTimeLordl:


Pierre-Gilles:
Thx ! Ideally, all functions of the Gladys API should begin with callback = callback || function(){};, but sometimes it's not the case. The API should never crash because of these little mistakes, so if you find another functions in "api/services" with this problem, don't hesitate.

Another problem : 
Callback response should always begin with an error param  : 

```
function test(a, callback){
    ...
    callback(null, 'ok');
} 

test(a, function(err, result){
    if(err) {
        console.log(err);
    }else{

    }
}
```

But sometimes it's not the case, we should correct that everywhere.

PR #35, opened 2015-08-21
Add angular-animate and prod to pm2

lTimeLordl:


Pierre-Gilles:
I dont think adding prod mode to the file is a good idea ! Because in fact you always need to start Gladys in dev mode one time because sails needs to build the database ( in prod mode, the database is in alter:safe mode and cannot alter the schema )

That make me think that it's the same for modules.. when you install a module, you need to start Gladys again in dev mode so it can create the tables 

For ng-animate, you need it ?

lTimeLordl:
I thought it was a good idea to start with prod, thanks for your details

For ng-animate yes, I don't like like Jquery and you say "Arriver à me séparer de certaines dépendances côté client ( Je pense à JQuery par exemple, que je suis obligé de mettre pour Bootstrap, MorrisJS, simpleWeather et datepicker ). J'aimerais pouvoir passer à une app full AngularJS bien plus légère, car actuellement côté client c'est trop lourd."

https://daneden.github.io/animate.css/

Pierre-Gilles:
Yeah of course i want to remove jquery ! 
I think we should make a list of all dependencies in Gladys, and their equivalent in angularJS ( maybe on the forum )

For bootstrap for example, I think next version of bootstrap ( 4) will come with a brand new JS client ( and I hope without jquery )

lTimeLordl:
Ok, so wait bootstrap 4 :)

Pierre-Gilles:
I read the code on bootstrap repository, seems that they are still using jquery... We'll find an alternative ;)

PR #41, opened 2015-10-24
Add npm as a prerequisite, and apt-get commands

wikijm:
- Add npm as a prerequisite ( - [npm](https://www.npmjs.com/) ( `sudo apt-get install npm` ) )
- Add "apt-get commands" for MySQL and npm ( ( `sudo apt-get install mysql-server` ) and ( `sudo apt-get install npm`) )

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ Prerequisites
 -------------
 
 - [Node.js](http://nodejs.org) ( = v.0.10.xx, some troubles appears with node.js 0.12.x )
-- [MySQL](http://www.mysql.com/)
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
Npm comes with Node.js, it's not the proper way to install it !

For MySQL, it depends of your system, if you are on Windows, or Mac, or any Linux without apt, it's not the same.. This part of the README is "universal", you already have an installation script ( https://github.com/GladysProject/Gladys/blob/master/rpi-install.sh ) for a Raspberry Pi on Debian :) )

Sorry for the late reply, I was really busy and forgot your post !

PR #43, opened 2015-11-17
Update dependencies

VonOx:
- MD5 replaced by md5 ( Deprecated )
- bcrypt replaced by bcryptjs ( plain javascript / no dependencies )

Need npm prune & npm install after update and of course need test


Pierre-Gilles:
Thanks for the PR, it works fine !

PR #45, opened 2015-11-30
Add queue to say sentence per sentence

lTimeLordl:
Hi,

Add queue with async
Add a optional callback to the say method because a few methods of the api have it.


Pierre-Gilles:
Thx for the PR, speaking is no more in Gladys core, but in a module. Will integrate the queue in the module :)

PR #48, opened 2016-01-09
Nevermind - Fixed some issues with PM2 on reboot

Molkobain:
I had 2 issues with PM2 :
- The script generated by pm2 startup was using the wrong user (root instead of pi)
- The dump file containing the applications to ressurect was set to empty by PM2 when rebooting, therefore the applications where not starting automatically when rebooting manually

So I made 2 changes to rpi-2-install.sh script.

This might need to be tested on distro others than raspbian.

### DIFF
--- a/api/controllers/LifeEventController.js
+++ b/api/controllers/LifeEventController.js
@@ -82,9 +82,9 @@ function getIntervalBetweenLifeEvents(userId,startEventType, stopEventType, call
   						var duration = (date1.getTime() - date2.getTime())/ (1000*60*60);
               // Round in the format :  8,2 Hours (for example) 
               duration = Math.round(duration*10)/10;
-              var month = date2.getMonth() + 1;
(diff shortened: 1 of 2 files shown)


Pierre-Gilles:
Thanks for the PR ! I'll look at it this week.

Pierre-Gilles:
Be careful, don't mix PR ! If you want to correct minor bugfix, correct in another PR :)

Molkobain:
I konw, but when I clicked on "Create a PR" on my repo, it asked for a title (that was different than this one) and then when i submitted it was merged with the first PR. Don't know why. Is it because the previous was still open ?

Molkobain:
I get it, I think I should have make a branch for each PR. Can't undo it unfortunately.

Molkobain:
PR to delete as I did another one cleaner. I can't delete the PR on my side.

PR #51, opened 2016-01-17
Fixed some issues with PM2 on reboot

Molkobain:
I had 2 issues with PM2 :

The script generated by pm2 startup was using the wrong user (root instead of pi)
The dump file containing the applications to ressurect was set to empty by PM2 when rebooting, therefore the applications where not starting automatically when rebooting manually
So I made 2 changes to rpi-2-install.sh script.

This might need to be tested on distro others than raspbian.

### DIFF
--- a/rpi-2-install.sh
+++ b/rpi-2-install.sh
@@ -44,6 +44,7 @@ sudo npm install -g pm2
 sudo npm install -g gladys --unsafe-perm
 
 # Starting Gladys at startup
-sudo pm2 startup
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
You're right :) Cannot test it tonight on my pi, will look at it when possible ! Thanks for the PR.

VonOx:
Ok on ubuntu

Pierre-Gilles:
Closing. It works in Gladys v3 :)

PR #52, opened 2016-01-17
Sleep graph date fix

Molkobain:
Fixed the date used to calculated the sleep time. It was using the goingToSleep date for the graph which caused some flaws when goingToSleep twice in a day (eg. at 0:30AM and 11:00PM). By using the wakeUp date we fix this.


aschelch:
+1

Pierre-Gilles:
Thx for the PR, this is different in Glady 3.0, so PR is no longer valid. Closing PR !

PR #53, opened 2016-01-17
Localisation - Sleep graph

Molkobain:
Added some translations about the sleep graph to the french localisation file.


Pierre-Gilles:
Works nicely, I forgot to add the translation for this, thanks for the PR :)

PR #54, opened 2016-01-17
Fixed GoogleCalendar service that was crashing when an event had no s…

Molkobain:
…tart or end date (eg. cancelled events)


Pierre-Gilles:
I simplified your code, with an `if else` statement. Looks like this now : 

``` javascript
if(Event.start){
        if(Event.start.dateTime){
            Event.start = Event.start.dateTime;
        }else if(Event.start.date){
            Event.start = Event.start.date;
        }
    } else {
        return callback('saveorUpdateEvent : No start date for the event.');
    }
```

Molkobain:
Yes, my bad. I started with a !Event.start for debugging purpose and did not paid attention when finalizing it. Thanks :)

PR #61, opened 2016-06-21
Add a route to get Device By Service

Jean-PhilippeD:


Pierre-Gilles:
Same here, sorry for not checking, I've added this in the list of issues, I'll fix it => https://github.com/GladysProject/Gladys/issues/281

PR #92, opened 2016-09-25
Add devicetype.getById and getByType

Jean-PhilippeD:
As I needed those functions :)

### DIFF
--- a/api/core/devicetype/deviceType.getById.js
+++ b/api/core/devicetype/deviceType.getById.js
@@ -1,0 +1,18 @@
+var queries = require('./deviceType.queries.js');
+var Promise = require('bluebird');
+
+module.exports = function(id){
+    
+    // get state for a given devicetype id
+    return gladys.utils.sql(queries.getById, [id])
+      .then(function(deviceType){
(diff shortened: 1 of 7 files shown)


Pierre-Gilles:
I don't know are unit tests are passing by your side, in your validator file you have a function : 

``` javascript
function validateDeviceType(devicetype) {
    device.should.be.instanceOf(Object);
    device.should.have.property('id');
    device.should.have.property('type');
    device.should.have.property('tag');
    device.should.have.property('sensor');
    device.should.have.property('device');
    device.should.have.property('min');
    device.should.have.property('max');
}
```

device does not exist, it's devicetype here ! 

To start the tests, you can run 

```
grunt test
```

On your machine

You can replace  `it('should return a deviceType', function (done) {`

by `it.only('should return a deviceType', function (done) {` when you are working on it, so that only this test is executed.

Btw, try to separate as much as possible your PR, so that I can merge what's working, and comment what's not working :)

Jean-PhilippeD:
I gonna fix the validator which is wrong

PR #94, opened 2016-09-25
Add devicetype.getById and getByType

Jean-PhilippeD:
I followe what you've said. :)


Pierre-Gilles:
Hi !
I just checked your PR, I've done a lot of changes before merging it. 

Here is my review : 
- Your deviceType.getById SQL was not right, you can't take the last deviceState value only by ordering the deviceState table by ID. We can have offline devices synchronizing their value only by going online a few seconds each day, so when they sync the values of the days are inserted in a random order and we can't trust the ID to give us the last value. You must rely on the datetime which is client fixed, and after that rely on the ID ( so it's a double MAX which is done to ensure unicity )
- There were still a lots of comments not related to the current file ( please be careful in the future )
- There were files modified just by adding a line or a space, please be careful on that too
- Don't require things which are not used in the current file ( ex : You were requiring bluebird 2 times without using it ). 

I corrected all that, and it's going to be live in a few minutes :)

Jean-PhilippeD:
Hello,

Thanks for the explanation !
Next time, I won't Pull my request late in the day ... and modify the files only in github instead of pushing it from my server where I run my "prod version" ..

PR #101, opened 2016-10-07
Fixed two typos in en.json

danships:
I noticed two typos in the Gladys control panel.


Pierre-Gilles:
Nice catch ! Thank you very much for the PR :)

PR #112, opened 2016-11-08
Added gender - Other

ruchirx:
Added a new Gender Type: Other


ruchirx:
I just think it will be cool to have inclusivity. It will be worth it even if one user will be happy with this change! :)

Pierre-Gilles:
I'm not against this kind of changes, but we need to look way forward than just adding an option in a select ^^ 

Why am I asking if the user is male or female ? Where is it used in Gladys ? It's to allow Gladys to put "Mr" or "Miss" when Gladys speak, these kind of things.. So what happens if we put other ? There are a lot of particular cases to think about... 

I just checked on big sites like Facebook, even Facebook does not have the options, it's just male or female ! 
So I really don't know as a small project if we have the ressources to handle theses cases everywhere... This asks a lots of philosophical questions, and maybe at this stage of the project, that's maybe too much and time consuming... 

And I don't think that's the role of a small project like Gladys to ask itself these kind of questions, when even countries are not capable of making a decision on this...

ruchirx:
Facebook has 64 gender options. You can go to settings in your account on Facebook and see.
Almost all companies have >2 gender options. 

You can always use First Name of the user while interacting. Don't need Mr or Ms Last Name! 

But it's your call. @Pierre-Gilles

PR #117, opened 2016-11-21
Add functions to test if it's night or not in a given house

Jean-PhilippeD:
Can be used by user in script or in scenario with gladys-data

### DIFF
--- a/api/core/sun/index.js
+++ b/api/core/sun/index.js
@@ -1,2 +1,3 @@
 
 module.exports.init = require('./sun.init.js');
+module.exports.isNight = require('./sun.isNight.js');
--- a/api/core/sun/sun.isNight.js
+++ b/api/core/sun/sun.isNight.js
@@ -3,7 +3,7 @@ module.exports = isNight;
 var SunCalc = require('suncalc');
 var Promise = require('bluebird');
 
-function isNight(data){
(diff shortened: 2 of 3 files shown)


Pierre-Gilles (changes requested):
Hi ! Thanks for your PR ! 
Just a feedback : getting all houses and looping threw them to find the right one, without even stopping the loop when you found the one, is clearly an inefficient way of doing this ^^ 

Instead, we should get the house by id to have the latitude & longitude, and then make the calculation, but looping here is inefficient ^^ I'll create for you a house.getById function right now.   

Second feedback, you are calculating two times the Date in your sunState function. That's not bad, but why not create the `now` variable once and use it at both place ? That could prevent weird bugs like if the calculation takes place at 00:00 and one get calculated before and the other after.

Theses a just feedbacks to make this great, thanks for your PR !

Pierre-Gilles:
house.getById done :)

Jean-PhilippeD:
ok, I was thinking about a house.getById which would be better.. thanks for this :)

I will check for the date calculation..

I will push a new PR once those things will be fixed. 

Moreover, I think I will write a getSunState and then a isItNight and isItDay so we will have 2 states for scenario instead of just testing if it's night.

PR #128, opened 2016-12-09
Add url to get last device value by devicetypeid

jeremy-rafflin:
Exemple : To get the last value and informations of the DeviceType  5 you can do : 
Get http://192.168.0.88/devicetype/devicetypeStates/5

### DIFF
--- a/api/controllers/DeviceTypeController.js
+++ b/api/controllers/DeviceTypeController.js
@@ -53,6 +53,14 @@ module.exports = {
         .catch(next);
   },
 
+   getById: function(req, res, next){
+      gladys.deviceType.getById({id: req.params.id})
+        .then(function(DeviceType){
+           return res.json(DeviceType); 
+        })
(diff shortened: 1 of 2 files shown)


Pierre-Gilles:
Hi ! 

Thanks for your PR :) 

I have a few feedbacks : 

First the route you created to get a deviceType is not really RESTful as the rest of Gladys, we will simply use : 

`GET /devicetype/:id`

Second, as each PR you should unit tests each functions/routes you write so that the test runner can verify that your new feature is running well :)

I'm gonna implement quickly all this so you have your route, it will be quicker and cleaner that merging+correcting everything ! 

But thanks for proposing this PR ;)

Pierre-Gilles:
On master commit =>  https://github.com/GladysProject/Gladys/commit/8c15aa4f18a2f3d11b4ea93b2b3d1c10ad2266a9

PR #130, opened 2016-12-14
Proposal for allowing users to set their own env vars directly in the project.

Magellol:
Hi everyone.
I actually just installed Gladys today to play with it and realized I could not get it running without modifying `config/connections.js` directly because I could not provide the require env variables outside of the project scope. Which is not good because that file will get overridden on each updates.

Instead here's a _proposal_ to set them up on a per-project scope. Means, everyone can create their own `.env` file at the root and set the env vars they need to work with their setup. Now, I realize it's kind of a breaking change because some people might have relied on the fact that we default to some values if the env vars are not there (we can bring that behaviour back if necessary) but I wanted to introduce some good practice first and get your opinions after since I'm not experienced with the project.

A `.env` file looks like:

```
MYSQL_HOST=localhost
MYSQL_PORT=3306
MYSQL_USER=root
MYSQL_PASSWORD=''
MYSQL_DATABASE=gladys

MYSQL_HOST_TEST=localhost
MYSQL_PORT_TEST=3306
MYSQL_USER_TEST=root
MYSQL_PASSWORD_TEST=''
MYSQL_DATABASE_TEST=gladystest
```

I can update the Readme file if we're going to agree on these changes. 👍


Pierre-Gilles:
First thanks for your proposal ! I'm 100% ok with the idea of finding another solution than actual way in Gladys. 

>Which is not good because that file will get overridden on each updates. 

Yes of course modifying `connection.js` is really dirty :) 

Your proposal seems great, the only problem I see is that it will breaks all current install if we remove the fallback as the current gladys raspbian image uses the fallback.. As you say for now we should keep the fallback.

But I agree on the changes, just put the fallback back and you can update the README ! 👍 

Again, thanks for this PR !

Magellol:
Sweet! One thing I should have mentioned is I'm wrapping the dotenv call into a `development` NODE_ENV ([see](https://github.com/GladysProject/Gladys/pull/130/files#diff-0364f57fbff2fabbe941ed20c328ef1aR39)) check because they should only be required in development mode. Other env should set these env variables directly. Do you see any problem with it? My understanding is if we put the fallback back in, it's backward compatible as people not having a .env would still be able to run Gladys as they were.

Will go around these changes today 👍 

EDIT: Changes are in. I took the chance to add a `.env-sample` file to list which variables are there to be customized.

Pierre-Gilles:
Thanks for your revision ! Just one feedback, we should add the .env file to the .npmignore too :) 

But that's just a detail, the rest is great !

Magellol:
Just added it in :)

Pierre-Gilles:
Thanks ! Merged :)

PR #132, opened 2016-12-18
User patch

Jean-PhilippeD:
Add features to forward either event's user or session's user to Script context

### DIFF
--- a/api/core/script/script.exec.js
+++ b/api/core/script/script.exec.js
@@ -5,10 +5,11 @@ module.exports = function(options) {
 
     // handle both direct call and scenario call
     var id = options.id || options.params.id;
+    if (options.scope && options.scope.user) var user = options.scope.user;
 
     return gladys.script.getById({id})
         .then(function(script) {
-            return execCode(script.text, options.scope.user);
+            return execCode(script.text, user);
         });
 };
 
--- a/api/core/script/script.exec.js
+++ b/api/core/script/script.exec.js
@@ -8,14 +8,15 @@ module.exports = function(options) {
 
     return gladys.script.getById({id})
         .then(function(script) {
-            return execCode(script.text);
+            return execCode(script.text, options.scope.user);
         });
 };
 
 // execute the code
-function execCode(code) {
+function execCode(code, user) {
     return new Promise(function(resolve, reject) {
         try {
+            shared.sandbox.user = user;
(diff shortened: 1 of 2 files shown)


api/core/script/script.exec.js
```diff
@@ -5,17 +5,19 @@ module.exports = function(options) {
 
     // handle both direct call and scenario call
     var id = options.id || options.params.id;
+    if (options.scope && options.scope.user) var user = options.scope.user;
```
Pierre-Gilles, line 8: You should create the user variable outside of the if scope, it's not really clear here where the user is defined

api/core/script/script.exec.js
```diff
@@ -5,17 +5,19 @@ module.exports = function(options) {
...
+            return execCode(script.text, user);
         });
 };
 
 // execute the code
-function execCode(code) {
+function execCode(code, user) {
     return new Promise(function(resolve, reject) {
         try {
+            shared.sandbox.user = user;
```
Pierre-Gilles, line 20: You cannot do that! You are modifying a global variable which is the same for all call to this function. So if two scripts are executed at the same time = problem !

Jean-PhilippeD:
Ok I understand, do you have an idea to export user (or maybe other variable when calling the script through rest api ?) to the script ?

Pierre-Gilles:
I think one good practice should be to pass to the sandbox a clone of all variables... I have an idea I will try to implement it ! 

I created an issue, it's referenced in the milestone for Gladys 3.4 which I'm trying to release this Sunday :)

Jean-PhilippeD:
Hi,
In v 3.4 you've changed the way you declare the user regarding this PR.
It works fine when you pass from Controller, but when you come from a Scenario, you don't push user in the _options_ variable in the _core.script.exec.js_ so _context.user_ is not declared inside script.

In _script.exec.js_, don't you should add  ?
   ` if(!options.user && options.scope.user) options.user = options.scope.user;`
just after line 8.

In my PR, I was doing : 
`if (options.scope && options.scope.user) var user = options.scope.user;`
but it was not clear as you said.

Pierre-Gilles:
Fixed here => https://github.com/GladysProject/Gladys/commit/e8428b11a743404041e269a756bfeae0fc7f2322 :)

PR #133, opened 2016-12-18
pass user as param to verify condition

Jean-PhilippeD:
So that you can use user id when verifying conditions


Pierre-Gilles:
Sorry, this PR is so old it can't be merged.. Sorry for not getting a look at this

I understand the need so I added an issue here => https://github.com/GladysProject/Gladys/issues/280

I'll fix it myself :)

PR #138, opened 2017-01-27
add purge function on devicestate

Jean-PhilippeD:
I propose to add a "purge" on devicestate using the delete from mysql statement, based on devicetype id and number of days from now you want to keep.

I've added test also, you can check the test and tell me if it is enough.

(Maybe we could think about historize data in a histo table .. ?)

### DIFF
--- a/test/unit/api/core/devicestate/deviceState.purge.test.js
+++ b/test/unit/api/core/devicestate/deviceState.purge.test.js
@@ -1,0 +1,46 @@
+var should = require('should');
+var validateDeviceState = require('../../validator/deviceStateValidator.js');
+
+describe('DeviceState', function() {
+
+  describe('purge', function() {
+    
+    it('should return 2 devicestates only', function (done) {
+     	
+        var state = {
+            devicetype : 1,
+            value: 12.1,
+            datetime: '2010-01-01 22:00:00'
+        };
+        
+        gladys.deviceState.createByIdentifier('THIS_IS_MY_IDENTIFIER', 'test', 'binary', state)
+        .then(function(result){
+           options = {
+             devicetype : result.devicetype,
+             days : 365
+           }
+
+           return gladys.deviceState.purge(options);
+        })
+        .then(function() {
+           states = {
+              devicetype : 1
+           }
+
+           return gladys.deviceState.get(states);
+        })
+        .then(function(result){
+           validateDeviceState(result);
+           result.should.be.instanceof(Array);
+           result.should.have.length(2);
+           done();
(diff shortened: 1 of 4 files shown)


test/unit/api/core/devicestate/deviceState.purge.test.js
```diff
@@ -0,0 +1,46 @@
...
+     	
+        var state = {
+            devicetype : 1,
+            value: 12.1,
+            datetime: '2010-01-01 22:00:00'
+        };
+        
+        gladys.deviceState.createByIdentifier('THIS_IS_MY_IDENTIFIER', 'test', 'binary', state)
+        .then(function(result){
+           options = {
```
Pierre-Gilles: Be careful, try to use always "var" when you create a variable ! Otherwise, it's kind of global, and it's a bad thing ^^

test/unit/api/core/devicestate/deviceState.purge.test.js
```diff
@@ -0,0 +1,46 @@
...
+             devicetype : result.devicetype,
+             days : 365
+           }
+
+           return gladys.deviceState.purge(options);
+        })
+        .then(function() {
+           states = {
```
Pierre-Gilles: Same here, use "var" when you create a variable. 

For simple tests like this you can maybe just do : 

```gladys.deviceState.get({devicetype: 1});```

So it's easier to read.

File: test/unit/api/core/devicestate/deviceState.purge.test.js [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+           }
+
+           return gladys.deviceState.get(states);
+        })
+        .then(function(result){
+           validateDeviceState(result);
+           result.should.be.instanceof(Array);
+           result.should.have.length(2);
+           done();
+        }).catch(function(err){
```
Pierre-Gilles: I know I've not always done that in all tests of Gladys, but you can simply do : 

`.catch(done):`

Pierre-Gilles (changes requested):
Nice PR ! Thanks for this function, this was really needed. 

I underlined a few points, theses are just best coding practice :)

Waiting for you changes and merging this

Jean-PhilippeD:
I've made the change

Jean-PhilippeD:
I've just tested in on my teleinfo module where I had about 2 millions rows since the beginning of december and I successfully purge data at 15 days :) I'm back at 245k rows.

Pierre-Gilles:
I'm merging this :)

PR #238, opened 2017-10-08
Add Scripts editor auto completion

C4rlit0:
PR 1 - Add dependencie and set option for auto completion in editor from scripts view
PR 2 - Add tooltip on input field in Params view


Pierre-Gilles:
Hello !

Thx for the PR, just one feedback, when you add several feature, try to do one PR per feature, otherwise is really hard to merge one by one the features for me :) 

- About the first one, how does the autocomplete works with Gladys API ? We want to autocomplete Gladys function ( gladys.house.create, gladys. .... ), having autocomplete with JS browser function is not really useful ^^ ( and won't help newcomers, as it will show them things that doesn't work in a Gladys script! 

- About the second one, as Gladys is international (working in FR, EN and soon DE ), don't hardcode string in a specific language. We have language config files => https://github.com/GladysProject/Gladys/blob/master/config/locales/en.json that can be used like this => https://github.com/GladysProject/Gladys/blob/master/views/device/index.ejs#L22

Thx again for the work !

Pierre-Gilles:
Hey ! As I'm checking all PR, what's the plan on this one ? :) 

( Don't worry, it's not urgent, if you want to focus on the UI rewrite I'm 100% in ! )

Just tell me what to do with this one ;)

C4rlit0:
I'll correct it soon, so sorry for the delay 😅

Pierre-Gilles:
Can I close this PR? 😂

PR #240, opened 2017-10-10
Update index.js

isokar:
add createValue.js


Pierre-Gilles:
I think you forgot to commit one file... :)

Pierre-Gilles:
Oh I see you did 2 PR, why ? Test are not passing if you separate each file :) The other PR is not complete as well, your code miss one else statement and I'm not totally sure the function name is explicit. It can be "createValueIfNotExist" for example. 

Btw, try to write unit-test when you write a feature in Gladys.

Example : https://github.com/GladysProject/Gladys/blob/master/test/unit/api/core/param/param.setValue.test.js

PR #247, opened 2017-11-12
Change the OnOff buton by the switch

MathieuAndrade:


Pierre-Gilles:
Really nice work ! Merging this right now :) 

Just one feedback about CSS files, you should not add the CSS file path to the layout.ejs file. You should add them to the "tasks/pipeline.js" file, because CSS files are concatenated + minified during build process ! And the resulting CSS file is injected in the layout.

PR #257, opened 2017-11-26
No change name the device

pjap93:
Update the device while not changing the name.


Pierre-Gilles:
Thx for this nice PR ! Merging this :)

PR #265, opened 2017-12-23
event purge

Jean-PhilippeD:
add event purge based on eventType and value


Jean-PhilippeD:
wrong syntax on event.json, gonna fix it

PR #271, opened 2017-12-28
event purge

Jean-PhilippeD:
add event.purge function based on value and eventType.


Boimb:
Really nice feature !
Thx

PR #306, opened 2018-03-05
Fix the height of ul menu

MathieuAndrade:
Cut the contents of the ul menu when there is too much devices


Pierre-Gilles:
Nice catch !

PR #314, opened 2018-03-19
Fix display param when  deviceType was updated

MathieuAndrade:


Pierre-Gilles:
Nice!

I fixed some missing semicolon + added the "tag" attribute too.

PR #348, opened 2018-05-01
Updating required node versions

jibaku:


Pierre-Gilles:
Thanks for this PR! Btw, Gladys should be working with Node 10 soon, this should not be hard

PR #349, opened 2018-05-10
Function is it right time

pjap93:
function isItRightTime.

Valide if the time is in the range definided.
For the addition in the conditions of the scenarios.


Pierre-Gilles:
Hey! Is it ready for merge ? :) What's your use case on this ?

pjap93:
Hi,
it is close if there is no error, but it is impossible to understand what is causing problems with Travis lol, even by putting the test file on it.
When I read the details of the error it indicates that the function does not exist while I declare everything (well I think).

I currently use it under scenario conditions to set a validation time range.

I would then make a PR in the data to add the condition if this is ok for you.

MathieuAndrade:
Use `Promise.resolve(true)` and `Promise.resolve(false)` instead of just returning a boolean ;)

pjap93:
is good, thanks

api/core/time/time.isItRightTime.js
```diff
@@ -0,0 +1,56 @@
...
+    // end time default is the end of the day
+    var endDate = new Date();
+    endDate.setHours(arrayOfEnd[0] || 23);
+    endDate.setMinutes(arrayOfEnd[1] || 59);
+    endDate.setSeconds(arrayOfEnd[2] || 59);
+
+    // settings default options if not defined
+    options = options || {};
+    options.start = date;
+    options.end = endDate;
```
Pierre-Gilles: Are these 3 lines useful ? I don't think so!

api/core/time/time.isItRightTime.js
```diff
@@ -0,0 +1,56 @@
...
+
+module.exports = function isItRightTime(options){
+
+    var now = new Date();
+
+    var arrayOfSart = options.start.split(":");
+    var arrayOfEnd = options.end.split(":");
+
+    // default start time 00:00:00
+    var date = new Date();
```
Pierre-Gilles: Consider renaming the variable to "startDate", more explicit

Pierre-Gilles (changes requested):
Hey ! Thanks for your PR, this is really nice :) 

I'm merging this just after these two changes in this review

api/core/time/time.isItRightTime.js
pjap93: ah yes well seen, I forget to remove them, initially I wanted to return data after I change a true / false

api/core/time/time.isItRightTime.js
pjap93: it's not my part that I prefer to give explicit names to variables lol
but you already noticed

pjap93:
and in the stride I can add in the base the condition for the scenarios !!

api/core/time/time.isItRightTime.js
Pierre-Gilles: Ok! Waiting for your review before merging :)

pjap93: it is done :-)

Pierre-Gilles:
It's now merged! :) 

I just renamed the function "isInTimeRange" because your old function name was clearly not explicit. 

Thanks for your work!

PR #353, opened 2018-05-13
Script correction

pjap93:
limitation of the display only on the 50 removed for getScripts


Pierre-Gilles:
I agree this is a problem, but I'm not a big fan of removing pagination. 

Pagination is here for something: It's to avoid overloading the UI if you have 200 scripts!

There are many options : 
- Replace the select by a dropdown with infinite scrolling (so each time you scroll, you load more)
- Add a script search bar with pagination

PR #376, opened 2018-07-11
🐳 Switch from node:carbon to node:8-alpine for Docker base image and add Timezone feature

VonOx:
Build and Run test with success

Purpose of this modification is to reduce images size

- Size with node:carbon = 979.9 MB
- Size with node:alpine = 375.4 MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,9 @@ RUN mkdir /src
 
 WORKDIR /src
 ADD . /src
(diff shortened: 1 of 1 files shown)


VonOx:
165 MB on docker hub ( was 408MB )

Pierre-Gilles:
Nice! Thanks for your PR :)

PR #384, opened 2018-07-22
:whale: Fix docker build

VonOx:
Solution is to install build deps and remove them when npm install finished

Image size 230Mb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:8-alpine
 
-# Add tzdata for timezone settings
-RUN apk add --no-cache tzdata
+# Add tzdata for timezone settings and build deps
+RUN apk --no-cache add --virtual builds-deps build-base python tzdata
 
 # Create src folder
 RUN mkdir /src
@@ -12,6 +12,9 @@ RUN npm install
 
 RUN npm install -g grunt-cli
 RUN grunt buildProd
+
+# Remove build deps
+RUN apk del builds-deps
 
 # Export listening port
 EXPOSE 8080


Pierre-Gilles:
Thanks for your PR :) 

This won't work! 

Because of the layered structure of Docker, each RUN statement in a Docker file create a new layer in the resulting docker image.

If you add data to a layer (build-deps), and remove it in the next layer, the resulting image won't be lighter, because data is still present in the first layer!

To reduce image size, you need to execute everything and remove the build deps in the same layer : 

```
RUN apk add --no-cache --virtual .build-deps make gcc g++ python git && \
    yarn install --production && npm install -g grunt-cli && grunt buildProd && npm uninstall -g grunt-cli \
    apk del .build-deps
```

I just did the change, here is the resulting Dockerfile : 

```
FROM node:8-alpine

# Add tzdata for timezone settings
RUN apk add --no-cache tzdata

# Create src folder
RUN mkdir /src

WORKDIR /src
ADD . /src

RUN apk add --no-cache --virtual .build-deps make gcc g++ python git && \
    yarn install --production && npm install -g grunt-cli && grunt buildProd && npm uninstall -g grunt-cli \
    apk del .build-deps

# Export listening port
EXPOSE 8080

CMD ["node" ,"app.js"]
```

Btw, I added a .dockerignore file to avoid pushing useless data from local machine (.git, node_modules, etc... )

VonOx:
Yes you are right, my first concern was "image must build" :grin:

Thank you for fix ( Pulling new image this evening )

PR #387, opened 2018-07-26
🔧 Update .travis.yml to manage docker build

VonOx:
So i have managed to build docker image with several tag:

- latest ( on master branch )
- major version ( on release/tag )

Image is pushed to DockerHub

You need to set env variable in Travis and disable automatic build in docker hub:

- DOCKER_PASSWORD
- DOCKER_USERNAME

⚠️ Complete build need to be tested ( with documentation )

Performance: added yarn cache to speed up installation of deps


VonOx:
Automatic Build need to be disabled in DockerHub

Pierre-Gilles:
Hey! 

Thanks for your PR :)

Why do you have to repeat your steps for all version of Node.js ? It feels hacky, isn't that possible to have only one step with all versions? 

Just a little feedback, from what I've seen in the industry, it's more common to have a Docker tag for each major version. 

Example: 

- [Docker node](https://hub.docker.com/_/node/) `docker pull node:8` / `docker pull node:6`
- [Docker Nginx](https://hub.docker.com/_/nginx/) `docker pull nginx:1`

Because if we switch to Gladys v4 with breaking changes, it will breaks everyone using this. A "stable" doesn't work!

Can you switch to a tag by major version ? (you split the version by "." and take only the first number)

VonOx:
> Why do you have to repeat your steps for all version of Node.js ? It feels hacky, isn't that possible to have only one step with all versions?

Need some experiment.

> Can you switch to a tag by major version ? (you split the version by "." and take only the first number)

Yep.

VonOx:
https://hub.docker.com/r/vonox/gladys/tags/
https://travis-ci.org/VonOx/Gladys-trytravis/jobs/433717903

Tested on release/tag for major V4.

Also added `cache: yarn` to test stage for performance.

Pierre-Gilles:
Really great job, it works nice! I love the multi-stage view on Travis :) 

👏 👏 👏 👏

PR #400, opened 2018-08-20
improve condition in launcher

Jean-PhilippeD:
I've added a trueIf field in state so that when a condition in a launcher is added, the user can either choose the condition to be true if state are true, or true if state are false.


Pierre-Gilles:
Thanks for your PR! 

I just answered on the forum => https://community.gladysproject.com/t/evolution-de-la-page-condition-dans-la-creation-des-scenarios/3140/7

Pierre-Gilles:
Hey! 

Thanks for your PR, the feature is really nice 👍 

I just merged it, but one feedback: 

Please write some tests 🙏 I had to write tests myself to test if your feature was working properly!

Here is the kind of test expected : 

screenshot 2018-10-12 at 10 27 10

I had to write many fixtures to write a consistent test, it's important!

PR #401, opened 2018-08-21
add mode confition in launcher

Jean-PhilippeD:
add condition if house is in specific mode in launcher


Jean-PhilippeD:
gonna move isMode in house core instead of mode

Issue #361, opened 2018-05-29
Display problem of multiple slider in device view


Pierre-Gilles:
See issue here => https://community.gladysproject.com/t/gladys-3-8-1-disponible-en-beta/2824/53

PR #409, opened 2018-09-02
issue #361

piznel:


Pierre-Gilles:
Thanks for your PR! 

It indeed works very well now :)

PR #410, opened 2018-09-09
Add a deviceType.getByTag function to Core

Boimb:
As a member of the community asked for this, this PR adds the ability to find deviceTypes by their tag.
Note: This is implemented only in the core API, note in the HTTP.

### DIFF
--- a/api/core/devicetype/deviceType.getByTag.js
+++ b/api/core/devicetype/deviceType.getByTag.js
@@ -1,8 +1,11 @@
 var queries = require('./deviceType.queries.js');
 
+/**
+ * options must have a tag attribute
+ * tag can be a simple string or an array of string to match multiple tags
+ * So {tag: 'myTag'} and {tag: ['myTag', 'anotherTag']} are both valid
+ * @param options
+ */
 module.exports = function(options){
-  const query = Array.isArray(options.tag)
-    ? queries.getDeviceTypeByTags
-    : queries.getDeviceTypeByTag;
-  return gladys.utils.sql(query, [options.tag]);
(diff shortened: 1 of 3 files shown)


Boimb:
I've intentionally split the queries getByTag (string parameter) and getByTags (array parameter) to have clear code.
I have to mention that two different queries is a bit overkill as the 'WHERE IN' clause in the 'getByTags'  is ok with a single string.

Pierre-Gilles:
Hey!

Thanks for your PR! :) 

What was exactly the use case of this ? Why do we want to be able to search by tag ? 

If you have the link to the discussion it would be great.

> I have to mention that two different queries is a bit overkill as the 'WHERE IN' clause in the 'getByTags' is ok with a single string.

Can we maybe keep only the WHERE IN then ? :)

Boimb:
> If you have the link to the discussion it would be great.

That discussion was on the discord :/
Hamtaro wanted to make a script to find all states of his "Door/Window opened" sensors. I think Tag was the only attribute (or maybe the most obvious at first glance) to filter those deviceTypes.
Maybe there's another (better) approach, like getting deviceTypes by service or category then filter results...
I must admit I didn't ask for more :D I just pushed this as I assumed this was a missing "findBy".

> Can we maybe keep only the WHERE IN then ? :)

For sure.

Pierre-Gilles:
Hey! Thanks for your PR, I just merged it! 🙂 

Please, next time: 

➡️ Write Unit tests (I had to do it myself here)
➡️ Update JS documentation (same, I had to write it myself here..)

PR #414, opened 2018-09-10
Correction of the slider width to no longer display the horizontal scroll + doc for deviceType.getByCategory

piznel:
If you set `slider-horizontal-width` to `100%`; the slider-tick-label goes beyond the container at max value :
**before:**
enter image description here
 **After:**
enter image description here


Pierre-Gilles:
Hey!

Thanks for the PR! 

I'm trying to reproduce your bug but i can't manage to reproduce it... 

When exactly does it happens? Is there with a specific screen resolution? :)

piznel:
I have it with Windows 7, on Chrome and Firefox (latest version), at the resolution recommended by my screen.
On Ipad too : 
3c50b6ef-5716-41f5-ba2e-7c813685b440

Pierre-Gilles:
Got it! I see what you mean and i'm able to reproduce indeed!

I'm merging right now

PR #415, opened 2018-09-10
Ui device Improvements 

piznel:
Adding :
- sorting on the columns of the device and devicetype tabs,
- choice of the devicetype display order in the first device tab and in the box device-room,
- choice of the background color of each devicetype line in the first device tab and in the device-room box.

image

image


piznel:
Need to redo it cleanly following the latest modifications of the core in assets.

PR #419, opened 2018-09-18
Fix area behavior

Pierre-Gilles:


Boimb (commented):
So Pepper works in Kazakhstan ?!?
:D

PR #422, opened 2018-09-20
Ui device improvements

piznel:
This PR allows us to transform  the view below, on which we can see several problems:

- deviceTypes of the same device are not grouped together,
- the deviceTypes to turn on and off, which for me are the most important, are drowned in the middle of all the others.

image

into this one:

image

___

The addition of 2 fields ("order" and "color") in the "deviceType" table allows you to order the display of deviceTypes and to choose the background color individually.

By clicking on this icon : 
image

you open the following modal window:

image

In the first column, you can choose the background color of the devicetype from a palette of 32 colors (from the Gladys theme), and in the second column, the display order. 

The "Peripherals per piece" box has also been modified to incorporate these changes, as well as the addition of the "color-picker" to the "color" type devicetypes:

image


Pierre-Gilles:
Hey! :) 


> deviceTypes of the same device are not grouped together,

Really good idea to fix this! Thanks!

> the deviceTypes to turn on and off, which for me are the most important, are drowned in the middle of all the others.

Ok so now you put it at first? nice! 

> The addition of 2 fields ("order" and "color") in the "deviceType" table allows you to order the display of deviceTypes and to choose the background color individually.

The order idea is a good one!

For the color, to be completely honest I don't know if it's something we want in Gladys ^^ 

Why do you want that in Gladys?

piznel:
Color, for example, allows you to quickly identify the deviceType you use the most, if you wish.
In my case, it allows me to identify devicetypes of the same device:
image

I know that these elements form a block, a bit like a chapter in bold and paragraphs. So this allows me to greatly simplify the names of the deviceType: there is no need to repeat to which device they are attached and they can have the same name.

And finally, I put the flat colors of AdminLTE, and I think it looks pretty :) .. But this is a matter of taste ;)

piznel:
Here is the PR without the choice of color.
I added 2 small modifications, because they concern the same view:
- Change due to the color of the toggle: indeed, it was not exactly the same as the color of the theme : before was #428BCA 
image
 instead of #3c8dbc
image

- Decrease the width of the color picker: I suggest you make it as wide as the toggle button : 
image

Of course, I made the same changes at the box.

I hope that's okay with you.

atrovato:
Fine for me (color-picker creator).

Pierre-Gilles:
Hey!

I agree for the small changes, but I'm still not in for the choose order view... 

This has a huge impact everywhere and I really prefer to take some time thinking about this before integrating these changes... sorry! 

That's why I'm not merging this now, I prefer that we took the time to talk about this together before.

PR #425, opened 2018-09-25
Emit event gladys-launched at startup.

Boimb:
Emit the event "gladys-launched" when Gladys starts up. 
Goes with https://github.com/GladysProject/gladys-data/pull/24


Pierre-Gilles:
Good idea, good work 👏 

Checking that tomorrow morning

PR #426, opened 2018-09-27
Feature/check module update

vincentBesseau:
Adding version check between the in-store module and the installed module.

If the ***store version is newer than the one installed*** then we create a notification like this:
capture d ecran 2018-09-27 a 20 17 29

### DIFF
--- a/api/core/module/index.js
+++ b/api/core/module/index.js
@@ -16,3 +16,4 @@ module.exports.getById = require('./module.getById.js');
 module.exports.uninstall = require('./module.uninstall.js');
 module.exports.update = require('./module.update.js');
 module.exports.upgrade = require('./module.upgrade.js');
+module.exports.checkUpdate = require('./module.checkUpdate.js');
--- a/api/core/module/module.checkUpdate.js
+++ b/api/core/module/module.checkUpdate.js
@@ -21,7 +21,7 @@ module.exports = function(){
                                     var notification = {
                                         title: sails.__({ phrase: 'notification-update-available-title', locale: admin.language.substr(0,2)}) + ` ${moduleOnStore.version}`,
                                         text: `${moduleOnStore.name} ` + sails.__({ phrase: 'notification-update-module-available-text', locale: admin.language.substr(0,2)}) + ` ${moduleOnStore.version}`,
-                                        icon: 'fa fa-puzzle-piece',
+                                        icon: 'fa fa-cloud-download',
                                         iconColor: '',
                                         link: '/dashboard/module',
                                         priority: 0,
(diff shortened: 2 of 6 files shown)


VonOx:
Great ! Maybe we can hide update button if no update was detected ?

MathieuAndrade:
No, do not hide the update button because some modules may come from github and not from the store. 😉

VonOx:
So before, check if it come from the store. Not a real problem.

But this button is useless 90% of the time

vincentBesseau:
Hi @VonOx,
I'm agree with @MathieuAndrade about to not hide the update button.
But if you want to hide this button, you can make an issue on Gladys to do this.

I think this feature must be only check modules and make a notification to the user if a module was not update.

@Pierre-Gilles What do you think about it ?

VonOx:
I'm talking about update button but I'm ok, no problem

UI could be smart 😉

Pierre-Gilles:
Hey! Really nice idea, I love it! 👏 

About your implementation, I have a few feedbacks: 

- When you do a `GET developer.gladysproject.com/api/v1/modules`, this API is paginated by 50, so at the moment it works but if the store grows it won't. Plus, downloading the entire store JSON is probably not that performant if the store get bigger. Instead, you can use the getBySlug route and get one by one each module that the users have installed. For sure it will be a little slower today (because our store is currently small, so 10 requests vs 1 requests is slower), but in the future when it'll grow this implementation will be faster. You can use this API route : `GET https://developer.gladysproject.com/api/v1/modules/:slug` 

Example : 

https://developer.gladysproject.com/api/v1/modules/hue

- When you loop and do some asynchronous tasks, please use Promise.map instead of  forEach which is for synchronous work. 

In your code I'm speaking of  `modules.forEach(function(module) {`

Plus, it'll help avoid the callback hell you've get into in this PR :D 

-  Your comparaison `if(moduleOnStore.version > module.version) {` doesn't work! 

If a module has a version 3.1.10 and the version installed is 3.1.9, your code will say that the version installed is more recent because 9 > 1. It's false! 

PS: I just realised I did the same mistake in Gladys code when I wrote the code to check for Gladys update a few year ago, my bad ^^ It should be corrected as well aha.

Use the Node.js semver package instead (we use it already in Gladys in the tasks.dbMigration.js)

### My idea here

- Get the list of modules installed
- do a Promise.map on the list. Please put a concurrency of 1 on the Promise.map to avoid DDOSing the gladys developer platform, and on each module:
  - check with semver if the module online is more recent
  - If yes, send a notification

vincentBesseau:
Thanks @Pierre-Gilles for your feedbacks.

I treated your feedback in the following commits 
7740063 & bac37c6

Pierre-Gilles:
I did a few fixes and merged it :) 

Thanks for your PR again!

PR #429, opened 2018-10-03
 Changing prod folder to install asset by module

### LINKED ISSUE #391: Investigate if sails is copying assets to the .tmp folder or www in prod mode and copy module assets to the right one
https://community.gladysproject.com/t/resolu-module-zwave-pas-de-traduction/3144/9

vincentBesseau:
fix #391


Pierre-Gilles:
Thanks for your PR! In fact we should copy at both place! 

I just did the change on master :)

PR #430, opened 2018-10-03
Ui push button : fix #362

### LINKED ISSUE #362: Create a new type of button "press" that just has 1 (no 0 state)
https://community.gladysproject.com/t/devicetype-de-type-push-button/2886

piznel:
This improvement gives Gladys the opportunity to have a push button in its interface.
It was done as an AngularJS directive, callable by HTML tag `<push-button > </ push-button>`
To use it, just assign the type 'push' to a devicetype.
We then get in the device page:

push-button

On a mouse click, the '**push-boutton**' sends the `ng-model` value, as a pulse, just like any push button or remote control button.
Thus by setting `ng-model="type.max"`, it is the maximum value of the devicetype that is sent to each click.

### DIFF
--- a/assets/js/app/app.module.js
+++ b/assets/js/app/app.module.js
@@ -3,5 +3,5 @@
     'use strict';
 
     angular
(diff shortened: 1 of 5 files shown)


vincentBesseau:
Cool ! 
I waited for this evolution before publishing the broadlink module.

Pierre-Gilles:
Great work!! 👏 Testing this tomorrow morning in my PR review time

PR #432, opened 2018-10-08
fix Issue #413 : Detection of devicetype by the brain

### LINKED ISSUE #413: When some devicetypes have no tags, there are still detected
Ref => https://community.gladysproject.com/t/brain-qui-classifie-bien-mais-agit-sur-le-mauvais-devicetype/3111/8

piznel:
fix #413, with brain improvements:
- the tag search is no longer case sensitive (for example : "chambre"="Châmbre"="chamBré")
- if a room is present in the sentence, only the devicetype of that room will be returned,
- on the "on" command, the devicetype receives its maximum value instead of 1 ; Same thing to turn off.

Thanks to @MathieuAndrade and @vincentbesseau for being top-of-the-range beta testers :+1:

### DIFF
--- a/api/core/brain/brain.classify.js
+++ b/api/core/brain/brain.classify.js
@@ -24,7 +24,6 @@ module.exports = function classify(user, message){
             if(scope.user.password) delete scope.user.password;
 
             var classifier = shared.getClassifier();
-            console.log('scope.replacedText', scope.replacedText);
 
             var classifications = classifier.classify(scope.replacedText);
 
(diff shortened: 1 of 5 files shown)


Pierre-Gilles:
Hey! 

Thanks for your PR, nice work! 👏 

Following our discussion on the forum, I think we should keep the actual behaviour for the deviceType selection before doing the "big evolution" that we are planning with category.

So can you remove your changes on point 2 "if a room is present in the sentence, only the devicetype of that room will be returned", and keep only point 1 and 3 on this PR ?

I'm doing a morning PR merge/review tomorrow morning to do a build then before the Gladys Gateway first alpha (I want to do a release without the gateway)

piznel:
Hello, @Pierre-Gilles,
It's done, death in the soul, because while waiting for the evolution, it was practical in everyday life.

PR #435, opened 2018-10-28
Create Dockerfile.arm32v7

anthonyraymond:
### Description of change

Since the `Dockerfile` use the alpine image of Node, It's not multi-arch compliant. So here is a Dockerfile for arm32v7 platform (like raspberry).

A better option IMO would be to base the `Dockerfile` on `node:8` instead of `node:8-alpine`, it will make it multi-arch compliant automatically. But i may understand if you don't wish to double the image size.

**EDIT**: i deliberately choosed stretch version to allow libopenzwave installation (if needed)

### DIFF
--- a/Dockerfile.arm32v7
+++ b/Dockerfile.arm32v7
@@ -1,0 +1,20 @@
+FROM arm32v7/node:8
+
+# Add tzdata for timezone settings
+RUN apt-get update && apt-get install -y tzdata
+
+# Create src folder
+RUN mkdir /src
+
+WORKDIR /src
+ADD . /src
+
+RUN apt-get update && apt-get install -y make gcc g++ python git && \
+    yarn install --production && npm install -g grunt-cli && grunt buildProd && npm uninstall -g grunt-cli \
+    apt-get remove -y make gcc g++ pyton git && \
+    rm -rf /var/lib/apt/lists/*
+
+# Export listening port
+EXPOSE 8080
+
+CMD ["node" ,"app.js"]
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
Nice work! One question: why Arm32v7 and not arm32v6? Here you're excluding all Raspberry Pi B/Zero!

anthonyraymond:
Hi 😄 

If you want arm32v6 the regular Dockerfile is the way to go,

Node is an *official docker image*, it means that from a unique base image (in this case node:8-alpine) you can build from multiple architecture (amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x).
If you try to **build** the `Dockerfile` from an arm32v6 arch it will automatically download `arm32v6/node:8-alpine` base image.

This is the wonderfull **theory** of docker's multi-arch official image. But in the real world, it does not always work. In our case `arm32v6/node:8-alpine` will works, but the `arm32v7/node` does not have a `8-alpine` tag so...

In this very case it's kinda weird since Debian and Alpine didn't quite understood on what is arm32v6 and arm32v7. See https://github.com/gliderlabs/docker-alpine/issues/298#issuecomment-309551513
This is the reason why arm32v7 does not have an alpine tag

VonOx:
This is build problem or usage problem of current alpine image ?

Can you add nmap package ?  ( Used by some module )

VonOx:
After some reading, i have understood. This PR is a workaround 

I think we need to use manifest to build multiarch images ( Gladys for ALL 😄  ) , but this require some learning/work/test.

For reference: https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list

Pierre-Gilles:
I'm merging this in the meantime, this is a great PR :) We'll investigate for a multiarch image in the future! 

Thanks!

PR #441, opened 2018-11-01
verify position relative to accuracy

Jean-PhilippeD:

- [x] Are tests passing?

### Description of change

I fix the issue in havershine formula (division by 2)
I've changed the way Gladys checked if the user left or entered an area.

- Gladys check before the area where the user is inside by verifying if he left it by taking accuracy into account. (it virtually increase the size of the radius)
- Then Gl

adys check if the user has entered the areas he was not inside, without taking accuracy into account this time.

It should cover mostly of the cases, but sometimes, user could really left or enter an area while Gladys would not register it because of a bad accuracy.


Pierre-Gilles:
Thanks for your PR! Just merged it :)

PR #451, opened 2018-11-13
ci: test Node.js 11

DanielRuf:

- [ ] Are tests passing?
- [ ] Is the linter (eslint) passing?

### Description of change

This PR adds `node` to the test matrix so we test also against the current stable Node.js version (which is currently 11) to ensure that Gladys also works there.


Pierre-Gilles:
Thanks for your PR, good idea :)

PR #454, opened 2018-11-15
Add 'service' option at deviceService.get

piznel:

- [ ] Are tests passing?
- [ ] Is the linter (eslint) passing?

### Description of change

Just add "service" as params at route. The doc is already good :
image

### DIFF
--- a/assets/js/app/device/device.service.js
+++ b/assets/js/app/device/device.service.js
@@ -1,97 +1,97 @@
 
 (function () {
-    'use strict';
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
Good idea, but as you modify just the service, why is it useful? ^^ :) 

Explain your use case!

Pierre-Gilles:
Ok, understood it was for your other PR. Merged

PR #456, opened 2018-11-21
Add modules.getMethods()

piznel:

- [x] Are tests passing?
- [x] If your changes modify the API (REST or Node.js), did you modify the documentation?

### Description of change
this function returns a list of methods passed in parameters with a true or false value if they are exposed by a module.
Example : Testing if a module expose API Television's method 
```
 params = {
            'module' :"livebox",
            'service':"television",
            'methods' :["getState", "getChannel", "getMuted"]
        };
```

  => `{ getState: true, getChannel: false, getMuted: false }`
The `param.service` is optionnel.

### DIFF
--- a/api/controllers/ModuleController.js
+++ b/api/controllers/ModuleController.js
@@ -38,6 +38,25 @@ module.exports = {
       .catch(next);
   },
 
+  /**
+   * @api {post} /module/getMethods Get Methods
+   * @apiName modulegetMethods
+   * @apiGroup module
+   * @apiPermission authenticated
+   *
+   * @apiUse param
+   *
+   * @apiSuccess {array} All module's methods tested
+   */
+  getMethods: function(req, res, next) {
+    gladys.module
+      .getMethods(req.body)
+      .then(function(result) {
+        return res.json(result);
+      })
+      .catch(next);
+  },
+
   upgrade: function(req, res, next) {
     gladys.module.upgrade({ id: req.params.id, version: req.body.version });
     return res.json({ message: 'Upgrade started with success' });
--- a/api/controllers/ModuleController.js
+++ b/api/controllers/ModuleController.js
@@ -44,7 +44,9 @@ module.exports = {
    * @apiGroup module
    * @apiPermission authenticated
    *
-   * @apiUse param
+   * @apiParam {String} module slug of the module.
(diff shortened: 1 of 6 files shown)


piznel:
First documentation, first test... Not so simple as it seems to be writing! Sorry for the last commits....
The strange thing is that in my case, the tests had passed, as well as the generation of the documentation.

Boimb:
Hey @piznel ,
Really nice feature.
I was self-questionning : 
Why does this return an array of objects with boolean values ? Wouldn't it be more easy and relevant to get back just an array of "the trully" values in the getMethods ?
From your example it could be : 
```
['getState']
```
As you put an array of strings in parameters, my point is that it would make more sense.
The result would be more pratictal to use as you just have to check if the method name is in the results.
The code could be like : 
```
module.exports = function getMethods(params) {
  return Promise.resolve(params.methods.filter(method => params.hasOwnProperty('service')
  // If 'service' check if service && method exists
  ? gladys.modules[params.module] && 
    gladys.modules[params.module[method]] &&
    typeof gladys.modules[params.module][params.service][method] === 'function'
  // Else check only if method exists
  : gladys.modules[params.module] && 
    gladys.modules[params.module[method]]));
};
```
And then, you could use it like : 
```
const params = {
  'module' :"livebox",
  'service':"television",
  'methods' :["getState", "getChannel", "getMuted"]
};

const methods = gladys.module.getMethods(params);
// Checks
const availables = methods
  .filter(methodName => params.methods.includes(methodName))
const missing = methods
  .filter(methodName => !params.methods.includes(methodName))
const allIn = missing.length === 0;
```

piznel:
Hey @Boimb,
By sending it back like that, you just have to do a value test, without `filter `: 
`if(methods.getState) {...`

Boimb:
Yep. Ofc. For a single test, knowing the name of the function you're testing, it's easier the "object way".
But if you don't know the name of the function, or you want to "abstract" it like in an array, it would become less readable: 
```
const availables = params.methods
  .filter(method => gladys.module.getMethods(params)[method])
```
But, hey... That was just a simple point of view (mine). If you're OK and confortable with it, stick to that ;)
Great work btw.
And with tests! 👍 
"Chapeau bas" ;)

PR #466, opened 2018-12-16
Add room origin to brain

Jean-PhilippeD:

- [x] Are tests passing?

### Description of change

I've improved the brain classification by adding an optional field which is the room label where does the request come from, if available.

With this information, you can : 

- Ask to turn on the light without telling the room and it will turn on the light in the room you are.
- Get an answer on this room if the notification type handle the room

I've also added a percentage parser so that you can ask to turn the luminosity to 80% and Gladys will understand this, and do it (if your associated device handle it).

And finally, I've added a word parser, but as Gladys does not recognized by herself the word, it means you need to provide an array with the words detected in your sentence so Gladys will be able to parse the sentence and forward this word list to a function which could handle those words, like a module.


Pierre-Gilles:
Hey! Thanks for your PR! :)

I have a few feedbacks on this PR but I'm not sure if we want to spend lots of time on it instead of working directly on that on Gladys 4...

What do you think? :)

Pierre-Gilles:
Closing in favor of development in Gladys 4.

PR #470, opened 2019-01-13
Run loadenv file

sescandell:
So SQL configuration is grabbed from the .env file (and not assuming SQL Server is on localhost with default root user).

### Description of change

Load loadenv so SQL's credentials are loaded from any eventual .env file. Without this, Gladys tries to connect with root credentials on localhost.


sescandell:
Actually, the same issue comes with `node app.js` command (following the Getting Started part of the readme file. Is this part still up to date?).

To fix it, I simply force loadenv every time (in app.js file without checking the environment). I can update the PR to match this more global fix. Just let me know.

Pierre-Gilles:
Hey! Thanks for your PR. 

So to give you the history on that.

Dotenv files are mostly used for development purpose and in general most people don't use them for production deployment. That's why the dotenv file was only loaded in development environnement. 

However, in Gladys 3, I realized that most people don't know how to set environnement variables and find it easier to write their configuration in a dotenv file, which I can completely understand. 

Gladys 4 will be slightly different from that as by default it will use SQlite so no configuration will be needed. 

In the meantime, I think it's ok to merge this to make it easier for people to spin up a Gladys 3 instance while we are building the future of Gladys 🚀 

Thanks again for your PR!

PR #507, opened 2019-07-03
Add caldav service

bertrandda:
### Pull Request check-list

- [x] If your changes modify the API (REST or Node.js), did you modify the documentation? (Documentation is based on comments in code)

### Description of change

I add caldav service to sync Gladys calendar with an external calendar using caldav protocol.


server/migrations/20190211042644-create-calendar-event.js
```diff
@@ -18,7 +18,6 @@ module.exports = {
       },
       name: {
         allowNull: false,
-        unique: true,
```
bertrandda: I'll change the migration to a new file

server/services/caldav/index.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /**
+   * @public
+   * @description This function stops the CalDAV service
+   * @example
+   * gladys.services.caldav.stop();
+   */
+  async function stop() {
+    logger.log('stopping example service');
```
bertrandda:
```suggestion
    logger.log('stopping CalDAV service');
```

server/migrations/20190211042644-create-calendar-event.js
Pierre-Gilles: Perfect!

Pierre-Gilles:
Salut @bertrandda ! De retour de congés, je regarde une par une les PR.

Où en es-tu sur celle-ci ?

bertrandda:
Salut @Pierre-Gilles, j'ai un soucis d'affichage sur les évènements `fullDay` et sur l'affichage mensuel (en essayant pourtant tout ce qui était dit dans la doc BigCalendar). Je n'y ai pas retouché depuis le message sur le forum mais j'espère bientôt m'y remettre, il y a aussi la partie traduction que je dois finir.

Pierre-Gilles:
Ah bizarre, c'est à dire un soucis d'affichage? (tu as un screen?) Pas de problème similaire dans les issues de BigCalendar ? 

Si jamais BigCalendar marche mal et que tu trouves une autre alternative:

- Pas trop lourde et qui n'inclut pas des centaines de dépendances (regarde https://bundlephobia.com/ pour voir ça)
- Aussi clean en terme d'UI 
- Facile à utiliser dans le code

N'hésite pas à faire des propositions ici

hotfix31:
Il existe [fullcalendar](https://fullcalendar.io/) en alternative intéréssante.

bertrandda:
> Il existe [fullcalendar](https://fullcalendar.io/) en alternative intéréssante.

Merci, mais finalement le passage à Preact X a fixé les problèmes, on reste donc sur `react-big-calendar`

Pierre-Gilles:
Is this PR still WIP or ready for review? If not ready, can you put "[WIP]" in the title? thanks!

front/src/routes/integration/all/caldav/CalDAV.jsx
```diff
@@ -0,0 +1,126 @@
...
+                        <span class="input-group-append">
+                          <button
+                            class="btn btn-primary"
+                            onClick={props.saveCaldavSettings}
+                            style={{ marginRight: '10px' }}
+                          >
+                            Save
+                          </button>
+                          <button class="btn btn-success" onClick={props.startSync}>
+                            Sync now
```
bertrandda: Use preact-i18n

server/test/services/caldav/lib/calendar/syncCalendarEvents.test.js
```diff
@@ -0,0 +1,83 @@
...
+        },
+      },
+    ],
+  },
+];
+
+const formatedEvents = [
+  {
+    external_id: 'e52c11e3-af8a-48c7-9f54-de7aba373c46',
+    selector: 'Event 1 2019-02-25-1000',
```
Pierre-Gilles: Just realized it's maybe not enforced enough, but a "selector" should not contains spaces, it should be ok to be in URL. Selector can be formatted with the [addSelector function](https://github.com/GladysAssistant/Gladys/blob/master/server/utils/addSelector.js)

server/services/caldav/lib/calendar/calendar.sync.js
```diff
@@ -0,0 +1,22 @@
+const logger = require('../../../../utils/logger');
+
+/**
+ * @description Sync calendars.
```
Pierre-Gilles: Name of the function is not what it is!

server/services/caldav/lib/calendar/calendar.formaters.js
```diff
@@ -0,0 +1,197 @@
...
+
+  caldavEvents.forEach((caldavEvent) => {
+    let icsEvent = this.ical.parseICS(caldavEvent.calendarData);
+    icsEvent = icsEvent[Object.keys(icsEvent)[0]];
+
+    if (icsEvent.type !== 'VEVENT') {
+      return;
+    }
+
+    if (typeof icsEvent.rrule === 'undefined') {
```
Pierre-Gilles: Is it rrule or "rule" ? (just asking)

server/services/caldav/lib/calendar/calendar.formaters.js
```diff
@@ -0,0 +1,197 @@
...
+      const newEvent = {
+        external_id: icsEvent.uid,
+        selector: `${icsEvent.summary} ${this.moment(icsEvent.start).format('YYYY-MM-DD-HHmm')}`,
```
Pierre-Gilles: A selector cannot have spaces! Just realized it's not enforced enough, but selector should be formatted with the "addSelector" function

server/services/caldav/lib/calendar/calendar.syncCalendars.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @description Sync user calendars.
+ * @param {Array} calendars - External calendars.
+ * @param {Object} userId - Gladys calendar owner.
+ * @returns {Promise} Resolving with all calendars sync.
+ * @example
+ * syncCalendars(calendars, userId)
+ */
+async function syncCalendars(calendars, userId) {
```
Pierre-Gilles: I'm not sure this function name really describe what it does. I see nothing about syncing here: no request is made in this function right?

server/services/caldav/lib/calendar/calendar.syncAllUsers.js
```diff
@@ -0,0 +1,28 @@
...
+ * @returns {Promise} All calendars are sync.
+ * @example
+ * syncAllUsers();
+ */
+async function syncAllUsers() {
+  try {
+    const users = await this.gladys.user.get();
+    const service = await this.gladys.service.getLocalServiceByName('caldav');
+
+    return Promise.all(
```
Pierre-Gilles: Be careful here, if you sync all users in parallel, you can spam the Caldav server and the Gladys instance can be rate-limited and lose access to the server for a short period of time.

Imagine we have 5 users in the family, having each 5-6 calendars, it's easily 30 requests/seconds to the provider, I don't think it's something we want.

There are several ways of rate limiting calls: 

- You can use bluebird [Promise.map](http://bluebirdjs.com/docs/api/promise.map.html) with a `concurrency` option.
- You can rate limit at the request level by creating a pool of request in the `request` module (see the maxSockets option https://github.com/request/request#requestoptions-callback)

server/services/caldav/lib/calendar/calendar.syncUser.js
```diff
@@ -0,0 +1,54 @@
...
+
+  await Promise.all(
+    davCalendars.map(async (calendar) => {
+      const events = await this.gladys.calendar.getEvents(userId, {
+        calendarId: calendar.id,
+      });
+
+      await Promise.all(events.map((event) => this.gladys.calendar.destroyEvent(event.selector)));
+
+      return this.gladys.calendar.destroy(calendar.selector);
```
Pierre-Gilles: If I understand well, you destroy/re-create all events at each sync?

I see two problems with that:

- During the sync, Gladys is in a "buggy" state. If the user asks Gladys something, or if a scene runs during the sync,  Gladys will not be able to work and it will break, because the DB will be empty. As the sync happens every 30 minutes, it's something which will happen quite frequently.
- Unique ID are re-generated at each sync, meaning it can break things: an external script setup by the user, a scene which rely on an event

Why didn't you choose to update existing events?

server/services/caldav/lib/calendar/calendar.syncUser.js
```diff
@@ -0,0 +1,54 @@
...
+ * syncUser(user.id)
+ */
+async function syncUser(userId) {
+  const account = await this.connect(userId);
+
+  // Clean all dav Calendar for full Sync
+  let gladysCalendars = await this.gladys.calendar.get(userId);
+  let davCalendars = gladysCalendars.filter((calendar) => calendar.service_id === this.serviceId);
+
+  await Promise.all(
```
Pierre-Gilles: Same here, be careful with concurrency

server/services/caldav/lib/calendar/calendar.syncUser.js
```diff
@@ -0,0 +1,54 @@
...
+      return this.gladys.calendar.destroy(calendar.selector);
+    }),
+  );
+
+  await this.syncCalendars(account.calendars, userId);
+
+  gladysCalendars = await this.gladys.calendar.get(userId);
+  davCalendars = gladysCalendars.filter((calendar) => calendar.service_id === this.serviceId);
+  // foreach calendar, sync events
+  return Promise.all(
```
Pierre-Gilles: Same here, careful with concurrency

Pierre-Gilles (changes requested):
Hi! 

First, thanks for this great PR, it's really great and I appreciate the time you took to develop this cleanly, congrats for the great work 👏 👏 

I wrote a few feedbacks on this review. Don't take this personally, it's just constructive feedback so we can make this integration really great before it goes live! 😄 

I have the feeling that you were inspired by the code of the v3 for this development. On this service, I feel that the v3 was not really well organized and maybe not a great model.

The structure of the code could be improved: for example, I feel the code right now is a cascade of functions call in a spaghetti order (sync, syncUser, syncAll, syncOneCalendar, ...) it's hard to follow the path when reviewing ^^ Don't hesitate to refactor this if you have any idea :)

Don't hesitate if you have any questions, and again good job 👍

server/test/services/caldav/lib/calendar/syncCalendarEvents.test.js
bertrandda: You are right, but here it is only the formatter output. We use `addSelector` here https://github.com/GladysAssistant/Gladys/blob/d0f956a29ab7671ddcad32a121305fc578a3e134/server/models/calendar_event.js#L52 when we save the data. Look the selector in database:
image

server/services/caldav/lib/calendar/calendar.formaters.js
bertrandda: It's `rrule` for recurrence rule -> https://www.npmjs.com/package/ical#recurrences-and-exceptions

server/services/caldav/lib/calendar/calendar.syncCalendars.js
bertrandda: Yes it's true, I think it is more a function to `updateCalendars`. I am going to change also `syncCalendarEvents` to `updateCalendarEvents`.

server/services/caldav/lib/calendar/calendar.syncAllUsers.js
bertrandda: Indeed not the best way, I'll try to improve that.

server/services/caldav/lib/calendar/calendar.syncUser.js
bertrandda: Yes, I choose to do it like that. In my first tests (when I develop the v3 module), I've been having problems by updating existing events when user choose to delete an event and he re-sync Gladys with caldav. I'm still gonna retry incremental synchronisation, we want the best solution.

server/services/caldav/lib/calendar/calendar.formaters.js
bertrandda: It is the formater, spaces are remove when we save the event.

server/services/caldav/lib/calendar/calendar.syncUser.js
bertrandda: I understand on the previous case but here we already have all data, there is no request all is executed locally, is it necessary to limit concurrency ?

server/services/caldav/lib/calendar/calendar.cleanUp.js
```diff
@@ -0,0 +1,27 @@
...
+async function cleanUp(userId) {
+  logger.info(`Start cleaning CalDAV data for user: ${userId}`);
+  const gladysCalendars = await this.gladys.calendar.get(userId, { serviceId: this.serviceId });
+  logger.info(`${gladysCalendars.length} calendars to clean`);
+  await Promise.all(
+    gladysCalendars.map(async (calendar) => {
+      const events = await this.gladys.calendar.getEvents(userId, {
+        calendarId: calendar.id,
+      });
+      await Promise.all(events.map((event) => this.gladys.calendar.destroyEvent(event.selector)));
```
Pierre-Gilles: I don't think you need to do that, there is a ON DELETE CASCADE on calendar deletion, no?

server/services/caldav/lib/calendar/calendar.cleanUp.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Object} userId - Gladys user to clean up.
+ * @returns {Promise} Resolving with client connected.
+ * @example
+ * cleanUp(user.id)
+ */
+async function cleanUp(userId) {
+  logger.info(`Start cleaning CalDAV data for user: ${userId}`);
+  const gladysCalendars = await this.gladys.calendar.get(userId, { serviceId: this.serviceId });
+  logger.info(`${gladysCalendars.length} calendars to clean`);
+  await Promise.all(
```
Pierre-Gilles: Be careful with promise.all on big accounts, if you have lots of calendar, can be too much in the call stack... 

You can use bluebird Promise.map instead with a concurrency set

server/services/caldav/lib/calendar/calendar.formaters.js
```diff
@@ -0,0 +1,198 @@
...
+function formatCalendars(caldavCalendars, userId) {
+  const calendars = [];
+  caldavCalendars.forEach((caldavCalendar) => {
+    const newCalendar = {
+      external_id: caldavCalendar.url,
+      name: caldavCalendar.displayName,
+      description: caldavCalendar.description || `Calendar ${caldavCalendar.displayName}`,
+      service_id: this.serviceId,
+      user_id: userId,
+      ctag: caldavCalendar.ctag,
```
Pierre-Gilles, line 184: What's a ctag and why is it useful ? (just asking)

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
```diff
@@ -0,0 +1,189 @@
...
+
+      if (formatedCalendar.ctag !== gladysCalendar[0].ctag) {
+        await this.gladys.calendar.update(gladysCalendar[0].selector, formatedCalendar);
+        return gladysCalendar[0];
+      }
+      return null;
+    }),
+  );
+
+  await Promise.map(
```
Pierre-Gilles, line 61: Starting here, the function starts to be reaaaally big and have almost no comment.

It's hard to understand what you are doing !

Please add more comments, and maybe cut this function in smaller one, because here I'm lost :)

server/services/caldav/lib/config/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // Get Home URL
+  const {
+    props: { calendarHomeSet },
+  } = await xhr.send(req, CALDAV_PRINCIPAL_URL);
+  const CALDAV_HOME_URL = url.resolve(CALDAV_PRINCIPAL_URL, calendarHomeSet);
+  logger.info(`CalDAV : Home URL found: ${CALDAV_HOME_URL}`);
+  await this.gladys.variable.setValue('CALDAV_HOME_URL', CALDAV_HOME_URL, this.serviceId, userId);
+
+  return Promise.resolve({});
```
Pierre-Gilles: Why do you resolve with an empty object?

If there is nothing to return, return nothing :)

server/services/caldav/package.json
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  "cpu": [
+    "x64",
+    "arm",
+    "arm64"
+  ],
+  "scripts": {},
+  "dependencies": {
+    "bluebird": "^3.7.0",
+    "dav": "git+https://github.com/bertrandda/dav.git#gladys-custom",
```
Pierre-Gilles: The package should be in NPM, not git !

server/services/caldav/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "ical": "^0.6.0",
+    "moment": "^2.24.0",
+    "request": "^2.88.0",
+    "request-promise-native": "^1.0.7",
```
Pierre-Gilles: This package is deprecated, see

https://github.com/request/request-promise-native

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
```diff
@@ -0,0 +1,189 @@
...
+  logger.info(`CalDAV : Found ${davCalendars.length} calendars.`);
+
+  const formatedCalendars = this.formatCalendars(davCalendars, userId);
+  const calendarsToUpdate = await Promise.all(
+    formatedCalendars.map(async (formatedCalendar) => {
+      const gladysCalendar = await this.gladys.calendar.get(userId, { externalId: formatedCalendar.external_id });
+      if (gladysCalendar.length === 0) {
+        const savedCalendar = await this.gladys.calendar.create(formatedCalendar);
+        delete savedCalendar.dataValues.sync_token;
+        return savedCalendar.dataValues;
```
Pierre-Gilles: MMm it's weird that you have to use the `. dataValues` attribute, you should not!

You can modify the calendar.create function to return only a pure JS object ( with calendar.get({ plain: true }))

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
```diff
@@ -0,0 +1,189 @@
...
+      }
+
+      if (formatedCalendar.ctag !== gladysCalendar[0].ctag) {
```
Pierre-Gilles, line 53: Can you explain what's happening here?

server/services/caldav/index.js
```diff
@@ -0,0 +1,68 @@
...
+   * gladys.services.caldav.start();
+   */
+  async function start() {
+    logger.log('starting CalDAV service');
+    interval = setInterval(async () => {
+      try {
+        const users = await gladys.user.get();
+        const service = await gladys.service.getLocalServiceByName('caldav');
+
+        await Promise.map(
```
Pierre-Gilles: Don't do all that in the start function ! 

The start function should be clean and concise, try to put that in a separate function so the code can be more easily tested, and more understandable !

Pierre-Gilles (changes requested):
Hi ! 

Thanks again for this great PR 👏 

This is amazing work !! 

I have lots of questions and little feedback that I put this in this review. As I have little knowledge of how caldav work, some questions are a bit naive but it's intentional :) I read only the server code for now.

If you can write a little message on the approach you took this time it would be very helpful for me to better understand what you did. 

Thanks 😄

bertrandda:
How it works now :
- List all account calendars
- For each calendar
	- If calendar not exist in DB -> create it
	- Else update existing one
- For each calendar
	- Get changes
	- Delete event if necessary
	- For each events
		- If event not exist in DB -> create it
		- Else update existing one

server/services/caldav/lib/calendar/calendar.cleanUp.js
bertrandda: Yes, you are rights

server/services/caldav/lib/calendar/calendar.formaters.js
bertrandda: `ctag` is like a calendar status. If `ctag` changes, it means events change inside the calendar (create/update/delete). With this, we can skip useless requests to fetch event changes.

server/services/caldav/lib/config/index.js
bertrandda: Indeed, I remove that. It's a waste of the old way

server/services/caldav/package.json
bertrandda: Also a waste from the old way.

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
bertrandda: Only if `ctag` in database and new `ctag` are different, we request the changes

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
bertrandda: Ok I'll work on it

server/services/caldav/package.json
bertrandda: Published :wink:

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
bertrandda: Done !

Pierre-Gilles:
I just tried your service on my machine! 

When I sync an account, in the logs it seems ok (no errors logs, just green logs saying everything is ok and looping through all my calendars), but the request returns a 500 with no error message and an UI message which is not very helpful:

Screenshot 2020-02-27 at 17 48 02

By looking at the code, it seems you're always doing a big try catch in your route, always returning a 500 when something is wrong.

Screenshot 2020-02-27 at 17 56 24

It's not really helpful for the user! 

- Are the credentials wrong? 
- Is the server which is unavailable?
- Is Gladys broken?
- Is the calendar somehow broken? 

The user doesn't know what happened 🤷‍♂️

bertrandda:
> I just tried your service on my machine!
> 
> When I sync an account, in the logs it seems ok (no errors logs, just green logs saying everything is ok and looping through all my calendars), but the request returns a 500 with no error message and an UI message which is not very helpful:
> 
> Screenshot 2020-02-27 at 17 48 02
> 
> By looking at the code, it seems you're always doing a big try catch in your route, always returning a 500 when something is wrong.
> 
> Screenshot 2020-02-27 at 17 56 24
> 
> It's not really helpful for the user!
> 
>     * Are the credentials wrong?
> 
>     * Is the server which is unavailable?
> 
>     * Is Gladys broken?
> 
>     * Is the calendar somehow broken?
> 
> 
> The user doesn't know what happened 🤷‍♂️

I just tried to add that, but I am not sure I'm doing it in the best way. Can you check and tell me how to improve it ?

Pierre-Gilles:
@bertrandda I just checked, and no this is not what I meant. 

You are throwing a 500 error no matter what: we don't want that. 

Little example, the service DarkSky is throwing very specific errors directly in his code: 

https://github.com/GladysAssistant/Gladys/blob/master/server/services/darksky/index.js#L72

The service PhilipsHue is throwing very specific error as well: 

https://github.com/GladysAssistant/Gladys/blob/master/server/services/philips-hue/lib/light/light.configureBridge.js#L23

The rule is: 

- Using native errors that you can find in the `coreErrors` or `httpErrors` files.
- Throwing a proper error code, that the frontend will map to the proper i18n translation
- Never display a raw error in the fronted, but only a beautiful translation.
- Don't wrap your calls in the controllers in a try catch, but throw propers errors directly in you service! Try /catch should not be in the controller !  (Example of a nice controller: https://github.com/GladysAssistant/Gladys/blob/master/server/services/philips-hue/api/hue.controller.js )

bertrandda:
> @bertrandda I just checked, and no this is not what I meant.
> 
> You are throwing a 500 error no matter what: we don't want that.
> 
> Little example, the service DarkSky is throwing very specific errors directly in his code:
> 
> https://github.com/GladysAssistant/Gladys/blob/master/server/services/darksky/index.js#L72
> 
> The service PhilipsHue is throwing very specific error as well:
> 
> https://github.com/GladysAssistant/Gladys/blob/master/server/services/philips-hue/lib/light/light.configureBridge.js#L23
> 
> The rule is:
> 
>     * Using native errors that you can find in the `coreErrors` or `httpErrors` files.
> 
>     * Throwing a proper error code, that the frontend will map to the proper i18n translation
> 
>     * Never display a raw error in the fronted, but only a beautiful translation.
> 
>     * Don't wrap your calls in the controllers in a try catch, but throw propers errors directly in you service! Try /catch should not be in the controller !  (Example of a nice controller: https://github.com/GladysAssistant/Gladys/blob/master/server/services/philips-hue/api/hue.controller.js )

Ok so I improved it by following your indications. Now we have multiple errors for each identified case, I hope errors are the same for all CalDAV server, I tested with Nextcloud + Apple

Pierre-Gilles (approved):
I tested, it's really nice and way easier to understand! 

I tested on my iCloud account, worked at the first time :) 

Thanks for this PR, merging this now!

PR #517, opened 2019-07-22
MQTT service

atrovato:
### Pull Request check-list

- [x] If your changes modify the API (REST or Node.js), did you modify the documentation? (Documentation is based on comments in code)

### Description of change

MQTT integration completion (front and back)


server/services/mqtt/lib/handleNewMessage.js
```diff
@@ -14,6 +15,9 @@ function handleNewMessage(topic, message) {
       case 'gladys/master/device/create':
         this.gladys.event.emit(EVENTS.DEVICE.NEW, JSON.parse(message));
         break;
+      case 'gladys/master/devicestate/create':
```
atrovato: Used for backward compatibility

atrovato: Used for backward compatibility.

server/services/mqtt/lib/handler/handleGladysMessage.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+function handleGladysMessage(topic, message) {
+  switch (topic) {
+    case 'gladys/master/device/create':
+      this.gladys.event.emit(EVENTS.DEVICE.NEW, JSON.parse(message));
+      break;
+    case 'gladys/master/state/create':
+      this.gladys.event.emit(EVENTS.DEVICE.NEW_STATE, JSON.parse(message));
+      break;
+    case 'gladys/master/devicestate/create':
```
atrovato: Used for backward compatibility.

atrovato:
@Pierre-Gilles ready for review ;)

front/src/routes/integration/all/mqtt/device-page/setup/Feature.jsx
```diff
@@ -0,0 +1,96 @@
...
+            <i
+              class={`mr-2 fe fe-${get(
+                DeviceFeatureCategoriesIcon,
+                `${props.feature.category}.${props.feature.type}`
+              )}`}
+            />
+            <Text id={`deviceFeatureCategory.${props.feature.category}.${props.feature.type}`} />
+          </div>
+          <div class="card-body">
+            <div class="form-group" class="form-label" for={`featueName_${props.featureIndex}`}>
```
Pierre-Gilles: Little typo, "featureName"

front/src/routes/integration/all/mqtt/device-page/setup/Feature.jsx
```diff
@@ -0,0 +1,96 @@
...
+              <label>
+                <Text id="integration.mqtt.feature.nameLabel" />
+              </label>
+              <Localizer>
+                <input
+                  id={`featueName_${props.featureIndex}`}
```
Pierre-Gilles: Same typo

front/src/routes/integration/all/mqtt/setup-page/actions.js
```diff
@@ -0,0 +1,67 @@
...
+  const actions = {
+    async loadProps(state) {
+      let mqttURL;
+      let mqttUsername;
+      let mqttPassword;
+      try {
+        mqttURL = await state.httpClient.get('/api/v1/service/mqtt/variable/MQTT_URL');
+        mqttUsername = await state.httpClient.get('/api/v1/service/mqtt/variable/MQTT_USERNAME');
+        if (mqttUsername.value) {
+          mqttPassword = 'FAKE_PASSWORD';
```
Pierre-Gilles: Why do you put a fake password?

server/services/mqtt/lib/handler/handleGladysMessage.js
Pierre-Gilles: I don't think we want backward compatibility. We've done nothing in Gladys 4 to stay compatible with Gladys 3 (both are completely different), and to stay consistent I would avoid trying to be compatible at some place if we are not doing it everywhere.

And it's not working here, because a device/deviceState objects in Gladys 3 and Gladys 4 are completely different. Just throwing the object is not enough!

Pierre-Gilles: I would suggest removing this part.

front/src/config/i18n/en.json
```diff
@@ -253,6 +253,51 @@
...
+        "minPlaceholder": "Enter feature minimum value",
+        "maxLabel": "Maximum value",
+        "maxPlaceholder": "Enter feature maximum value",
+        "addButton": "Add feature",
+        "deleteLabel": "Delete feature"
+      },
+      "setup": {
+        "title": "MQTT configuration",
+        "urlLabel": "Broker URL",
+        "urlPlaceholder": "Ex: http://[mqtt-broker-address]:[port]",
```
Pierre-Gilles: It's not really how a MQTT address looks like, no? More like mqtt:// ?

front/src/config/i18n/en.json
```diff
@@ -253,6 +253,51 @@
...
+        "userLabel": "Username",
+        "userPlaceholder": "Enter server username",
```
Pierre-Gilles: Not really the server username, more the MQTT username ?

front/src/config/i18n/en.json
```diff
@@ -253,6 +253,51 @@
...
+        "passwordLabel": "Password",
+        "passwordPlaceholder": "Enter server password",
```
Pierre-Gilles: Same !

server/services/mqtt/lib/handler/devicestate.js
```diff
@@ -0,0 +1,22 @@
...
+ * @description Handle a new message receive in MQTT.
+ * @param {Object} mqttHandler - The Gladys MQTT client.
+ * @param {Object} message - The message sent.
+ * @example
+ * devicestate(gladys.service.mqtt.client, { identifier: 'deviceExternalId', type: 'temp', state: {value: 30} });
+ */
+function devicestate(mqttHandler, message) {
+  // Prepare new state event
+  const event = {
+    device_feature_external_id: `${message.identifier}:${message.type}`,
```
Pierre-Gilles: I'm not sure this is working, is it?

Here you ask for a "Feature MQTT message key" (not sure if it's understandable..)

Screenshot 2019-09-10 at 08 37 25

And then here, you use "identifier" + "type" 

The type has nothing to do here !

We should just have an identifier like "mqtt:${A_UNIQUE_KEY}" , we let the user choose the second part, and add the `mqtt:` each time (let's not show it on the front side, it's just internal stuff). We could even pre-generate the external_id with the name slugified.

On save, we need to make sure the unique error is handled in case the user use an identifier which already exist, so that a clear message is displayed.

front/src/routes/integration/all/mqtt/setup-page/actions.js
```diff
@@ -0,0 +1,67 @@
...
+        connectMqttStatus: RequestStatus.Getting
+      });
+      try {
+        await state.httpClient.post('/api/v1/service/mqtt/variable/MQTT_URL', {
+          value: state.mqttURL
+        });
+        await state.httpClient.post('/api/v1/service/mqtt/variable/MQTT_USERNAME', {
+          value: state.mqttUsername
+        });
+        if (state.passwordChanges) {
```
Pierre-Gilles, line 48: Why do you save everything all the time, and the password only if it changes? Just asking.

Pierre-Gilles (changes requested):
Hi! Back from holidays, I just had a look at your PR :) 

First, this one is really great, congrats on the work done here, it's huge and I like the approach you took 🙂

I wrote a few feedbacks and asked a few questions in comments about things I didn't understand.

front/src/routes/integration/all/mqtt/setup-page/actions.js
atrovato: To not retrieve the real one in a "clear" string, and to get a ***** string in password field to tell the user that a password is set.

front/src/routes/integration/all/mqtt/setup-page/actions.js
atrovato: Because I do not retrieve the real one from param table, so in the field, the value will be FAKE_PASSWORD by default, and in don't want to save it as is.

server/services/mqtt/lib/handler/devicestate.js
atrovato: The devicestate function is only for backward compatibility with Gladys V3. I have some devices publishing through MQTT with old device state format, I didn't want to change the code of this device. But it's like a proposal. We can force everybody to rewrite MQTT device code to work with V4.

server/services/mqtt/lib/handler/handleGladysMessage.js
atrovato: Ok, so the comment on devicestate function is not relevant, I will remove this part and force people to rewrite their device code.

front/src/config/i18n/en.json
atrovato: You're right :)

front/src/routes/integration/all/mqtt/setup-page/actions.js
Pierre-Gilles: Got it! The problem is that, if the user password is "FAKE_PASSWORD", it messes up everything right ?

We can maybe use a Symbol instead ( https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Symbol ) to avoid using a string const :)

server/services/mqtt/lib/handler/devicestate.js
Pierre-Gilles: I think it's much more complex to provide backward compatibility! The body asked in JSON message are not the same. Let's focus on Gladys 4 I would say, it's not really consistent to handle old messages

Pierre-Gilles:
I just tested locally all this, and I'm impressed this is really great !! :)

I'm going to help you on this today and I'll push improvements to your PR. 

Here is the todo list:

- [x] Changing a MQTT device name doesn't work ( updateDeviceProperty function in device-page/setup/index.js has wrong parameters)

- [x] Bug in the way Gladys get an existing device by selector. Should save the raw device not the sequelize object (this is a Gladys bug, my fault)
- [x] Add a websocket message to indicate that the MQTT broker was connected with success.
- [x] Add a mandatory "mqtt:" external_id. Pre-composed based on name, can be changed but not the prefix.
- [x] Bug in the way Gladys save device: when a feature is removed, the feature is not removed on the DB side. (Delete feature not coded on Gladys side yet)
- [x] Remove Gladys 3 compatibility
- [x] Add message on Device creation page to explain how it works and link to Gladys MQTT documentation ( https://documentation.gladysassistant.com/fr/api#mqtt-api )
- [x] Add a complete message on the setup page explaining the goal of MQTT + with a message linking to the Gladys MQTT Doc. 
- [x] Fix all previously noted typo/bugs

- [x] Verify that a message is displayed when unique device_feature_external_id constraints fails

- [x] Add a way to specify unit of device feature with a select box (by type of feature)

- [x] Change service image to remove Owntracks

front/src/routes/integration/all/mqtt/setup-page/actions.js
Pierre-Gilles: Never mind, I executed the code and now I understand. I didn't understood what you were doing here.

atrovato:
Wow !!! You've done all todos !??

Pierre-Gilles:
Yes! Worked all day on MQTT :) 

Can you have a look and see if you have any bugs/feedbacks ?

Something I didn't manage to do: if you enter a wrong mqtt URL, it seems we never receive any "cannot connect" error. Right now I added 2 popups: A blue one written "Connecting..." followed by a green one "connected with success". I would like to have a red one when the connection failed (I did it but it doesn't work). Maybe it's a question of timeout, the doc writes that the timeout is 30 seconds... on my side I see nothing. 

I will test more deeply tomorrow, and it everything seems good I think we can merge this as a first MQTT service :)

Pierre-Gilles (approved):

PR #552, opened 2019-10-12
[WIP] Broadlink service

atrovato:
### Pull Request check-list

- ~~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~
- ~~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).~~

### Description of change

Handle Broadlink WiFi IR devices and smart plugs.


Pierre-Gilles:
Hi! 

Thanks for this great PR 👏 

I began looking at the code/UI, it looks great but I think I'm missing some informations to really understand how this works/what it does ! 😄

Could you pitch me in a few lines what this service does? What kind of peripherals can be controlled?

Maybe the service itself should explain a little be more how this works, because for now I'm a little lost ^^

Example, I went on the UI and saw this:

Screenshot 2019-12-23 at 14 32 42

As a user, I have no idea what it is: 

- What is "R", "G", "B" buttons ? Should I click on buttons?
- Why is there 4 "-" buttons? 
- Why is there a "play" button in black and white gradient?

In the code, I saw that you are adding a lot of type for the "Light" feature, is this service related to controlling lights? If yes, why the control is different from others lights like Philips Hue? 

Thanks again for your work, I'm looking forward to better understand this 😉

atrovato:
Hi, thans for looking.
Remotes are designed for IR devices, which expect to get a specific code by IR.
So codes should be saved as key / button, then an action on it will call the `setValue` method according to the service.

And as it is IR and specific codes, we can't compute the code.
For example, for lights, the hue color UI is not appropriate, there os too many choices. My device is 16-color led band, and it knows only this predefined colors.

R/G/R are red/green/blue buttons for colors, and to not surcharge the screen, I let "-" for other colors.
If you mouseover each button, you will get a tooltip.
Please take a look at https://www.google.com/search?q=remote+16+color&oq=remote+16+color&aqs=chrome..69i57j0l2.3899j0j4&client=ms-android-samsung-ss&sourceid=chrome-mobile&ie=UTF-8#imgrc=SVYo6AAmuD3YpM:

On the fabulous remote, there is also 4 "scenes" represented by the colored butons and the black and white one.
You can have flash / strombe / white to dark / color smooth "scenes".

On the dashboard, some butons are disabled, this means that the control is not configured for this buton, not saved from "broadlink" service.

Pierre-Gilles:
Mmm ok I see, thanks for the explanation 😃 

I'm still not convinced by the UI to control lights, it's not user friendly. I didn't understand what it was for (and I'm still not sure of every button), and I'm convinced nobody in my family would be able to understand what it is

I understand that is way different from Philips Hue which have a wide range of color (16 millions colors vs 16 colors), but I think we can do the same thing we wants to for Philips Hue for this control.

For Philips Hue, I don't want to give access to the 16 millions colors available: nobody needs that! We'll put a color picker of a subset of "nice" colors ( See my message [here](https://community.gladysassistant.com/t/parlons-de-gladys-v4/3578/661?u=pierre-gilles) )

I think we need to apply the same philosophy here, and treat those lights as normal lights, not create a special box, what do you think of that ?

**From what I see, those lights have 4 features:** 

- On/Off
- Brightness
- Colors (from a subset of 16 colors)
- Special Mode (Flash/Strobe/Fade/Smooth)

What do you think of that?

atrovato:
It can be great, but I'll need to get this box to adapt the remote configurator.

And what about the TV remote ? Does it look fine for you ?

Pierre-Gilles:
It looks fine but it's big on the dashboard, no? Maybe we could have a compressed view somehow? 

Let's try to understand: what does people want to do with this remote control? Most of the time you just want to:

- Turn On/Off the TV
- Change Volume
- Change channel

Right? 

Maybe a compressed view (expandable) can be enough, no?

atrovato:
Hi,
I just remember that remotes are only push-button with no feedback state.
So, we need to create push button features to switch on / off, and sometimes, one for on, one for off...
So I'm not sure that it will be easy to reuse light graphical component.

PR #602, opened 2019-11-09
Fix zWave port listing and binding

sescandell:
### Pull Request check-list

- [x] ~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~
- [x] ~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).~
- [x] ~Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~

### Description of change

The request to list available ports to select the zWave USB stick remains unfinished.
This is because of a BC in SerialPort: the list method is now returning a Promise and doesn't support the callback parameter. See here: https://github.com/serialport/node-serialport/blob/master/UPGRADE_GUIDE.md#upgrading-from-7x-to-8x

In the BC, the `comName` field is now named `path`.

By the way, this PR also fix an issue when selecting a value in the selector and clicking "Connect": the value sent to the server was an empty string.

### DIFF
--- a/front/src/routes/integration/all/zwave/settings-page/SettingsTab.jsx
+++ b/front/src/routes/integration/all/zwave/settings-page/SettingsTab.jsx
@@ -26,7 +26,7 @@ const SettingsTab = ({ children, ...props }) => (
               </option>
               {props.usbPorts &&
                 props.usbPorts.map(usbPort => (
-                  <option selected={props.zwaveDriverPath === usbPort.comName}>{usbPort.comName}</option>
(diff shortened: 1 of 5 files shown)


Pierre-Gilles:
Many thanks for this PR, this was indeed the bug!

PR #608, opened 2019-11-14
Some fixes: zWave device integration and BinaryDeviceFeature component

sescandell:
### Pull Request check-list

- ~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).~
- ~[ ] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~

### Description of change

This PR initially was intended to fix the Device.setValue call for zWave devices. But, making tests with a real BinarySwitch device I identified another issue with the BinarySwitch front component and the Gateway forwarder.

Changes:

 * Add the zWaveManager instance as the device property of zwave service (following the PhilipsHueService and its handler scheme)
 * An error on device feature test about setValue feedback
 * BinaryDeviceFeature: a mix between bool and int making impossible to Switch On then Switch Off the device
 * Do not try to Forward messages to the Gateway when not connected to it

### DIFF
--- a/front/src/components/boxs/device-in-room/device-features/BinaryDeviceFeature.jsx
+++ b/front/src/components/boxs/device-in-room/device-features/BinaryDeviceFeature.jsx
@@ -7,7 +7,7 @@ const BinaryDeviceType = ({ children, ...props }) => {
       props.deviceFeature,
       props.deviceIndex,
       props.deviceFeatureIndex,
-      props.deviceFeature.last_value === 1 ? 0 : 1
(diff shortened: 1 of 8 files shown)


Pierre-Gilles:
Thanks for this great PR, tested on merged! 👍 👏

Issue #614, opened 2019-11-19
zWave Version discussion


sescandell:
Hi,

I'm opening this issue in order to firstly open the discussion before going on any modification.
There is an issue with how "openzwave lib is installed". I'm taking here Docker as a sample. But, I imagine, the same goes for the raspberian version (even if for this one I didn't yet dive into the workflow to generate the image).

There is 2 issues from my POV with the openzwave installation in Docker:

 1. the version installed is the 1.4.164: that is... more than outdated (released around... April 2016 if I'm not wrong... based on [this package](http://old.openzwave.com/downloads/openzwave-1.4.164.tar.gz) and the same Changelog that can be found on [this Commit](https://github.com/OpenZWave/open-zwave/blob/23a6ff0f4f0401fdb4d2d557f5006886d3234fd3/ChangeLog) but not [that one made just after](https://github.com/OpenZWave/open-zwave/commit/78ad4ed699976fb34d9c7239f210853e16f2c507#diff-02f0b547c2779d25cff89672135f20e3) )
 1. the OZW database is so really outdated and most of the devices are unknown (this is the most important issue in my opinion)

I see two ways to fix this:

 1. use a manual install through a `git clone && make && make install` into the build script based on a known version from Github
 1. stay with the `apk add openzwave` installation but manually update the device database (said simply by cloning the repo then copy/pasting the config folder)

Issue with first solution: if I'm not wrong, version 1.6.X is not fully supported by node-openzwave-share. That being said, bases on some simple tests on a custom installation, it seems to be working so far... But I didn't (yet) tested all Alarm/Notification/ControlScene features (features that might need changes...)

With the second solution, I have no idea if there might be any "new feature" that could be referenced in the device database that would not be supported by the version 1.4...

I can work on these changes. But before working on a PR that is not aligned with your vision, I prefer to open this issue and discuss with you about this.

Waiting for your feedbacks

Pierre-Gilles:
Hey! Thanks for your message :) 

There are 2 topics here: 

- The "Unknown device" problem on the Docker image
- Installed Open-Zwave version in the Docker image

**Unknown device problem**

This is not related to the version of openzwave, there was a bug in the Docker image that I fixed yesterday in this commit => https://github.com/GladysAssistant/Gladys/commit/5685b452d01aed1e4b2abeda62980f624509a773 (haven't fully tested yet so i didn't communicate on this yet)

[...]

PR #617, opened 2019-11-21
Import OpenZwave DB from Github

sescandell:
### Description of change

Following the 2 comments in #614 here is a first PR about upgrading the OpenZwave DB but staying on the 1.4 OpenZwave version.
I don't think it would impact anything in code... just newer and recent devices will now be recognized.

Another PR is coming (tomorrow or after) to move on OpenZWave 1.6 (or not...);

I don't know where your script to generate the RPI Image reside. Let me know so I can update it too.

Thanks!

### DIFF
--- a/.circleci/build-image.sh
+++ b/.circleci/build-image.sh
@@ -15,6 +15,12 @@ fi
 # </qemu-support>
 # ============
 
+# ============
+# <zwave-db>
+git clone https://github.com/OpenZWave/open-zwave.git open-zwave
+# </zwave-db>
+# ============
+
 # Replace the repo's Dockerfile with our own.
 docker build -f ./docker/Dockerfile \
   -t ${IMAGE_ID} \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,6 +11,10 @@ COPY qemu-* /usr/bin/
 # System dependencies
 RUN apk add --no-cache tzdata nmap openzwave-dev openzwave-libs openzwave ffmpeg sqlite openssl gzip
 
+# Update Open-zWave DB
+RUN rm -rf /etc/openzwave/* 
+COPY open-zwave/config/* /etc/openzwave/
+
 RUN mkdir /src
 WORKDIR /src
 ADD . /src


Pierre-Gilles:
Thanks for this PR! Did you test it? 

This is a little hacky (imagine if the XML format changes between 2 versions, it'll break our image) so I would prefer switching directly to 1.6 if it works! Waiting for your other PR so :)

sescandell:
> Thanks for this PR! Did you test it?

Yes... Well, basically but yes. I run the Docker with a custom volume doing the same thing as in this PR. I can validate that now, one of my device is well recognized (I can provide you screenshot if you want).

 > (imagine if the XML format changes between 2 versions, it'll break our image)

You totally right! I didn't thought about that. We can force a given commit (or a tag or...) to avoid that issue and be in control about the changes.

I'll do PR about the 1.6 update.

Thanks for your time!

sescandell:
Waiting for the other PR, I'm closing that one with a "local solution" for users whom would want to use the latest OpenZwave DB without affecting everybody.

Simply make the same manipulation on your machine and use the cloned repo as a mounted volume. In terms of commands:

```bash
# Clone the repo
git clone https://github.com/OpenZWave/open-zwave.git openzwave

# Run the Docker image with an additional volume (latest -v option here):
docker run -d \
    --restart=always \
    --privileged \
    --network=host \
    --name gladys \
    -e NODE_ENV=production \
    -e SERVER_PORT=80 \
    -e TZ=Europe/Paris \
    -e SQLITE_FILE_PATH=/var/lib/gladysassistant/001/gladys-production.db \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /var/lib/gladysassistant/001:/var/lib/gladysassistant/001 \
    -v /dev:/dev \
    -v `pwd`/openzwave/config/:/etc/openzwave/ \
    gladysassistant/gladys:4.0.0-beta-arm
```

That way you can easily test it.

As soon as I have time (next week) I'll propose a PR for an full update to the 1.6

Thanks,

PR #647, opened 2020-01-06
Mi-light service

link39:
### Pull Request check-list

- [x] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).

### Description of change

Draft to add a mi-light service in Gladys v4. 
 
- Connect/disconnect bridge
- Add/Remove mi-light zone
- Change state of zone (On/Off)

image

image

image

image


server/services/mi-light/index.js
```diff
@@ -0,0 +1,38 @@
+const logger = require('../../utils/logger');
+const MiLightLightHandler = require('./lib/light');
+const MilightController = require('./api/milight.controller');
+
+module.exports = function MiLightService(gladys, serviceId) {
+  // require the node-milight-api module
+  // @ts-ignore
```
atrovato: Why ignoring this ?

server/services/mi-light/lib/light/index.js
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ * const exampleLightHandler = new ExampleLightHandler(gladys, client, serviceId);
+ */
+const MiLightLightHandler = function MiLightLightHandler(gladys, milightClient, serviceId) {
+  this.gladys = gladys;
+  this.milightClient = milightClient;
+  this.serviceId = serviceId;
+  this.bridges = [];
+  this.connectedBridges = [];
+  this.bridgesByMac = new Map();
```
atrovato, line 21: Are you sure to need a Map object ? Simple (JSON) object is not enough ?

server/services/mi-light/lib/light/light.configureBridge.js
```diff
@@ -0,0 +1,66 @@
...
+ * @description Configure the milight bridge.
+ * @param {string} Mac - Mac Address of the Mi-Light Bridge.
+ * @example
+ * configureBridge('F0:FE:6B:A6:03:EB');
+ */
+async function configureBridge(Mac) {
+  const bridge = this.bridgesByMac.get(Mac);
+  if (!bridge) {
+    throw new NotFoundError(`BRIDGE_NOT_FOUND`);
+  }
```
atrovato, line 22: Is it really the expected error message ?

server/services/mi-light/lib/light/light.configureBridge.js
```diff
@@ -0,0 +1,66 @@
...
+  logger.info(`Connecting to milight bridge "${Mac}", ip = ${bridge.ip}, name = ${bridge.name}`);
+  try {
+    const bridgeMacWithoutColomn = Mac.replace(/:/g, '');
+    const deviceCreated = await this.gladys.device.create({
```
atrovato, line 26: Can't you create device from front ? Calling the create devie API.

server/services/mi-light/lib/light/light.setValue.js
```diff
@@ -0,0 +1,60 @@
...
+  for (let i = 0; i < 11; i += 2) {
+    macAddressStdFormat += bridgeMac.slice(i, i + 2);
+    if (i !== 10) {
+      macAddressStdFormat += ':';
+    }
+  }
+  const bridge = this.bridgesByMac.get(macAddressStdFormat.trim());
+
+  if (!bridge) {
+    throw new NotFoundError(`BRIDGE_NOT_FOUND`);
```
atrovato: Please review error message.

server/services/mi-light/lib/light/light.setValue.js
```diff
@@ -0,0 +1,60 @@
...
+    `Connecting to milight bridge : Mac = "${bridgeMac}", ip = ${bridge.ip}, type = ${bridge.type}, name = ${bridge.name}`,
+  );
+  logger.debug(
+    `Changing state of light ${device.external_id} with IP ${bridge.ip} . Zone = ${zoneId}. New value = ${value}`,
+  );
+
+  // const commands = await this.milightClient.commandsV6;
+  // const MiLight = await this.milightClient.MilightController;
+  const commands = require('node-milight-promise').commandsV6;
+  const MiLight = require('node-milight-promise').MilightController;
```
atrovato: Can't it be required at the top of the file ?

atrovato (changes requested):
Please find here first (light) review comments.
Please revert .gitignore and root package-lock.json files, as you didn't change it.

server/services/mi-light/lib/light/light.setValue.js
link39: I will change the error message : `"No bridge saved in Gladys with the following mac address : ${macAddressStdFormat.trim()}"`

server/services/mi-light/lib/light/light.configureBridge.js
link39: I will be change with a clearer error message : No bridge discovered on the network with the following mac address : ${Mac}

server/services/mi-light/lib/light/light.configureBridge.js
link39: I used the phillips hue as example. Pierre-Gilles did it like this, so I don't know what is the good way to create the device.

server/services/mi-light/lib/light/light.setValue.js
link39: Yes I think. I move it to top

server/services/mi-light/lib/light/index.js
link39: Is it a big deal ? Because I'm not sure to understand the difference between these two object types.

atrovato: @Pierre-Gilles any opinion ?

atrovato (approved):

server/services/mi-light/lib/light/index.js
Pierre-Gilles: Map is really great! We should really use more of them, it's way better and safer that simple JSON objects, because it provides safe methods to check properties existence for ex.

server/services/mi-light/lib/light/light.configureBridge.js
```diff
@@ -0,0 +1,66 @@
...
+  BRIDGE_TYPE,
+} = require('../utils/consts');
+
+/**
+ * @description Configure the milight bridge.
+ * @param {string} Mac - Mac Address of the Mi-Light Bridge.
+ * @example
+ * configureBridge('F0:FE:6B:A6:03:EB');
+ */
+async function configureBridge(Mac) {
```
Pierre-Gilles, line 18: We use camelCase to name variables, so "Mac" should be "mac".

I'm surprised that ESLint is not checking this... I'm going to add this to eslint !

server/services/mi-light/lib/light/light.configureBridge.js
Pierre-Gilles: This function is only called from the front, so it's the same thing :) 

The thing I want to avoid, is to have "background creation of device" => we don't want that, because it can results in hard to understand errors.

Here, this route is called by the front, so if it fails, we can display an error message, so it's ok.

server/services/mi-light/lib/light/light.setValue.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  logger.info(
+    `Connecting to milight bridge : Mac = "${bridgeMac}", ip = ${bridge.ip}, type = ${bridge.type}, name = ${bridge.name}`,
+  );
+  logger.debug(
+    `Changing state of light ${device.external_id} with IP ${bridge.ip} . Zone = ${zoneId}. New value = ${value}`,
+  );
+
+  // const commands = await this.milightClient.commandsV6;
+  // const MiLight = await this.milightClient.MilightController;
```
Pierre-Gilles: Please remove dead code

server/services/mi-light/lib/light/light.setValue.js
Pierre-Gilles: Those require should be in the `index.js` file of the service and injected in the handler class.

Example: 

https://github.com/GladysAssistant/Gladys/blob/master/server/services/philips-hue/index.js#L8

server/services/mi-light/lib/light/light.setValue.js
```diff
@@ -0,0 +1,60 @@
...
+  switch (deviceFeature.type) {
+    case DEVICE_FEATURE_TYPES.LIGHT.BINARY:
+      value ? light.sendCommands(commands.rgbw.on(zoneId)) : light.sendCommands(commands.rgbw.off(zoneId));
+      break;
+    default:
+      logger.debug(`Mi Light : Feature type = "${deviceFeature.type}" not handled`);
+      break;
+  }
+
+  light.pause(100);
```
Pierre-Gilles, line 51: Why light.pause(100) ? Just asking 😄

server/services/mi-light/lib/light/light.setValue.js
```diff
@@ -0,0 +1,60 @@
...
+  return light.close();
```
Pierre-Gilles, line 52: Why light.close? Should we open/close connection on each light control? Couldn't the light object saved somewhere? Won't it hurt performance to open/close each time?

Pierre-Gilles (changes requested):
Hey! 

Thanks for your PR :) 

It's really great, I just had a few feedbacks on the code itself. 

One other more general feedback, and I realize it's not your fault but mine : There are no "empty state" on this integration.

When I arrive on this integration, the thing I see is this: 

Screenshot 2020-03-28 at 10 51 00

The user is lost: what should I do? 

I know that this code is based on Philips Hue, and Philips Hue is no better, but we need to improve both.

The user should know what to do ! 

Could you add empty states? I'll do it on Philips Hue on my side. 

There should be some text explaining:

- Why it's empty
- What to do next

server/services/mi-light/lib/light/light.setValue.js
link39: It works like this in each example of the "node-milight-promise". I thinks there is no performance impact.

server/services/mi-light/lib/light/light.setValue.js
link39: Why light.pause(100) ? --> Not necessary ^^ 

Concerning requires, is it okay if I let a global require in index.js : 
`const milightClient = require('node-milight-promise');`

and I put this in setvalue.js : 

```
const commands = await this.milightClient.commandsV6;
const MiLight = await this.milightClient.MilightController;
```
?

Pierre-Gilles: Yes it's good! Why do you need await here btw? 

Sorry for the slow answer, I have literally thousands of pending messages and it takes me hours to read+reply on everything 😁

server/services/mi-light/lib/light/light.setValue.js
Pierre-Gilles: Ok!

PR #643, opened 2020-01-03
CI and 🐳 improvement

VonOx:
Resolves #640, resolves #641 and #635 

### Description of change

Building docker image and manifest

On master branch image is `gladys:dev`
On tag image is `gladys:latest` `gladys:vX` and `gladys:vX.Y.Z`

PR #675, opened 2020-03-04
Update OZW to 1.6 and use the same version in all context

sescandell:
### Pull Request check-list

- [ ] Is the linter passing? (`npm run eslint` on both front/server)
- [ ] Did you run prettier? (`npm run prettier` on both front/server)
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).

### Description of change

As explained in https://github.com/GladysAssistant/Gladys/issues/614#issuecomment-593143583 image used in Docker reference OZW 1.6.

Let's try to make CircleCI in sync with the "production" environment and use node-openzwave-share version 1.6 (to stay sync regarding the OZW version).


sescandell:
Hi @Pierre-Gilles 

With these limited changes:

 - the docker container used for tests is now using OZW 1.6 with the same revision as the alpine3.11 (revision 974... as far as I think...)
 - node library has been updated to the 1.6.2
 - alpine version has been forced so Gladys stay in control about the OZW version running in prod (with Docker at least... don't know how your RPI image is generated and what should I change to "force" a specific version for OZW)

All tests are green. BUT it doesn't necessary means everything is working. It simply means that all interfaces with the lib' are okay. If anything has changed inside OZW, it will not appear in the current tests.

That being said, the OZW v. 1.6.974 is 3 months old. As you can see, I have to manually checkout and compile the lib'. 3 months is a lot. Many fixes have been made on the lib' since then. I would like to use a newer version by using, let's say one of the latest commit ID. And doing this in the Docker container used for tests AND in the Dockerfile used for generated images (so removing the `apk add openzwave ...` and replacing it with a custom install like in the test container). Before doing this, I would like your feedbacks? Since updating to the 1.6 version, we might as well use the latest version available. What do you think? Should I checkout a newer version or stay with this one? (my preference is for a new one...)

As soon as possible, I will try to connect my ZWave key and see if things are working with a real device and real events (but I have a very limited set of devices to test "everything"). If you want to check on your side with your devices if everything is working, that would be great!

Thanks,

Pierre-Gilles:
Hi @sescandell ! Thanks for this great PR ! :) 

> Before doing this, I would like your feedbacks? Since updating to the 1.6 version, we might as well use the latest version available. 

I'm ok with using a newer version, but we really need to lock the version used in the Dockerfile, because we don't want to have unwanted side effects when building at a random time if one OZW guy pushed some buggy code on master at the same time (it happens a lot with OZW). 

Do you know how they work? They don't seem to use tags frequently on their repo...

I don't understand how they version/release their lib

sescandell:
> we really need to lock the version used in the Dockerfile

Totally agree with this!

> Do you know how they work? They don't seem to use tags frequently on their repo...

Unfortunately, I don't. I'm not even sure they have any rule. They have a milestone for the 1.7 version, but nothing very clear about patches or fixes for a given version. Looking at Alpine distribution, I think they just compile "one version" at the time of updating...

Let's try with the latest one. Then we would have to follow the OZW repo to update it "when we think it deserve it".

Update of the PR coming.

sescandell:
Hi @Pierre-Gilles 

PR updated. Forced on a given commit. Dockerfile and CircleCI updated.

I just don't have any idea how you generate the RPI image and if I should change anything for it.
Do not hesitate to tell me if something else must be updated.

I didn't yet tested this on a real zWave Controller. I should be able to do it in the coming days. If you have any chance to test it, let me know.

sescandell:
I've made some very simple tests: the zWave doesn't initialize as expected.

I'll dive into the changes of the NPM lib in the coming days,

Still not ready to merge for now.

Pierre-Gilles:
Hey ! I saw you pushed more things, is it ready for review? :) 

Thanks for your great work !

sescandell:
Hi @Pierre-Gilles ,

I tested yesterday to run Gladys with OZW on a MacOSX and it worked (well... actually the communication with the OZW library worked... I had other issues but I think they are not tied to the update).

I'm now going to test the same thing in Docker and let you know.

By the way, @Pierre-Gilles  could you please precise your environment? What's your OS? How do you develop Gladys? Which tools? How do you test ZWave integration? How do you debug? What's your IDE? 


Thanks,

sescandell:
Well, actually I have issues to test the Docker environment as you cannot mount an USB device into the Docker container.
It's working on the Mac. If you want more tests, I'll try to reconfigure another computer on a Linux environment and test from there.

Do not hesitate if you have any question,

Albenss:
I don't understand why you cannot mount an USB device into Docker, have you tried to use the volume section to mount the device directly? 
Like it is used in Zigbee2Mqtt (https://www.zigbee2mqtt.io/information/docker.html#running).

VonOx:
Without your Docker run command, it's hard to help.
You need to launch with privileged and mount dev, like this.
```
docker run -d \
--restart=always \
--privileged \
--network=host \
--name gladys \
-e NODE_ENV=production \
-e SERVER_PORT=80 \
-e TZ=Europe/Paris \
-e SQLITE_FILE_PATH=/var/lib/gladysassistant/gladys-production.db \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /var/lib/gladysassistant:/var/lib/gladysassistant \
-v /dev:/dev \
gladysassistant/gladys:4.0.0-beta-arm
```

sescandell:
I already tried the following, but none of them is working:

```bash
docker run -d \
--restart=always \
--privileged=true \
--name gladys \
-e NODE_ENV=production \
-e SERVER_PORT=80 \
-e TZ=Europe/Paris \
-e SQLITE_FILE_PATH=/var/lib/gladysassistant/gladys-production.db \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /Users/Shared/Gladys:/var/lib/gladysassistant \
--device=/dev/tty.usbmodem14201 \
-p 8080:80 \
gladystest
```

```bash
docker run -d \
--restart=always \
--privileged=true \
--name gladys \
-e NODE_ENV=production \
-e SERVER_PORT=80 \
-e TZ=Europe/Paris \
-e SQLITE_FILE_PATH=/var/lib/gladysassistant/gladys-production.db \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /Users/Shared/Gladys:/var/lib/gladysassistant \
-v /dev:/dev \
-p 8080:80 \
gladystest
```

Are you working with OSX (10.15.4) on your side too?

VonOx:
Windows and Linux most of the time

Access USB stick on Mac is a bit tricky, I saw this article https://christopherjmcclellan.wordpress.com/2019/04/21/using-usb-with-docker-for-mac/

sescandell:
Actually, it seems this is not supported by Docker on Mac OSX at all... See https://github.com/docker/for-mac/issues/900

Otherwise, I need to falback on the boot2docker and VM customization... Going back to Windows / Linux

sescandell:
@VonOx 

> Windows and Linux most of the time

Should I understand USB share in Docker is supported on Windows host?

VonOx:
Sorry I don't know, I'm not using USB share on Windows.

sescandell:
I've made some additional tests on my Mac with a zWave controller playing with a binary device: and it works pretty well.

Appart from this, I have a Walli Dimmer Fibaro device, and Dimmer are not working at all. But I will propose a fix into another PR, this is not the aim of that one.

sescandell:
@Pierre-Gilles not sure if you followed everything here, but it is ready to merge

docker/Dockerfile
```diff
@@ -9,16 +9,29 @@ LABEL \
 COPY qemu-* /usr/bin/
 
 # System dependencies
-RUN apk add --no-cache tzdata nmap openzwave-dev openzwave-libs openzwave ffmpeg sqlite openssl gzip
+RUN apk add --no-cache tzdata nmap ffmpeg sqlite openssl gzip
 
+WORKDIR /tmp
+
+# Install OZW in a dedicated RUN to benefit from Docker cache on local dev'
+RUN apk add --no-cache --virtual .build-deps make gcc g++ python libffi-dev linux-headers git udev coreutils \
```
Pierre-Gilles: Be careful, here the build deps are going to be included in the final image! 

Why? Because all docker layers are cached. 

So here, the `build-deps` are installed, and removed on a separate RUN statement, meaning that they won't be removed from the previous layer. 

Please roll back to the previous behavior :) You can maybe do a RUN statement dedicated to the installation of open-wave

front/src/components/boxs/device-in-room/device-features/BinaryDeviceFeature.jsx
```diff
@@ -7,7 +7,7 @@ const BinaryDeviceType = ({ children, ...props }) => {
       props.deviceFeature,
       props.deviceIndex,
       props.deviceFeatureIndex,
-      props.deviceFeature.last_value === 0 ? 1 : 0
```
Pierre-Gilles: This is not linked to openzwave, please remove that from this PR :)

Pierre-Gilles (changes requested):
Thanks for this PR ! :)

This is really nice and it'll help many people. Congrats 👏 

I noted some minor changes to your PR, can you have a look?

sescandell:
@Pierre-Gilles all fixed

docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG target
 
-FROM $target/node:12-alpine
+FROM $target/node:12-alpine3.11
```
Pierre-Gilles: Just one last question, why can't we use the latest node:12-alpine image? (just asking)

Pierre-Gilles:
Many thanks for the PR!! :) Except my last question, I'm good to go!

docker/Dockerfile
sescandell: Looking at the commits history it was to fix the OZW version without using Git. But then, we decided to move on manual compilation on a given commit.

So this constraint is not needed anymore.

I'll remove it.

sescandell: Done

Pierre-Gilles:
@VonOx How can we merge both your changes on Docker + this PR ? Is your PR compatible with that ?

VonOx:
@Pierre-Gilles I think so

PR #643 , the only modification on Dockerfile is base image ( node-lts from node-12 )
But merge #675 before, i will rebase my PR

PR #793  use independent Dockerfile ( buildx ) , i will implement OZW part as @sescandell did before you merge.

PR #709, opened 2020-03-19
Fix #576 Map - Fit zoom to markers

### LINKED ISSUE #576: Maps zoom level can be a little more zoomed when a user is located.
- [ ] If several users are located, make sure all users are visible. Zoom level should adapt accordingly

VonOx:
### Pull Request check-list

### Description of change

Resolves #576

Leaflet map now use `fitBounds` to auto zoom on markers.
Only active if marker >= 1

### DIFF
--- a/front/src/routes/map/Map.jsx
+++ b/front/src/routes/map/Map.jsx
@@ -89,7 +89,7 @@ class MapComponent extends Component {
     this.displayUsers();
     if (markerArray.length >= 1) {
     var group = leaflet.featureGroup(markerArray);
-    this.leafletMap.fitBounds(group.getBounds(), {padding: [25, 25]});
+    this.leafletMap.fitBounds(group.getBounds(), { padding: [25, 25] });
     }
   }
 
--- a/front/src/routes/map/Map.jsx
+++ b/front/src/routes/map/Map.jsx
@@ -42,7 +42,7 @@ class MapComponent extends Component {
               zIndexOffset: 1000
             })
             .addTo(this.leafletMap);
-            markerArray.push(this.userMarkers[user.id]);
+          markerArray.push(this.userMarkers[user.id]);
         }
       });
     }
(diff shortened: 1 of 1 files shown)


front/src/routes/map/Map.jsx
```diff
@@ -4,6 +4,7 @@ import 'leaflet/dist/leaflet.css';
 import style from './style.css';
 
 const DEFAULT_COORDS = [48.8583, 2.2945];
+var markerArray = [];
```
Pierre-Gilles, line 7: Warning, here you're creating a global variable ! ⚠️ 

DEFAULT_COORDS is global because it's a const that will never change.

Any variable which is modified during the lifecycle of an object should be in the component.

front/src/routes/map/Map.jsx
```diff
@@ -84,6 +87,10 @@ class MapComponent extends Component {
   componentDidUpdate() {
     this.displayHouses();
     this.displayUsers();
+    if (markerArray.length >= 1) {
+      var group = leaflet.featureGroup(markerArray);
```
Pierre-Gilles, line 91: `var` should not be used anymore, here `const` need to be used. 

I just realized that eslint was not checking .jsx file..  my bad!

front/src/routes/map/Map.jsx
Pierre-Gilles: More info 

https://www.freecodecamp.org/news/where-do-i-belong-a-guide-to-saving-react-component-data-in-state-store-static-and-this-c49b335e2a00/

Pierre-Gilles (commented):
Thanks for your PR, this is really nice 👍 

It'll be merged in https://github.com/GladysAssistant/Gladys/pull/711

I had to make some changes to your PR, that I detailed here so you know for next time!

Pierre-Gilles:
Thanks for your PR ! 🙂

I just merged it in https://github.com/GladysAssistant/Gladys/commit/65624b65f61429deab2437547fc27bcafe8c2e3d !

VonOx:
Thanks for the feedback. I need to progress and theses comments are helpful

PR #731, opened 2020-04-04
Replace react-select

atrovato:
### Pull Request check-list

- ~~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).~~

### Description of change

Change react-select lib to downshift: https://downshift.netlify.com/

@Pierre-Gilles Do you think we need to change all "classical" selects to keep graphical consistency?


atrovato:
According to some user screenshot, the select input may sometimes keep the browser default skin, so I suggest to replace all original HTML select tags with the react rendered one.

And more, I will look to make build fail in case of HTML <select /> usage to enforce GUI consistency.

atrovato:
So I changed all `select` tag to a custom one, and forbid usage of orignal HTML `select` tag.
I suppose to have test every changes, but it would be better to get some testers to ensure there is no regression on select behaviors (specialy on Gladys Plus part).

Pierre-Gilles:
Very nice PR ! Thanks @atrovato ! I'll test this extensively before merging

Pierre-Gilles (changes requested):
Hi ! 

Thanks for this PR.

I just tried it, and I have some issues: 

Screenshot 2020-07-27 at 15 14 00
Screenshot 2020-07-27 at 15 14 25

Did I do something wrong?

atrovato:
Yes you did something wrong, you didn't try on fixed version ;)
I just updated it right now, I saw an issue on multi-select with groups (dashboard, devices in room), while trying to unselect.
I will have a look later.

atrovato:
Now fixed :)

Pierre-Gilles:
Merci pour cette PR ! :) Je regarde la semaine prochaine lors de mes jours Gladys.

Pierre-Gilles:
I just had a look at your PR, I know it's lot of work so first thanks for all of this! 

I'm questioning the superiority of downshift vs react-select.

Example for multiple-select:

**Downshift**

Screenshot 2020-08-24 at 10 26 14

**React-Select** 

Screenshot 2020-08-24 at 10 26 22

To me, react-select is more clear.

One other worry I have is the added complexity and maintenance. The `Select.jsx` file seems more complexe and is a library in itself, and means a new developer needs to learn to use `Select.jsx` every time he needs to put a select box, and that without documentation, where before he could just go to the `react-select` documentation to learn how to use react-select. 

On top of that, a newcomer will hesitate to modify the `Select.jsx` file as it could break somewhere else in the frontend (as it's used everywhere). So I really question if this PR will bring something to Gladys or make it harder to develop for other developers.

Don't take it personally I know this is lot of works and I appreciate the quality of the PR your deliver in general 🙂

atrovato:
You're totally right, it's not always clear to me too.

Why downshift? 
  - It has a quite better integration with any front css/components. React-select is limited to its own css, and it is not so easy to adapt it to our styles

So we can do the expected ui render with downshift too.

I agree, I created a lib to get our own select inputs, which is not really a good idea (too big file, no doc...)...

The main goal was to unify all select fields in the ui. On some screenshot, we saw select fields naked, no css, and it hurts me.
So that's why I wanted to force our own select box.

Forget it for now, it will try later with something lighter... I hope react-select will be bootstrap compatible.

Pierre-Gilles:
Thanks for your answer :) I don't 100% agree about banning native select box. 

Sure, on some part of Gladys, it looked bad because the list was too big (in this case we can migrate to react-select), but native select boxes are very fast and more adapted to each platform when it's just a small list. On mobile for example, native select box are really great and faster than JS select box. 

Example of the native select on iOS:

ios-2

> it will try later with something lighter

Sure, but don't spend too much time on this, I think we already have tons of features missing in Gladys 4 and it's better to spend time on features!

atrovato:
We'll see later.

PR #733, opened 2020-04-05
[TBD] Manage secret params and variables

atrovato:
### Pull Request check-list

- ~~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).~~

### Description of change

Ok, so I propose:
 - by default, never retrieve secret value in device params or variables
 - for variable, should explicitly call `getValue` with `secret = true` argument, and throws 403 from controller if request to get secret variable (but post in allowed to override value)
 - for device params, a `getParam` function has been added, as device never load secret value params
 - in both cases, if a value is stored as secret, it should stay as it until it is deleted, whatever comes from any update call


Pierre-Gilles:
Thanks for this PR ! 

I'm 100% agree that we need secret variables, I had a quick look at first at the code and it seems nice :) 

But about the MQTT, I'm not sure: Do we want to hide the MQTT_PASSWORD?

To me, the next step for the MQTT services is to add the auto-start of a Mosquitto container with a generated MQTT_PASSWORD, so it means the password will have to be displayed in the UI.

Sure, if we display the MQTT password once, it's super secure. But when the user will have to connect another connected thing, and has forgotten the MQTT password (like 99.99% of the time), what does he do? Delete the whole integration and re-configure everything? He'll just stop using Gladys..

There is always a balance between "ease-of-use" and security, and I question the usage of this feature for the MQTT service. 

What do you think?

atrovato:
Ok, I understand the needs when we will integrate the MQTT broker, which is not the case right now.
I can simply remove the MQTT changes for now, or wait for this integrated broker.
As you prefer.

Pierre-Gilles:
I think we'll need this in the future, but not right now.

This PR is great, but considering the very huge amount of todos in the issues, I'm not sure I'll be available to review this before some time... 

Right now, my focus is on the 25 opens issues in the [Gladys 4 RC milestone](https://github.com/GladysAssistant/Gladys/milestone/6). 

Sorry for saying that after you spent some time on implementing this 😞Don't take it personaly, it's great work you're doing 😄 

But I really want to move forward on Gladys 4 RC otherwise it'll never go out...

Pierre-Gilles:
Coming back after seeing your PR on RTSP camera! 

I'm not sure what's the next step: 

- Can we merge the RTPS camera PR without this one?

atrovato:
Yes we can merge the other, it includes both.
I just separated PRs to get a better view on changes.

But please take a look at the UI of the other PR, camera form is getting bigger...

Pierre-Gilles:
**Switching in french to be clear**

J'ai compris que l'autre PR incluais les deux, le soucis c'est que je ne suis pas sur qu'on veuille les variables secret dans Gladys maintenant...

Je pense que c'est un sujet qui demande une vrai réflexion d'UX pour chaque service, et je ne veux pas mettre en avant cette façon de faire dans Gladys avant qu'on ait eu la réflexion.

- Du côté sécurité, c'est très bien que certaines variables soit "cachée"
- Du côté expérience utilisateur, ça dégrade l'expérience. L'utilisateur aurait du mal à retrouver ses secrets dans le cas du MQTT, dans le cas de la caméra, si l'utilisateur doit vérifier que sa caméra a bien les bons identifiants, comment il fera? Quels sont les best-practices?

C'est toujours une balance entre les deux. Trop secure c'est inutilisable, pas assez secure c'est pas terrible. 

Je pense qu'il faut qu'on discute de tout ça, moi de mon côté j'aimerais me renseigner, mais je ne pense pas que ce soit la priorité là vu les autres issues en attentes.

Désolé je ne veux pas te faire perdre du temps.. :/ 

Pour le RTSP, pour que ce soit merge là, il faudrait retirer le code des variables secret... Désolé..

atrovato:
Ok, je suis du même avis que toi, il a toute une reflexion à mener à ce sujet.
Je nettoie l'autre PR ;)
Aucun soucis.

Pierre-Gilles:
Top! Merci beaucoup pour tes PRs en tout cas! 🙂

PR #757, opened 2020-04-27
zWave - Dimmer integration

sescandell:
### Pull Request check-list

- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).

### Description of change

Manage zWave Dimmer devices (devices supporting Command Class Switch MultiLevel (38)).
Introducing Command Class Object (to spread all over in the zWave service).

Update to OZW 1.6 must be merged first (#675 )


VonOx:
@link39 I think you and @sescandell are working on the same things ( zwave service ).

sescandell:
Oh, did I missed something?  A conversation somewhere else or another PR?

link39:
Hello,

I created a topic to discuss about zwave bug and improvement : 
https://community.gladysassistant.com/t/zwave-aide-pour-la-v4/5312/55

I invit you to join-us. 

In my side, I added : 
- dimmer
- multiple zwave instance
- mapping some different comclass
Before doing a PR I want to develop some other things.  
- A debug page with a button to export openzwavecache.xml file
- implement the node configuration function

sescandell:
Ok, talking there

VonOx:
>       Before doing a PR I want to develop some other things.

@link39 you can make a draft PR so others users can see what is under developpment. ;)

sescandell:
@link39 here is what I was talking about for "Command Classes objects".

For now, they are only used to manage `setValue` and `valueChanged` (and a part of `getNodes` for min/max values).
Next step would be to use them more globally in the zWave service for all things that would play with zWave values.

For features and params, it  can be based on this logic (I can provide a sample...) but this PR is not the good place for that.

Pierre-Gilles:
Hey ! What's the status with this PR ? :) 

I really want to push Z-Wave improvements in the RC, so let me know if you need help

sescandell:
> Hey ! What's the status with this PR ? :)
> 
> I really want to push Z-Wave improvements in the RC, so let me know if you need help

Just need to double check live updates.

I should be able to work on it soon. I'll let you know!

sescandell:
Hi @Pierre-Gilles 

Just pushed a last commit. Everything is working as expected.

Next to this PR, I would be able to work on managing "features" tied to devices (following far discussions with @link39 )

The @link39 PR and this one have some things in common. That one is more focus on a specific subject and is going further on how dimmer work. Don't know if @link39  has time, but I should be able to take some of its work in another PR after that one is merged and continue to improve the zWave module.

Let me know if you have any question,

Thanks,

server/services/zwave/lib/comClass/defaultCommandClass.js
```diff
@@ -0,0 +1,83 @@
+const { transformValue, normalizeValue } = require('../utils/valueBinders');
+
+const DefaultCommandClass = function DefaultCommandClass() {};
+
+DefaultCommandClass.prototype = {
```
Pierre-Gilles, line 5: Why didn't you use functions? I don't really approve the use of this design pattern here. 

Just write the function and exports them, the simpler the better!

server/services/zwave/lib/comClass/switchMultilevelCommandClass.js
```diff
@@ -0,0 +1,119 @@
+const { DefaultCommandClass } = require('./defaultCommandClass');
+const logger = require('../../../../utils/logger');
+const { COMMAND_CLASSES, INDEXES } = require('../constants');
+
+const SwitchMultilevelCommandClass = function SwitchMultilevelCommandClass() {
+  DefaultCommandClass.call(this);
```
Pierre-Gilles, line 6: Same here, I don't think we want this kind of inheritance here, it complexifies the code for not much value added. 

Can you just use functions? The simpler, the better

server/services/zwave/lib/constants.js
```diff
@@ -260,10 +261,24 @@ const CATEGORIES = [
   },
 ];
 
+const VALUE_MODES = {
+  READ_ONLY: 1,
+  WRITE_ONLY: 2,
+  READ_WRITE: 3,
+};
+
+const GENRES = {
```
Pierre-Gilles, line 270: This variable is in french, please write in english :)

server/services/zwave/lib/comClass/switchMultilevelCommandClass.js
```diff
@@ -0,0 +1,119 @@
...
+ * @description Test if provided comClass support V4.
+ *
+ * @param {Object} comClass - The comClass zWave object.
+ *
+ * @returns {boolean} True if the current node supports V4. False otherwise.
+ *
+ * @example
+ *  const isSupportingV4 = supportV4(zWaveComClassObject);
+ */
+function supportV4(comClass) {
```
Pierre-Gilles, line 25: This function name is not really explicit, which v4 is it? Gladys v4? 

From an outside point of view, it's not clear enough :)

front/src/actions/device.js
```diff
@@ -56,6 +54,9 @@ function createActions(store) {
                   [deviceFeatureIndex]: {
                     lastValue: {
                       $set: value
+                    },
+                    lastValue_changed: {
```
Pierre-Gilles, line 58: Please use camelCase :)

Pierre-Gilles (changes requested):
Hi ! 

Thanks for your PR. 👍 

Here is a few feedback on the code quality of the PR. Nothing much, just a few feedbacks so the code is easier to maintain and understand in the long run.

By the way, your code coverage is not enough for the PR to pass. You need to add more tests!

See CodeCov : https://codecov.io/gh/GladysAssistant/Gladys/pull/757

server/services/zwave/lib/comClass/defaultCommandClass.js
sescandell: Because they are not just "functions" it's a real object with real behaviors. It's not just a set of tools or utils. It represents an object with a set of functionnalities. And it's actually how Javascript is working for this. Finality here is to have the possibility to get one fully representative object per zWave CommandClass to handle specific cases on specific CommandClasses.

Look at how the factory is working, look at the SwitchMultilevelCommandClass, take a deeper look on the `getMinMax` method and keep in mind that zWave command classes have common and specific behavior. It would make more sense than a set of exported functions.

server/services/zwave/lib/comClass/switchMultilevelCommandClass.js
sescandell: See previous response, the goal is to get a real object definition here to represent CommandClasses. All CommandClasses have some generic behavior and some specific ones. POO is perfectly adapted for this.

server/services/zwave/lib/comClass/switchMultilevelCommandClass.js
sescandell: You are here in a method from the switchMultilevelCommandClass class. We are so talking about the V4 of the SwitchMultilevelCommandClass... I can rename it if you prefer...

Pierre-Gilles: Yes, it's not clear enough

server/services/zwave/lib/comClass/defaultCommandClass.js
Pierre-Gilles: I read the code, I don't see the added value in using POO here. It's a singleton class with only static functions, it could just be functions.

sescandell: They are not static methods. 

Writing this comment... I just think about one thing: I maybe mixed two things in your comment (or mixed the two comments, my bad): code splitting and POO (the other comment).

This code commented here is exactly the same thing as here: https://github.com/GladysAssistant/Gladys/blob/master/server/services/zwave/lib/index.js#L64 (and all over the project). So, if your point is to explode each function in a dedicated file to finaly put it in the prototype: this code is doing exactly the same thing... but in a single file. It's far most easier to navigate in your class within a single file (knowing which properties are available) than in one class with multiple files (for example, when I'm in the file [zwave.setValue](https://github.com/GladysAssistant/Gladys/blob/4286e003c1a46e49463149dd95dc338404f4ff15/server/services/zwave/lib/commands/zwave.setValue.js#L15) `this.zWave` doesn't mean anything. It's totally out of its context)


About the POO, the other comment, look at the usage of the getMinMax value. Without it, you would have a function growing over and over with command classes. The function would consist of a big `switch()`.

Look at the code when a "value changed" event is received. Without POO, you will also need a function based on switch and functions and so on... It's not relevant enough now because we have only the SwitchMultiLevelCommandClass implemented. With time, you will have more and more.



You are the owner of the project, you will have the final word. 
So, about splitting the code in multiple files to put all of them in the prototype in a `index.js` file... Whatever... if you want... but result is exactly the same: there is no design pattern difference here... This is far most usual Javascript than using export everywhere to associate them to the prototype (to be honest, when I started to deep dive in Gladys and see this, I've been surprised... but OK, result is exactly the same). But if you absolutely want to split, I'll do.

For inheritance, I'm convinced this is a long term advantage and avoid errors in coming CommandClasses implementations. There is no magick here, just usual Javascript.

Let me know

Pierre-Gilles: > They are not static methods.

They are not using "this". They are just using function parameters to return a result, it's just utils functions. 

>  result is exactly the same

Yes, but it'll be readable. Right now, yes it works but it's hard to understand and it'll lead to weird bugs in the future. 

Inheritance is debatable, to me it'll just make the maintenance more complex in this case.

From what I see, your function are just "utils" functions, it could be a simple file with functions and you export the functions. 

**About the singleton problem**

I don't agree, it's not the same as others design patterns used. You're exporting the object, not the class.

Screenshot 2020-09-03 at 11 51 28

In short: you're using the filesystem (require) as a way to export a singleton.

In Node.js, when you do a "require" on a file, the result is cached. So if N files are requiring your file, just one instance of your class will be created. 

This is not right, because it means the object is now a global, and tests are sharing a global variable. 

In the example you take, I'm exporting the class, not an object of the class.

Screenshot 2020-09-03 at 11 54 54


> Let me know

Just do a simple file with functions and call the functions: simpler, easier to read & maintain.

sescandell: Even if I don't share all your points, you're the boss.

As soon as I have time, I'll update the PR.

Pierre-Gilles: We can always debate about that in a community call, it'll be more easy to talk about that vocally ;) 

> As soon as I have time, I'll update the PR.

Thank you very much for your time and contribution to Gladys 🙏 Don't take personally my feedbacks, computer science is a topic full of controversy and everyone has different theory. It's great to have such discussion in PR 🙂

sescandell:
See #1221

Pierre-Gilles:
Why did you change to a new PR? All my previous comment are still valid, and are now lost!

sescandell:
> Why did you change to a new PR? All my previous comment are still valid, and are now lost!

Parce que suite à la migration des autres PR (notamment celle de la gestion du multi-instances) et le fait de ne pas vouloir d'héritage pour la gestion des CommandClass (ce qui, en ré-écrivant le code, ne me semble pas une bonne idée), le travail était plus important que de rebase / adapter celle-ci à priori.

Je peux tenter un foce-push sur cette branche pour autant... les noms de fichiers sont les mêmes, peut-être que Git va bien s'en sortir... je peux tester

Pierre-Gilles:
Ok! Tu peux rester dans une nouvelle PR si c'est plus simple pour toi :)

sescandell:
A priori, malgré un force push => impossible de ré-ouvrir ici... il se perd GIT je pense :)

PR #781, opened 2020-05-09
Philips hue model LOM002

Terdious:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [x] If your changes affects code, did your write the tests?
- [x] Are tests passing? (`npm test` on both front/server)
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [ ] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- [ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/gladys-4-docs) and the [website](https://documentation.gladysassistant.com).
- [ ] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change
Add model LOM002 Hue Smart plug
https://community.gladysassistant.com/t/philips-hue-accessoires/5339/3?u=terdious

PR #853, opened 2020-08-20
✨ add device Philips Hue LWA001

guillaumeLamanda:
### Pull Request check-list

  I just added the JSON data 
- [x] Are tests passing? (`npm test` on both front/server). 

### Description of change

This pull request adds the compatibility for the Philips Hue LWA001 #851

### DIFF
--- a/server/services/philips-hue/lib/light/light.getLights.js
+++ b/server/services/philips-hue/lib/light/light.getLights.js
@@ -44,6 +44,7 @@ async function getLights() {
           lightsToReturn.push(getPhilipsHueColorLight(philipsHueLight, serialNumber, this.serviceId));
           break;
         case 'LWO001': // Hue White Filament Bulb G93 E27
+        case 'LWA001': // Hue White Bulb E27 (Dimmable light 2700K)
         case 'LWB010': // Hue white bulb with fixed warming light (2700K)
         case 'LWB006': // Hue white lamp
         case 'LWG004': // Hue white spot
--- a/server/test/services/philips-hue/lights.json
+++ b/server/test/services/philips-hue/lights.json
@@ -1182,6 +1182,68 @@
       },
       "_id": 25,
       "id": 25
+    },
+    {
+      "state": {
+        "on": true,
+        "bri": 254,
+        "alert": "select",
+        "mode": "homeautomation",
+        "reachable": true
+      },
+      "swupdate": {
+        "state": "transferring",
+        "lastinstall": null
+      },
+      "type": "Dimmable light",
+      "name": "Hue white lamp 2",
+      "modelid": "LWA001",
+      "manufacturername": "Signify Netherlands B.V.",
+      "productname": "Hue white lamp",
+      "_rawData": {
+        "state": {
+          "on": true,
+          "bri": 254,
+          "alert": "select",
+          "mode": "homeautomation",
+          "reachable": true
+        },
+        "swupdate": {
+          "state": "transferring",
+          "lastinstall": null
+        },
+        "type": "Dimmable light",
+        "name": "Hue white lamp 2",
+        "modelid": "LWA001",
+        "manufacturername": "Signify Netherlands B.V.",
+        "productname": "Hue white lamp",
+        "capabilities": {
+          "certified": true,
+          "control": {
+            "mindimlevel": 5000,
+            "maxlumen": 800
+          },
+          "streaming": {
+            "renderer": false,
+            "proxy": false
+          }
+        },
+        "config": {
+          "archetype": "classicbulb",
+          "function": "functional",
+          "direction": "omnidirectional",
+          "startup": {
+            "mode": "safety",
+            "configured": true
+          }
+        },
+        "uniqueid": "00:17:88:01:08:9f:97:bd-0c",
+        "swversion": "1.53.3_r27175",
+        "swconfigid": "D6781B83",
+        "productid": "Philips-LWA001-1-A19DLv5"
+      },
+      "_id": 26,
+      "id": 26
     }
   ]
 }


guillaumeLamanda:
That if the lint phases was done in the CI ?  
I also think prettier should be a pre-commit hook (`husky` + `lint-staged`).

Pierre-Gilles:
Thanks for your first PR ! 🚀

**About the linting**

I recommend using an IDE (like VS Code) with the Prettier + Eslint extension installed + “Format on save”/“Format on paste” parameter activated. With that, your code will be formatted and linted each time you save a file.

Example:

Screenshot 2020-08-20 at 17 50 13

**About pre-commit hooks**

I’m not really in favor of pre-commit hooks in general (even pre-push hooks). 

With pre-commit hooks, committing is slower and do not work consistently (when prettier returns an error for example).

Git should be as fast as possible, so that developers can commit as much as they wants. We want to incentivize a rich git history and frequents commits so developers can go back to their previous code as much as possible, and avoid data loss in case of broken/stolen/lost laptop. 

It can sound paranoid for a one file PR like this one, but many big PRs on Gladys can be worked on for months, so we want to incentivize developers to commit as frequently as possible, even if prettier/eslint is not passing. 

**About linting on CI**

About the linting on CI, I’m not in favor of it either. I think linting is for the developer, it helps the developer see what’s wrong in his code visually (you should have linting displayed live in your IDE), so he can improve and understand his mistakes. 

As [this article](https://www.thoughtworks.com/insights/blog/pre-commit-don-t-git-hooked) says it nicely

“Ironically, The discipline of maintaining a stable build is gained through the experience of breaking the build.”

**Running tests**

About the tests, as you modified the backend only, I recommend that your run only the backend tests suites locally. 

Go to the `server` folder, and run `npm test`.

Here, your didn’t write the tests for your service so the coverage is going down and coverage check is failing.

You need to modify the server/test/services/philips-hue/lights.json file and add the JSON of the light you just added. 

It’ll make sure that your light is well handled now, and that it keeps working in the future. 

See this changes in the PR #781 https://github.com/GladysAssistant/Gladys/pull/781/files#diff-ba109fc7d8fc8cc0fdab8d523fa70749

guillaumeLamanda:
Ok, I understand the main point. That's ok for me, every project has his own rules, and I'm ok to use those ones.  

About running tests, I was not able to run them due to errors with the node module version. I can't find any package specifying a version for the project. Did I miss something? 

Anyway, the contribution should be good by now

Pierre-Gilles:
> About running tests, I was not able to run them due to errors with the node module version. I can't find any package specifying a version for the project. Did I miss something?

I suppose you mean “Node version”, we use Node LTS (Node 12).

Without more logs it’s hard to help. Don’t hesitate to create a specific post on our forum if you want help on this.

About your changes, you didn’t put the correct informations in the JSON, see the other rows in the file.

guillaumeLamanda:
> I suppose you mean “Node version”, we use Node LTS (Node 12).  

Yes, that's the part I missed. I was using node 14, and the tests don't run with this version.  
Just a thought: maybe specifying the node version could help future developers?  
It can be specified under the `engine` key in `package.json`, or `nvm` is a common solution to do so.  
What do you think?

> About your changes, you didn’t put the correct informations in the JSON, see the other rows in the file.

Indeed, I didn't pay enough attention to the JSON structure. Change done ✅

Pierre-Gilles (changes requested):
>Just a thought: maybe specifying the node version could help future developers?

Good idea, I added the Node version used in the documentation for now! 

> It can be specified under the engine key in package.json, or nvm is a common solution to do so.
What do you think?

Good idea for the engine key :) 

So I just reviewed your code, and it seems your PR is breaking the LOM002 integration. If you have a close look at the switch case, the LOM002 used to return an ON/OFF plug device, and now it returns a Philips Hue white light.

Can you fix this?

Screenshot 2020-08-24 at 10 42 06

It makes me think that tests are not strict enough because this should have been caught in tests. 

Before we improve philips hue tests, can you fix your code + add a comment on the ` case 'LWA001':`l line because when another developer will read this file, he won't understand what is a "LWA001" and can make the same mistake as you :) 

By the way, the `getPhilipsHueWhiteLight` line is now two times in the file and it should be there only once.

Screenshot 2020-08-24 at 10 49 19

You should put your case statement before the first getPhilipsHueWhiteLight()

guillaumeLamanda:
> Good idea, I added the Node version used in the documentation for now!
> Good idea for the engine key :)

Awesome! 

> So I just reviewed your code, and it seems your PR is breaking the LOM002 integration. If you have a close look at the switch case, the LOM002 used to return an ON/OFF plug device, and now it returns a Philips Hue white light.
> 
> Can you fix this?

Indeed. It is corrected by now. 😄 

> It makes me think that tests are not strict enough because this should have been caught in tests.

When I looked at the test case, it seems ok IMO. But maybe the controller should be handled by another file dedicated to controllers? It is a bit strange to have controllers in a file called `lights` 🤔 

Another thought: I have seen Ikea light in this folder called `philips-hue`. Did they work the same way?

Pierre-Gilles:
Thanks for fixing, it's way better now.

> When I looked at the test case, it seems ok IMO. B

What I meant, is that our tests should verify that the expected light object are the expected one, right now we don't really verify in tests, thus validating your PR despite it being broken.

 > But maybe the controller should be handled by another file dedicated to controllers? It is a bit strange to have controllers in a file called lights 🤔

The `lights.json` file is just a fixture file used for mocking in the `mocks.test.js` file. 

The tests code is in files like this one: https://github.com/GladysAssistant/Gladys/blob/master/server/test/services/philips-hue/light/light.getLights.test.js

> Another thought: I have seen Ikea light in this folder called philips-hue. Did they work the same way?

Yes, IKEA lights are philips hue compatible and handled in the Philips Hue service the same way

Pierre-Gilles:
@all-contributors please add @guillaumeLamanda for code

PR #913, opened 2020-10-25
Notify device creation / update / deletion

atrovato:
### Pull Request check-list

### Description of change

As we have set `setValue` method on device service, it propose to have `onNewDevice`, `onUpdateDevice` and `onDeleteDevice` functions on service to handle device changes.


server/lib/device/device.notify.js
```diff
@@ -0,0 +1,44 @@
...
+  } else {
+    // send global event
+    this.eventManager.emit(event, device);
+
+    // notify concerned service
+    const service = this.serviceManager.getServiceById(device.service_id);
+    if (service === null) {
+      logger.warn(`Service ${device.service_id} was not found.`);
+    } else if (typeof get(service, `device.${serviceFuncName}`) !== 'function') {
+      logger.warn(`Function device.${serviceFuncName} in service ${service.name} does not exist.`);
```
Pierre-Gilles: I think a warn is too high, people are scared in the forum when they see a warning in their logs ! 

A simple "info" might be enough IMO

server/lib/device/device.notify.js
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      service.device[serviceFuncName](device);
```
Pierre-Gilles: What happens if the function in the service throw an error ?

It shouldn't make the initial create/update/delete device fail.

Quick question: why did you choose to do a function call, and not an event here? (this is not a trap question :D just to understand your choice!)

Pierre-Gilles (changes requested):
Thanks for your PR ! 

I have a few questions on it.

server/lib/device/device.notify.js
atrovato: Ok, I will :)

server/lib/device/device.notify.js
atrovato: I'm not sure this should undo the device modification.
According to my vision, this is an non blocking action, as it a "post create" action, handled outside the core "classical" process. But this can change.

Why not using an event on device related service:
1. I think this could be an extension of the service definition for developers (as setValue or poll), so no need to explicitly subscribe to any event if this need is required on the service -> make dev easier
2. to limit event usage (and nodejs MaxListeners warn)
3. as we only want to suscbribe to bluetooth events, the event name will looks like "bluetooth_device_created", so for all services, we'll get a specific event name (eg. "philips-hue_device_created"), this could be a source of errors (typos in event name for instance)

But the global event is still an event, so thrid party services can handle it.

server/lib/device/device.notify.js
atrovato: I think we should keep 2 others in warn level, which could reveals bugs.

server/lib/device/device.notify.js
Pierre-Gilles: Ok, good point.

I'm ok with functions, but as least please add a try catch around the function call !

atrovato: Changes on 2nd commit.

atrovato:
@Pierre-Gilles now I'm not sure of the method names. We'll have `service.device.onNewDevice`, that's lot of device, no?

Pierre-Gilles:
@atrovato you're thinking of `service.device.onNew` ?

atrovato:
@Pierre-Gilles `onCreate/onUpdate/onDelete`?
`onNew` is too short.

atrovato:
Or `postCreate/postUpdate/postDelete`?

Pierre-Gilles:
> Or postCreate/postUpdate/postDelete?

I like it ! It shows the real flow

atrovato:
Rename done :)

atrovato:
More than rename, I add tests on all new lines.

atrovato:
I may change `setValue` and `poll` too.
Both methods (and new `notify`) are very very close.

Pierre-Gilles (approved):
Thanks for this great PR!

PR #918, opened 2020-11-01
Use common integration components

atrovato:
Try to use a common device component to normalize each integration.
Every time I create a new service, I copy/paste this components and i18n properties, I think we can do it better.

Components:
- Device (device card form device fields, and save / delete / edit buttons)
- DeviceFields (name, house)
- DeviceList (title, filters, device list and empty state)
- DeviceListOptions (only filters)

That's just a proposal.
What's your opinion?


Pierre-Gilles:
Yes! great idea!

Pierre-Gilles:
You say "it's just a proposal", but is this ready to be merged? If yes, I'm 100% ok. 

I didn't do a shared device components at first because we lacked experience of "what needs each service". But now that we have many services running, and we know what is needed, I'm all need to share this code between services

atrovato:
It is ready to merge, only device box and device list for now, but we can improve it with time.
As you see, you can use only some sub components (for MQTT service for instance, we need to add more data in feature component, so we only use some parts).

atrovato:
I will redo this once cypress tests are available.

PR #920, opened 2020-11-02
Fix tasmota doc link

atrovato:
Only change tasmota documentation key (sonoff to tasmota).


Pierre-Gilles:
Yes! I had that in mind! Thanks ! :D

PR #928, opened 2020-11-08
Dashboard : Device value on 2 lines

### LINKED ISSUE #925: Dashboard : Device value on 2 lines
I don't know why...
Browser : Opera GX on Mac OS
Same thing on Chrome, obviously (Opera based on Chrome)

Capture d’écran 2020-11-08 à 14 28 44

atrovato:
Fixes #925

Do not allow to display device value on multiple lines (except for "no value received").

### DIFF
--- a/front/src/components/boxs/device-in-room/device-features/SensorDeviceFeature.jsx
+++ b/front/src/components/boxs/device-in-room/device-features/SensorDeviceFeature.jsx
@@ -1,5 +1,6 @@
 import { Text } from 'preact-i18n';
 import get from 'get-value';
+import cx from 'classnames';
 import dayjs from 'dayjs';
 import relativeTime from 'dayjs/plugin/relativeTime';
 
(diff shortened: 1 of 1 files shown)


Pierre-Gilles:
So how does it looks like when the text is long ?

front/src/components/boxs/device-in-room/device-features/SensorDeviceFeature.jsx
```diff
@@ -23,7 +24,7 @@ const SensorDeviceType = ({ children, ...props }) => (
     </td>
     <td>{props.deviceFeature.name}</td>
     {SPECIAL_SENSORS.indexOf(props.deviceFeature.category) === -1 && (
-      <td class="text-right">
+      <td class={cx('text-right', { 'text-nowrap': props.deviceFeature.last_value })}>
```
Pierre-Gilles: Can you do a stricter type check on this ? If last_value = 0, (ex: temperature = 0°C) then it'll be "false" here, but we want true. 

check only if null

front/src/components/boxs/device-in-room/device-features/SensorDeviceFeature.jsx
```diff
@@ -51,7 +52,7 @@ const SensorDeviceType = ({ children, ...props }) => (
       </td>
     )}
     {props.deviceFeature.category === DEVICE_FEATURE_CATEGORIES.MOTION_SENSOR && (
-      <td class="text-right">
+      <td class={cx('text-right', { 'text-nowrap': props.deviceFeature.last_value })}>
```
Pierre-Gilles: same here

Pierre-Gilles (changes requested):
Thanks for the PR ! 

Just a little feedback on type, nothing serious but this can lead to very weird bugs

front/src/components/boxs/device-in-room/device-features/SensorDeviceFeature.jsx
atrovato: Here it should be `last_value_changed` not `last_value`.

atrovato:
Before:



After:

Pierre-Gilles:
Thanks for your changes! Good for me.

Issue #947, opened 2020-11-12
Platform i386 removed from node alpine v12 & v14


VonOx:
This https://github.com/nodejs/docker-node/issues/1344

will break i386 build on node alpine v12 & v14 ( new lts ) based on manifest

`docker manifest inspect -v library/node:12-alpine | jq '.[].Descriptor.platform'`

```json
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v6"
}
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v7"
}
{
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
{
  "architecture": "ppc64le",
  "os": "linux"
}
{
  "architecture": "s390x",
  "os": "linux"
}
```

This is not a problem for now with circlci because we do not use manifest ( of node-alpine )  for our multiarch build ( force pull arch specific )
![image](https://user-images.githubusercontent.com/1528694/98954098-b8c24f80-24fd-11eb-8bc4-9512ad9e1042.png)

PR #948, opened 2020-11-12
Github Workflows implementation

VonOx:
This PR add 2 Github workflows/actions

# docker-pr-build.yml

This workflow is triggered on no draft PR.
Here we build amd64 docker image without docker login and without pushing to DockerHub

# docker-dev-build.yml

This one is a manual action
It build and push multi arch image to gladys:dev tag based on master branch.

Currently there is a bug in node alpine ( i386 is unavailable for building ) #947 

Tested successfully in repo => https://github.com/VonOx/GLADYS_CI

----------------
@Pierre-Gilles like circle ci you need to set repo secrets:
```
DOCKERHUB_PASSWORD
DOCKERHUB_USER
```

### DIFF
--- a/.github/workflows/docker-dev-build.yml
+++ b/.github/workflows/docker-dev-build.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup nodejs
         uses: actions/setup-node@v1
         with:
-            node-version: 14.x
+            node-version: 12.x
       - run: cd front && npm install && npm run build
       - name: Upload build artifact
         uses: actions/upload-artifact@v1
--- a/.github/workflows/docker-dev-build.yml
+++ b/.github/workflows/docker-dev-build.yml
@@ -1,0 +1,87 @@
+name: Build Gladys dev images
+
+on:
+    workflow_dispatch:
+
+jobs:
+  build-front:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup nodejs
+        uses: actions/setup-node@v1
+        with:
+            node-version: 14.x
+      - run: cd front && npm install && npm run build
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v1
+        with:
+          name: static
+          path: front/build
+  docker:
+    needs: build-front
+    name: Build and Push Gladys dev images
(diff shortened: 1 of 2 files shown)


Pierre-Gilles:
Nice! 

And how does it work for the dev build? With a manual trigger?

VonOx:
> 
> 
> Nice!
> 
> And how does it work for the dev build? With a manual trigger?

Yep , Actions Tab and run job

Pierre-Gilles:
Amazing! Let me know when this is ready for review :)

VonOx:
It's ready but if you merge we cant build i386 platform

Pierre-Gilles:
Ah yes alright. What's the next step for that then? is i386 used right now ? (I don't remember for which platform we build it?)

VonOx:
Yes we build ( with circleci ) x86 images , I have no idea how many pulls but maybe for dev images it is acceptable to not have i386 images

Pierre-Gilles:
> Yes we build ( with circleci ) x86 images 

Which machines exactly are using those images ?

It's maybe ok to drop support for this honestly, even more for dev images :)

VonOx:
Some NAS I suppose and old 32bits processors

Majority of running instance use arm I think ( we build arm v6/v7 and arm64 ). We maybe cover about 95% of the scope

I removed the draft flag, you can see workflow running
https://github.com/GladysAssistant/Gladys/actions/runs/366056548

.github/workflows/docker-dev-build.yml
```diff
@@ -0,0 +1,87 @@
...
+      uses: actions/checkout@v2
+    - name: Download build artifact
+      uses: actions/download-artifact@v1
+      with:
+        name: static
+    - name: Set up image tag
+      run: |
+          set -vx
+          echo "TAG=dev" >> $GITHUB_ENV
+          echo "DOCKER_BASE=${DOCKERHUB_USER}/${DOCKERHUB_REPO}" >> $GITHUB_ENV
```
Pierre-Gilles: I just realized this won't work ! 

My dockerhub_user is my personal account but we push the image to the gladysassistant organization, so it can't be the same 

Maybe something like in CircleCI like REGISTRY/IMAGE ?

PR #977, opened 2020-11-24
Dockerfile - Add bluez sys deps ( missing )

VonOx:
bluez is missing in dockerfile used by Github Actions


atrovato (approved):
Oops, I only add this to Dockfile, missing this one.

Pierre-Gilles (approved):
Nice catch!

PR #978, opened 2020-11-25
chore: Add node/npm engines in package.json

cicoub13:
https://docs.npmjs.com/cli/v6/configuring-npm/package-json

After discussion https://community.gladysassistant.com/t/fresh-development-installation/5785

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -1,5 +1,9 @@
 {
   "name": "gladys-front",
+  "engines": {
+    "node": "12.x",
+    "npm": "6.x"
+  },
   "scripts": {
     "start": "per-env",
     "start:production": "npm run -s serve",
--- a/package.json
+++ b/package.json
@@ -3,6 +3,10 @@
   "version": "4.0.1",
   "description": "A privacy-first, open-source home assistant",
   "main": "index.js",
+  "engines": {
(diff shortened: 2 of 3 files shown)


VonOx:
Not sure if this help because no warning is produce and user can run anything ( Maybe I'm wrong ? )

cicoub13:
> Not sure if this help because no warning is produce and user can run anything ( Maybe I'm wrong ? )

You're right, no warning but developers can now easily which version of node/npm to use. If you think documentation (https://gladysassistant.com/fr/docs/api/setup-development-environnement) is enough, I'll close this PR.

There is one way to force node version, it's by adding a `preinstall` script in `package.json` that checks version. But I find it too restrictive.

VonOx:
> You're right, no warning but developers can now easily wich version of node/npm to use. If you think documentation (https://gladysassistant.com/fr/docs/api/setup-development-environnement) is enough, I'll close this PR.

No , I'm just asking ;)

> I find it too restrictive.

I agree

Pierre-Gilles (approved):
Good idea!

Pierre-Gilles:
@all-contributors please add @cicoub13 for code

PR #1003, opened 2020-12-05
MQTT: UI bug when connecting external broker

### LINKED ISSUE #943: MQTT: UI bug when connecting external broker
If you go to the MQTT Integration, then configure an external broker, then save, then leave and come back, you see this as it's was not configured : 

5SZkOKSU

Source : 

Screenshot 2020-11-11 at 10 26 48

atrovato:
Fixes #943


Pierre-Gilles (approved):
Thanks for your fix!

PR #1004, opened 2020-12-05
Disk space rounded

### LINKED ISSUE #989: On "settings" -> "System" page, the disk occupied space in percentage is too precise (15 digits), should be rounded
L'affichage du % de la capacité de la carte mémoire est trop long, 
il y a 15 zéro après la virgule.

atrovato:
Fixes #989
Using Math.ceil instead of rounded, to me I prefer to 26% when 25.3% is real.


Pierre-Gilles (approved):
Thanks for your PR!

PR #1022, opened 2020-12-20
Fix #1021 service status

### LINKED ISSUE #1021: Service status detection
Many services are disabled on startup.
And UI is not explicit enough.

atrovato:
Fixes #1021 

### Pull Request check-list

### Description of change

Change service status management.
No NOT_CONFIGURED status.
Service STOPPED status only on user request.


server/migrations/20201220100508-service-remove-not-configured-status.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  up: async (queryInterface) => {
+    await queryInterface.sequelize.query(
+      "UPDATE t_service SET status='RUNNING' WHERE status='NOT_CONFIGURED' OR status='STOPPED'",
```
Pierre-Gilles: Could you write this in Sequelize instead of SQL?

For now we didn't use raw SQL queries because maybe in the future we may allow other types of DB (MySQL? PostgreSQL?) and I want to avoid having raw SQL (if possible) that could work in one database system and not another. 

As it's just a simple query, I think here it's possible to do it in Sequelize: https://sequelize.org/v4/manual/tutorial/instances.html#working-in-bulk-creating-updating-and-destroying-multiple-rows-at-once-

What do you think?

Pierre-Gilles (approved):
Good for me!

PR #1025, opened 2020-12-20
[ZWAVE] Update openzwave 1.6 to Rev 133

VonOx:
Updating openzwave in docker to REV 133 and a little improvment in DockerFIle


Pierre-Gilles:
Nice! Tell me when it's stable, would love to merge this :)

VonOx:
@Pierre-Gilles I've discovered that openzwave have precompiled base image. ( https://github.com/OpenZWave/openzwave-docker )

This can improve build time ( I've tested on my own machine with arm/v6 , 30 minutes to build ) and we can get a specific release of openzwave

`FROM chrisns/openzwave:alpine-1.6.1570 as openzwave`

VonOx:
I'm also considering layer caching of node_modules for npm ci steps that require a lot of time ( sqlite / openzwave shared).
It's difficult because each services use their own package.json. ( Need POC and another PR)

Pierre-Gilles:
@VonOx I'm not sure we want to start from OpenZwave as base images... I don't really trust their dev process/what they put inside the image. Honestly, I don't mind the actual build time to have secure and stable images :) 

And currently, I think we do have Docker cache for building image, no?

VonOx:
> I'm not sure we want to start from OpenZwave as base images... I don't really trust their dev process/what they put inside the image. Honestly, I don't mind the actual build time to have secure and stable images :)

Base image stay at node12:alpine , this is just multistage build , we take builded openzwave lib from their alpine image

> And currently, I think we do have Docker cache for building image, no?

Yeah, but because we do npm install on the same layer with apk install and previous step that add codebase we have to rebuild and reinstall all stuff

```
RUN apk add --no-cache --virtual .build-deps make gcc g++ python git libffi-dev linux-headers \
    && npm ci --unsafe-perm --production \
    && npm cache clean --force \
    && apk del .build-deps
```
What I propose is to COPY package.json and npm ci before copy all codebase. This way if we don't modify npm deps, cached layers are used. I think we can build all images in few minutes with cached node_modules.

There is no impact on Gladys, just the way we build.

-----------------------------------------------------------------------
After some test, just with cached /server/node_modules and after fake modification in codebase :
* amd64 => 57 seconds
* arm/v6 => 8 minutes
Huge improvment !

Pierre-Gilles:
Oh ok, in that case why not! It's definitely something that could help us release / and test release faster!

Pierre-Gilles:
About OZW Rev 133, did you test it or should it be tested before merge?

VonOx:
Tested with my only one device ( Qubino Shutter ) and Aeotec dongle, but it's not correctly handled so i can't confirm there is no bug introduced.

This update is low risk for me but nothing is sure with zwave :D

VonOx:
I will make another PR for docker improvment

VonOx:
Useless #1039

PR #1031, opened 2020-12-23
Fix dashboard device boxes

atrovato:
Showcase on PR #1029 allows to find 2 bugs on dashboard device boxes:
1. dimmer doesn't update the right device index
2. color 0 (black) introduce a color picker bug


Pierre-Gilles (approved):
Nice catch!

front/src/components/boxs/device-in-room/device-features/ColorDeviceFeature.jsx
```diff
@@ -64,7 +64,7 @@ class ColorDeviceType extends Component {
 
   componentDidMount() {
     const deviceLastValue = this.props.deviceFeature.last_value;
-    const color = deviceLastValue === null ? undefined : `#${intToHex(deviceLastValue)}`;
+    const color = !deviceLastValue ? undefined : `#${intToHex(deviceLastValue)}`;
```
Pierre-Gilles, line 67: Mmm wait, does it means that when the color is 0, it shows an `undefined`color ? Shouldn't it be the black hexadecimal color?

atrovato: The color pocker doesn't have black color, it causes trouble, and I don't think your light has any black color too.
Black means switched off, it is not really a color.

Pierre-Gilles: True!

PR #1036, opened 2021-01-10
Fix CI login and improve

VonOx:
Login from TTY is not possible anymore this PR use official docker actions
I've improve readability by using official docker action for buildx


VonOx:
Implemented in #1039

PR #1047, opened 2021-01-18
Inject scene variables in other boxes

Pierre-Gilles:
### Pull Request check-list

### Description of change


Pierre-Gilles:
Hi @cicoub13 @VonOx @atrovato, I would love a review from you on this PR! 

This is how it looks like: 

https://user-images.githubusercontent.com/7365207/104939530-d229c100-59b0-11eb-93e1-87bf4c4bcfe1.mp4

Is the shortcut straightforward? (having to type " {{ " to open the variable list ?) Should we use another trigger ? Like "#" ? 

This feature will be extended to basically all input field in scenes, for exemple in the HTTP request it could be super useful to automatically inject a variable into a POST body :)

front/src/routes/scene/edit-scene/actions/SendMessageParams.jsx
```diff
@@ -43,6 +46,80 @@ class SendMessageParams extends Component {
...
+              variableReady = true;
+            }
+            if (!option.ready) {
+              variableReady = false;
+            }
+            variableWhileList.push({
+              id: `${groupIndex}.${index}.${option.name}`,
+              text: `${groupIndex + 1}. ${index + 1}. ${option.label}`,
+              title: `${groupIndex + 1}. ${index + 1}. ${option.label}`,
+              value: `${groupIndex}.${index}.${option.name}`
```
atrovato, line 78: All these values can be computed once and stored in a const instead of cumputing them n times in the loop.

front/src/routes/scene/edit-scene/actions/SendMessageParams.jsx
```diff
@@ -43,6 +46,80 @@ class SendMessageParams extends Component {
...
+    });
+    await this.setState({ variableWhileList, variableReady });
+    if (variableReady) {
+      this.initTagify();
+    }
+  };
+  setRef = dom => (this.tagifyInputRef = dom);
+  initTagify = () => {
+    if (this.tagify) {
+      this.tagify.settings.whitelist.length = 0;
```
atrovato: Is this empty an array?

front/src/routes/scene/edit-scene/actions/SendMessageParams.jsx
```diff
@@ -43,6 +46,80 @@ class SendMessageParams extends Component {
...
+      dropdown: {
+        enabled: 1,
+        position: 'text',
+        mapValueTo: 'title'
+      },
+      whitelist: this.state.variableWhileList,
+      mixTagsInterpolator: ['{{', '}}']
+    });
+    if (this.props.action.text.search('{{') !== -1 && this.props.action.text.search('}}') !== -1) {
+      const textFormatted = this.props.action.text.replaceAll('{{', '[[').replaceAll('}}', ']]');
```
atrovato: Can we store `{{` and others in file constants? This will make it easier to change if needed.

cicoub13 (commented):
It's ok for me. Tested in multiple conditions 👍 
On little glitch when I decided to change again the user

inject_user_glitch

{{ is ok to inject variable.

atrovato (changes requested):

front/src/routes/scene/edit-scene/actions/SendMessageParams.jsx
Pierre-Gilles: Variables are changing all the time, so it's not as easy :D  

I tried to cache it the best I could, if you have another idea how we could cache it better..

Pierre-Gilles: mm I don't follow you

We have: 

- M rows (before this row)
    - Each rows has N boxes 
       - Each box has L variables

We loop on each rows before this one, then on each box in each row, then on each variable in each box.

How could we do better?

front/src/routes/scene/edit-scene/actions/SendMessageParams.jsx
Pierre-Gilles: Good idea ! Done.

Pierre-Gilles:
@cicoub13 zIndex problem fixed! :) Thanks!

PR #1051, opened 2021-01-25
fix(conversation): Spread chat messages to same user open tabs

### LINKED ISSUE #553: Spread own chat messages to other opened window of the same user (websocket)

cicoub13:
Fix #553

### Pull Request check-list

### Synchronize discussion messages between tabs

To test, open two discussion tabs in your browser (or on two different machines) and send a message. Full conversation must be synchronized between all tabs.

A question about where I put the websocket send in the server (in the controller). Is it the good place ?
https://github.com/GladysAssistant/Gladys/pull/1051/files#diff-53be8e5edecb97124c3eb6ea3c770a16db9acaf8563fa01c24b09cfc4eefae85R32


cicoub13:
https://user-images.githubusercontent.com/1773153/105813586-5b905300-5fb0-11eb-98df-8d65d1f46b93.mov

front/src/actions/message.js
```diff
@@ -74,11 +78,12 @@ function createActions(store) {
       });
       const messageText = state.currentMessageTextInput;
       try {
+        const tempId = uuid.v4();
         const newMessage = {
           text: messageText,
-          created_at: new Date()
+          created_at: new Date(),
+          uuid: tempId
```
Pierre-Gilles: Why did you use a "uuid" attribute and not "id" like at most places?

server/api/controllers/message.controller.js
```diff
@@ -25,8 +25,15 @@ module.exports = function MessageController(gladys) {
       source_user_id: req.user.id,
       user: req.user,
       created_at: req.body.created_at || new Date(),
+      uuid: req.body.uuid,
```
Pierre-Gilles: Be careful, this will be sent to the `message.create` function, that isn't supposed to accept a `uuid` attribute.

If you really need this id, you could generate the id here and use it when inserting the message in DB then, so ID are consistent.

front/src/actions/message.js
```diff
@@ -43,13 +43,17 @@ function createActions(store) {
       });
     },
     pushMessage(state, message) {
+      let newMessages = store.getState().messages;
+      if (message.uuid && message.uuid === newMessages[newMessages.length - 1].uuid) {
```
Pierre-Gilles: This won't necessarily work, because if you sent 2 messages in a row, and the websocket event comes back after the 2nd message is received, then one message will be displayed twice.

I think we have no choice to deduplicate on the whole list of messages, or maybe we could keep a store of messages already displayed (like a javascript Map() )  ?

Pierre-Gilles (changes requested):
Nice PR, thanks for this ! 

I spotted a few issues with your implementation, let me know what you think about that :)

front/src/actions/message.js
cicoub13: Like it was really an UUID (v4), I found more meanfull `uuid` than `id`. But it's fixed

server/api/controllers/message.controller.js
cicoub13: It's just used to communicate via websocket. No need to persist it. Fixed

front/src/actions/message.js
cicoub13: Fixed by using filter on all messages.

atrovato (changes requested):
Easy, nice, but do we want to add code in controllers?
Maybe we need to move the WebSocket event emit somewhere else, in the MessageHandler?
And add test on event content?

server/api/controllers/message.controller.js
```diff
@@ -27,6 +27,13 @@ module.exports = function MessageController(gladys) {
       created_at: req.body.created_at || new Date(),
     };
     gladys.event.emit(EVENTS.MESSAGE.NEW, messageToSend);
+    // We emit message.sent to synchronize all discussions tabs in Front
+    messageToSend.id = req.body.id;
```
Pierre-Gilles: Why do you edit the objet here? It'll affect the object sent previously as well (in JS, parameters are passed by reference when it's an object not by copy)

Pierre-Gilles (changes requested):

cicoub13:
> Easy, nice, but do we want to add code in controllers?
> Maybe we need to move the WebSocket event emit somewhere else, in the MessageHandler?

If we do this, the websocket event will be emitted with every process that send messages (including Telegram). Is it what we want ?

> And add test on event content?

Yes, I will add it 😉

server/api/controllers/message.controller.js
cicoub13: I need to create a copy object to add the id OR add the id only in the event ?

```
    gladys.event.emit(EVENTS.WEBSOCKET.SEND, {
      type: WEBSOCKET_MESSAGE_TYPES.MESSAGE.SENT,
      payload: messageToSend,
      userId: req.user.id,
      messageId: req.body.id, 
    });
```

Pierre-Gilles:
> If we do this, the websocket event will be emitted with every process that send messages (including Telegram). Is it what we want ?

It is what we want indeed! 

When you are on Facebook web, if you send a message to a friend from your mobile, it appears on the web too :)

server/api/controllers/message.controller.js
Pierre-Gilles: I don't know, you added the id :p

cicoub13:
I moved the code to the `handleMessage`. Like this, it works with Telegram.
And I added test 😄 

> When you are on Facebook web, if you send a message to a friend from your mobile, it appears on the web too :)

I agree but in our case, we will never be able to do :
- you write something in Chat in Gladys interface
- conversation is synchronized in live in Telegram

server/lib/message/message.handleEvent.js
```diff
@@ -8,6 +9,13 @@ const logger = require('../../utils/logger');
  */
 async function handleEvent(message) {
   try {
+    console.log(message);
+    this.event.emit(EVENTS.WEBSOCKET.SEND, {
+      type: WEBSOCKET_MESSAGE_TYPES.MESSAGE.SENT,
+      payload: message,
+      userId: message.user.id,
+    });
+    delete message.id;
```
Pierre-Gilles: again, why do you do this? you're mutating an object in live that was already sent previously. 

This will lead to weird behavior, because object are sent as references in JS.

server/lib/message/message.handleEvent.js
```diff
@@ -8,6 +9,13 @@ const logger = require('../../utils/logger');
  */
 async function handleEvent(message) {
   try {
+    console.log(message);
```
Pierre-Gilles: console.log should be removed

Pierre-Gilles (changes requested):
I still don't understand the flow with the id of the message. You create an id, then delete it, then another id is created in DB. All that on the same JS object. Why?

This implementation will lead to weird bugs!

server/lib/message/message.handleEvent.js
cicoub13: After forum discussion, behavior has changed. This discussion is outdated.

server/lib/message/message.handleEvent.js
cicoub13: Fixed

cicoub13:
After forum discussion, behavior has changed. ID is generated in frontend and share during all the conversation.
I've made a small fix on Telegram to have the same `created_at` field in the front to be able to sort messages.

front/src/actions/message.js
```diff
@@ -43,13 +43,18 @@ function createActions(store) {
       });
     },
     pushMessage(state, message) {
+      let newMessages = store.getState().messages;
+      // Check if message is already in the list
+      if (message.id && newMessages.filter(m => m.id === message.id).length) {
```
Pierre-Gilles: .filter will scan the whole array of message, it could be long just to verify that the message is absent! 

I recommend using .find() instead (https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/find). This function is optimized and will stop when the result was found

front/src/actions/message.js
```diff
@@ -74,15 +79,16 @@ function createActions(store) {
...
-          created_at: new Date()
+          created_at: new Date(),
+          id
         };
-        const tempId = uuid.v4();
         // we first push the message
         const newState = update(state, {
           messages: {
-            $push: [Object.assign({}, newMessage, { tempId })]
+            $push: [Object.assign({}, newMessage, { id })]
```
Pierre-Gilles: why do you do that? the id is already in newMessage

front/src/actions/message.js
```diff
@@ -96,7 +102,7 @@ function createActions(store) {
         actions.scrollToBottom();
         // then we send the message
         const createdMessage = await state.httpClient.post('/api/v1/message', newMessage);
-        const messagesWithoutTempMessage = store.getState().messages.filter(message => message.tempId !== tempId);
+        const messagesWithoutTempMessage = store.getState().messages.filter(message => message.id !== id);
```
Pierre-Gilles: I'm not sure this is useful anymore, now that the ID is generated client-side? what's the purpose here?

server/services/telegram/lib/message.new.js
```diff
@@ -30,7 +30,7 @@ async function newMessage(msg) {
     user_id: user.id,
     user,
     language: user.language,
-    date: msg.date,
+    created_at: new Date(msg.date * 1000).toISOString(),
     text: msg.text,
```
Pierre-Gilles, line 35: Now you require the `message.create`  function to have an `id`, but here there is no id passed. 

It can be ok, but it should be clear. Where is the id generated now? does it work? (I guess yes, and the id is generated by sequelize, but the code is not that clear on that point)

Pierre-Gilles (changes requested):
Hi! I have a few feedbacks :)

server/services/telegram/lib/message.new.js
cicoub13: I generated one UUIDv4, more clear in code.

front/src/actions/message.js
cicoub13: Deleted

front/src/actions/message.js
cicoub13: Deleted

front/src/actions/message.js
cicoub13: Modified to use find()

server/services/telegram/lib/message.new.js
```diff
@@ -30,8 +31,10 @@ async function newMessage(msg) {
...
+    // We generate ID as uuidv4 as Telegram doesn't give this type of ID
+    id: uuid.v4(),
```
atrovato, line 37: Do we really an UUID v4 ID, telegram provides a message_id, which is an integer: https://core.telegram.org/bots/api#message

cicoub13: We cannot mix UUIDv4 and int in Database
https://github.com/GladysAssistant/Gladys/blob/master/server/models/message.js#L6

atrovato: So we store telegram messages in database? We don't delegate it to telegram?

cicoub13: Exactly. We saved a copy (of user and gladys messages).

Pierre-Gilles: yes we save them in the database

front/src/actions/message.js
```diff
@@ -74,15 +90,16 @@ function createActions(store) {
...
-          created_at: new Date()
+          created_at: new Date(),
+          id
         };
-        const tempId = uuid.v4();
         // we first push the message
         const newState = update(state, {
           messages: {
-            $push: [Object.assign({}, newMessage, { tempId })]
+            $push: [Object.assign({}, newMessage)]
```
Pierre-Gilles: Still, why do you do `Object.assign({}, newMessage)` ?

No need to create a new object here, as there is no longer a new attribute to add

front/src/actions/message.js
```diff
@@ -93,15 +110,12 @@ function createActions(store) {
...
-        actions.scrollToBottom();
         // then we send the message
-        const createdMessage = await state.httpClient.post('/api/v1/message', newMessage);
-        const messagesWithoutTempMessage = store.getState().messages.filter(message => message.tempId !== tempId);
-        messagesWithoutTempMessage.push(createdMessage);
+        await state.httpClient.post('/api/v1/message', newMessage);
         // then we remove the message loading
         const finalState = update(state, {
           messages: {
-            $set: sortMessages(messagesWithoutTempMessage)
```
Pierre-Gilles, line 104: why did you remove the sort? 

The goal of the sort is to avoid the case when you ask something to Gladys, and the replies appears before the question.

Pierre-Gilles (changes requested):

front/src/actions/message.js
cicoub13: Fixed

front/src/actions/message.js
cicoub13: Fixed

atrovato (approved):
Looks fine.

Pierre-Gilles (approved):

Issue #1055, opened 2021-01-30
Bog on create House in FR


Terdious:
First, are you sure that you found a Gladys bug?

If you are not sure, you can come discuss with us on [Gladys Community](http://community.gladysassistant.com/) 🙂

**Describe the bug**
When creating a house with a name that is too long, the default returned by the UI is not the correct one :
![image](https://user-images.githubusercontent.com/35010958/106350937-e89d1a00-62d8-11eb-8757-1f52a294bcb9.png)

This should read: "Erreur de validation : Le nom de votre maison est trop long"
The EN version is good: "House name should be between 1 and 40 characters" :
![image](https://user-images.githubusercontent.com/35010958/106350995-652ff880-62d9-11eb-8ae9-cb7770bc5e36.png)

VonOx:
Good catch @Terdious 😉

Terdious:
Ok, I started to watch. There are actually 2 errors. In the file /front/src/actions/house.js - Line 230 we have:
```
} else if (status === 422 && url.includes ('/ room /')) {
  store.setState ({
  houseUpdateStatus: {
      [house.id]: RequestStatus.RoomValidationError
  }
});
```
while this error normally deals with the house error. So we should have:
`} else if (status === 422 && url.includes ('/ house')) {`
and line 233:
`[house.id]: RequestStatus.HouseValidationError`
So this error is not handled in the front / src / components / house / EditHouse.jsx file and does not exist in fr.json and en.json.

PR #1056, opened 2021-01-30
Modification of the error handling for the name of the house on the front side

Terdious:
### Pull Request check-list

### Description of change

To follow up on issue #1055

### DIFF
--- a/front/src/actions/house.js
+++ b/front/src/actions/house.js
@@ -233,6 +233,12 @@ function createActions(store) {
               [house.id]: RequestStatus.RoomValidationError
             }
           });
+        } else if (status === 422 && url.includes('/house')) {
+          store.setState({
+            houseUpdateStatus: {
+              [house.id]: RequestStatus.HouseValidationError
+            }
+          });
         } else if (status === 422) {
           store.setState({
             houseUpdateStatus: {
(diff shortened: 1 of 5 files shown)


front/src/actions/house.js
```diff
@@ -233,6 +233,12 @@ function createActions(store) {
               [house.id]: RequestStatus.RoomValidationError
             }
           });
+        } else if (status === 422 && url.includes('/house')) {
+          store.setState({
+            houseUpdateStatus: {
+              [house.id]: RequestStatus.HouseValidationError
```
atrovato, line 239: Is there really a difference with RequestStatus.ValidationError?

Terdious: Yes, it is different @atrovato. Look at the issue I specified in the PR description (#1055). When you fill in the field with a name that is too long, the RequestStatus.ValidationError error message returns that the name of the house cannot be empty.
image

atrovato: I think you only have to change to current message to inform both information: name is mandatory and max size is N chars.
And maybe limit the name input field to N chars.

On this page, if you empty the name, you will get a 422 too.

Pierre-Gilles:
Fixed in https://github.com/GladysAssistant/Gladys/commit/183742e610a37b0fc33065b43f0d5c8b0e82ff0d

PR #1063, opened 2021-02-01
Add docker legacy tags step to Github Workflow

VonOx:
Add 3 legacy tags on docker images on release workflow

v4-amd64
v4-arm
v4-arm64


Pierre-Gilles (approved):
Many thanks for this @VonOx 🙏

PR #1064, opened 2021-02-03
Log level info of Start/Stop function for all services

VonOx:
Just to get consistent I've edited all log of Start Stop function for all services

Related to https://community.gladysassistant.com/t/probleme-dintegration-philips-hue/6075/24?u=vonox


VonOx:
I definitely don't understand how codecov works

atrovato (dismissed):
Easy !

Pierre-Gilles:
@VonOx Me neither.... why is the coverage going down to 83% here 🤦‍♂️

Pierre-Gilles:
@VonOx Ok I think I may have understood. The lines you changed were not covered, and are still not covered, so that's why codecov is not happy.

What do you think of that?

VonOx:
Mmmh how is that possible ? We are using codecov from the beginning right ?

What codecov say is "Some of the services you have edited is not completely tested" ? If true, its weird .

Pierre-Gilles:
> We are using codecov from the beginning right ?

Yes and no, they were some really early service development that were not covered at 100% because codecov was a pain in the a** at this moment 😄

VonOx:
Ok time to take hands dirty ( Trying to increase coverage ).

Not so easy @atrovato 😅

Pierre-Gilles:
It seems it's just the "rtsp-camera" integration, there is no test on the stop function! 

Screenshot 2021-02-05 at 17 17 54

atrovato:
No asserts in tests... but... OK

atrovato (approved):

Pierre-Gilles (approved):

PR #1075, opened 2021-02-13
Disable scene

callemand:
### Pull Request check-list

- [X] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

### Description of change
Adding disable scene  and change start button by test

Screenshot 2021-02-13 at 21 11 58
Screenshot 2021-02-13 at 21 11 52

### DIFF
--- a/server/lib/scene/scene.addScene.js
+++ b/server/lib/scene/scene.addScene.js
@@ -32,67 +32,65 @@ function addScene(sceneRaw) {
   // first, if the scene actually exist, we cancel all triggers
   this.cancelTriggers(scene.selector);
 
-  if (scene.active) {
+  if (scene.active && scene.triggers) {
     // Foreach triggger, we schedule jobs for triggers that need to be scheduled
-    if (scene.triggers) {
-      scene.triggers.forEach((trigger) => {
-        // First, we had a trigger key, import to uniquely identify this trigger
-        trigger.key = uuid.v4();
-        if (trigger.type === EVENTS.TIME.CHANGED && trigger.scheduler_type !== 'interval') {
-          const rule = new schedule.RecurrenceRule();
-          rule.tz = this.timezone;
-          switch (trigger.scheduler_type) {
-            case 'every-month':
-              rule.date = trigger.day_of_the_month;
-              rule.hour = parseInt(trigger.time.substr(0, 2), 10);
-              rule.minute = parseInt(trigger.time.substr(3, 2), 10);
-              rule.second = 0;
-              break;
-            case 'every-week':
-              rule.dayOfWeek = trigger.days_of_the_week.map((day) => nodeScheduleDaysOfWeek[day]);
-              rule.hour = parseInt(trigger.time.substr(0, 2), 10);
-              rule.minute = parseInt(trigger.time.substr(3, 2), 10);
-              rule.second = 0;
-              break;
-            case 'every-day':
-              rule.hour = parseInt(trigger.time.substr(0, 2), 10);
-              rule.minute = parseInt(trigger.time.substr(3, 2), 10);
-              rule.second = 0;
-              break;
-            case 'custom-time':
-              rule.year = parseInt(trigger.date.substr(0, 4), 10);
-              rule.month = parseInt(trigger.date.substr(5, 2), 10) - 1;
-              rule.date = parseInt(trigger.date.substr(8, 4), 10);
-              rule.hour = parseInt(trigger.time.substr(0, 2), 10);
-              rule.minute = parseInt(trigger.time.substr(3, 2), 10);
-              rule.second = 0;
-              break;
-            default:
-              throw new BadParameters(`${trigger.scheduler_type} not supported`);
-          }
-          trigger.nodeScheduleJob = schedule.scheduleJob(rule, () => this.event.emit(EVENTS.TRIGGERS.CHECK, trigger));
-        } else if (trigger.type === EVENTS.TIME.CHANGED && trigger.scheduler_type === 'interval') {
-          let intervalMilliseconds;
-          switch (trigger.unit) {
-            case 'second':
-              intervalMilliseconds = trigger.interval * 1000;
-              break;
-            case 'minute':
-              intervalMilliseconds = trigger.interval * 60 * 1000;
-              break;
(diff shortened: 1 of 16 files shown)


atrovato (commented):
Code is fine to me,
but I tried on my local env, and when I disable a scene, it is always enabled.
For more, why don't we show the "test" button when scene is disabled ? We may want to test is even if it is not enabled, no?

callemand:
> Code is fine to me,

Thank you

> but I tried on my local env, and when I disable a scene, it is always enabled.

It's weird. is it on debug mode ?

> For more, why don't we show the "test" button when scene is disabled ? We may want to test is even if it is not enabled, no?

Ok, it's a mistake on my side. When I read the code, I had the impression that the scene had to be configured in node-schedule to be able to be forced but by rereading the code, I realized that it was not. I will make the changes accordingly

server/lib/scene/scene.addScene.js
```diff
@@ -31,65 +31,68 @@ function addScene(sceneRaw) {
...
-            break;
-          default:
-            throw new BadParameters(`${trigger.unit} not supported`);
-        }
-        if (intervalMilliseconds > MAX_VALUE_SET_INTERVAL) {
-          throw new BadParameters(`${trigger.interval} ${trigger.unit} is too big for an interval`);
+
+  if (scene.active) {
+    // Foreach triggger, we schedule jobs for triggers that need to be scheduled
+    if (scene.triggers) {
```
atrovato: Why not both if in single one?

If scene is active and have trigger, no.

atrovato (changes requested):
A simple code comment, and build now fails.
Once both fixed, I'll try it again.

server/migrations/202102212184604-add-active-scene.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return Promise.all([
+      queryInterface.addColumn('t_scene', 'active', {
+        type: Sequelize.BOOLEAN,
+        defaultValue: true,
```
Pierre-Gilles, line 6: we might want a `allowNull: false,` here

server/models/scene.js
```diff
@@ -115,6 +115,10 @@ module.exports = (sequelize, DataTypes) => {
       last_executed: {
         type: DataTypes.DATE,
       },
+      active: {
+        type: DataTypes.BOOLEAN,
+        defaultValue: true,
```
Pierre-Gilles, line 121: same, `allowNull: false,`

server/lib/scene/scene.addScene.js
callemand: You are right 👍

atrovato:
@callemand 
Same issue, scenes is never disabled.

> > but I tried on my local env, and when I disable a scene, it is always enabled.

> It's weird. is it on debug mode ?

Yes it is on debug mode.
No log displayed when I switch scene on/off.
I think, you miss a "active" condition on scene.executeActions.js (at execution time).

callemand:
@atrovato 
Thank for your return

> Same issue, scenes is never disabled.
> 
> > > but I tried on my local env, and when I disable a scene, it is always enabled.
> 
> > It's weird. is it on debug mode ?
> 
> Yes it is on debug mode.
> No log displayed when I switch scene on/off.

Normally, the scene is disable now :). I had only tested on the periodic trigger because i don't have any device connected on my dev environment.

> I think, you miss a "active" condition on scene.executeActions.js (at execution time).

Yes, I have miss an "active" condition. I have added a condition into scene.execute.js not into scene.executeAction.

cicoub13:
Will fix #980

atrovato:
Now you miss tests :/

callemand:
> Now you miss tests :/

@atrovato
Sorry, but I don't understand, all unit test works.

But the pipeline failed with this error: '85.71% of diff hit (target 95.55%)'
But on the top of the PR the codecov return: 
```
@@           Coverage Diff           @@
##           master    #1075   +/-   ##
=======================================
  Coverage   95.55%   95.55%           
=======================================
  Files         531      531           
  Lines        7112     7114    +2     
=======================================
+ Hits         6796     6798    +2     
  Misses        316      316      
```     

What i missed ?

atrovato:
What I understand from codecov comment, 

`server/lib/scene/scene.execute.js | 92.85% <83.33%> (+1.19%)` is not tested enough.

server/test/lib/scene/scene.execute.test.js
```diff
@@ -71,4 +71,27 @@ describe('SceneManager', () => {
...
+          {
+            type: 'BAD-TYPE',
+            devices: ['light-1'],
+          },
+        ],
+      ],
+      active: true,
+    };
+    sceneManager.addScene(scene);
+    await sceneManager.execute('my-bad-scene');
```
atrovato, line 97: But no assertion here!

atrovato (changes requested):
Test added, but no assert.

server/test/lib/scene/scene.execute.test.js
callemand: Yes, I know, it was just to test the codecov return.

atrovato (approved):
Ok, approved 👍

Pierre-Gilles:
Hey! Thanks for your PR. 

I'm not really sure this is the behavior we want.

It's weird to me that a disabled scene could not be tried...

To me, the "disable scene" feature is more a "disable trigger for this scene" feature, not a "disable the scene completely".

I think people want to disable a scene while they are still building it, or if they want to pause the automatic start of the scene (while they are in holiday for example). But still, I think that people still want to be able to do manual runs. 

What do you think about that?

atrovato:
That's right, we already talked about the fact we need to keep the "test" button,
but since you added the test on active in `scene.execute.js`, test is disabled.
Maybe the "active" test should be only on triggers (all triggers).

Pierre-Gilles:
Thanks for the development! 

I merged a modified version in https://github.com/GladysAssistant/Gladys/pull/1229

PR #1085, opened 2021-03-01
chore: Update Sequelize to v6

cicoub13:
From v4 to v5:
- operatorsAliases is deprecated
- sequelize.close is not needed anymore, but I let it by security

From v5 to v6:
- nothing particular

Due to https://github.com/sequelize/sequelize/issues/11225, I added aliases for `created_at` and `updated_at`here : https://github.com/GladysAssistant/Gladys/pull/1085/files#diff-183dfcbb2c260fee7e3468389eaf038313e4ef51d52a3d8627590ba8daab4ce8R9

Can you please double check `server/package-lock.json` ? I don't know how to update only **sequelize** ?


Pierre-Gilles:
> Due to this issue sequelize/sequelize#2641, we used created_at and update_at in Javascript objects but like the issue is fixed, fields are now createdAt and updateAt (columns in database stay the same)

Are you 100% sure the "underscored" option could not work like before?

We use snake_case for all database columns, I'm not sure we want to move to camelCase. It's weird to mix snake_case and camelCase in same objects...

atrovato (approved):
Good enought for me.
Tested with existing database, and with blank one. All migrations are well applied.

cicoub13:
> Are you 100% sure the "underscored" option could not work like before?
> 
> We use snake_case for all database columns, I'm not sure we want to move to camelCase. It's weird to mix snake_case and camelCase in same objects...

Sorry. I mixed two issues. With underscore option, everything stays snake_case.

Pierre-Gilles:
> sequelize.close is not needed anymore, but I let it by security

Ah, I don't know if you saw my last PR, my I added this: 

https://github.com/GladysAssistant/Gladys/blob/master/server/index.js#L29

You say it's not needed anymore with Sequelize v6 ?

From the doc, I see that it's still in the doc: 

Screenshot 2021-03-15 at 10 26 09

cicoub13:
I'm confuse. Documentation here https://sequelize.org/master/manual/connection-pool.html says that 

> Sequelize will set up a connection pool on initialization

and here https://sequelize.org/v5/manual/upgrade-to-v5.html#pooling that

> You no longer need to call sequelize.close to shutdown pool

But I let the close() I found in the shutdown file.

https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#instance-method-close

> Normally this is done on process exit, so you only need to call this method if you are creating multiple instances, and want to garbage collect some of them.

Pierre-Gilles:
They are talking about connection pooling only :) As we use Sqlite, there is no pooling (I guess), and I think closing the DB before stopping Gladys is safer to avoid letting the DB in a bad state

Pierre-Gilles (approved):

PR #1108, opened 2021-03-24
Change feature category

atrovato:
### Pull Request check-list

### Description of change

This change allows to modify feature category (from swith to light) by adding an `updateCategory` method on common feature update page.
Here, only Tasmota implements this feature for BINARY features which are not known as specific light devices (according to Tasmota definition).


atrovato:
@GladysAssistant/contributors 
Anyone to review this?
This allow to change device category according to conditions (dedicated to each service).

front/src/config/i18n/fr.json
```diff
@@ -1660,6 +1660,7 @@
     "nameLabel": "Nom",
     "roomLabel": "Pièce",
     "featuresLabel": "Fonctionnalités",
+    "featureCategoryLabel": "Categorie",
```
Pierre-Gilles: Catégorie

Pierre-Gilles (commented):
Except this very small typo, I'm good on this PR, but I haven't tried it as I don't have Tasmota devices! 

I would like to have some feedback on someone who really tried it before merging :)

front/src/config/i18n/fr.json
atrovato: Thanks, fixed.

atrovato:
I will generate a docker image and ask for community for feedbacks.

atrovato:
Community exchanges start here https://community.gladysassistant.com/t/editer-le-type-de-feature-des-devices-philips-hue/5806/31?u=alextrovato

front/src/routes/integration/all/tasmota/edit-page/index.js
```diff
@@ -2,14 +2,25 @@ import { Component } from 'preact';
...
+
+    return false;
+  };
+
   render(props, {}) {
     return (
       <TasmotaPage user={props.user}>
         <UpdateDevice
           {...props}
+          canEditCategory={this.canEditCategory}
```
Pierre-Gilles, line 23: Just one question on this, why is it a function and not a boolean?

We have all the data we need here to calculate if the device feature is editable here, so why not pass just the final result?

atrovato: This is a method because the entrypoint is the device, but the option is on feature.
Each feature should be able to determine it's own capability to modify its category.

That's a bit heavy, I admit, but more precise.
And this can open the option on all features, type by type according to device model... and more and more.

But anyway, I admit this is not mandatory here.

Pierre-Gilles: You're right, I didn't see this. It makes sense indeed!

Pierre-Gilles (approved):
As @Terdious tried it already with success on the forum, and I'm good on the review, I'm merging this 🙏

Thanks for this PR, I'm looking forward to adapt this to other services!

Pierre-Gilles (approved):

PR #1116, opened 2021-04-01
fix(zigbee2mqtt): Fix USB listing

cicoub13:
comVid (vendorId) is sometimes empty from serialports/list library

https://community.gladysassistant.com/t/gladys-assistant-4-2-est-disponible-avec-la-compatibilite-zigbee2mqtt/6225/20

### DIFF
--- a/front/src/routes/integration/all/zigbee2mqtt/settings-page/SettingsTab.jsx
+++ b/front/src/routes/integration/all/zigbee2mqtt/settings-page/SettingsTab.jsx
@@ -51,9 +51,11 @@ const SettingsTab = ({ children, ...props }) => (
               {props.usbPorts &&
                 props.usbPorts.map(
                   usbPort =>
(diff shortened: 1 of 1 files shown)


cicoub13:
With `/run/udev` mounted in container (full detail). It will work like this with the new image from @VonOx

Capture d’écran 2021-04-01 à 19 35 08

Without `/run/udev` mounted in container (only name). It will be like this for **actual installations**.

Capture d’écran 2021-04-01 à 19 37 04

Docker image tested on my RPI3

VonOx:
OK great you manage both case, we need to update the doc too ( I will) 

Edit: Updated te documentation https://github.com/GladysAssistant/v4-website/pull/67

VonOx (approved):

VonOx (approved):

Pierre-Gilles (approved):

PR #1124, opened 2021-04-05
Allows a scene to trigger another scene

### LINKED ISSUE #797: Scene: be able to start a scene from a scene
⚠️  The user shouldn't be able to select the same scene !!

rob-mccann:
### Pull Request check-list

### Description of change

This change adds an action to scenes which triggers another scene.

image

A few assumptions have been made here which I'd be happy to revisit based on feedback:

- There's no validation that a scene exists in the backend
- A scene can only trigger another scene once across all action groups
  - In the cyclical scenario (for example,  Scene A that triggers Scene B that triggers Scene A), each scene will be called once only and a warning message logged. This is handled at execution time by remembers which scenes were already called.
  - There's no cyclical checks/validation done when a scene is created at the moment; I suggest this is handled in a separate PR to make this easier to review and in the interest of making it incrementally better :) 
- At the moment, there isnt trigger shown in the scene being called

If you're happy with the approach, I'll go ahead and update the docs.

Closes #797


server/lib/scene/scene.execute.js
```diff
@@ -8,11 +8,15 @@ const logger = require('../../utils/logger');
...
  * sceneManager.execute('test');
  */
-function execute(sceneSelector, scope) {
+function execute(sceneSelector, scope = {}) {
   try {
     if (!this.scenes[sceneSelector]) {
       throw new Error(`Scene with selector ${sceneSelector} not found.`);
     }
+
+    scope.sceneCircularDependencyPreventionList = scope.sceneCircularDependencyPreventionList || new Set();
```
atrovato: Hum... I'm not really fan of `sceneCircularDependencyPreventionList` name, a bit long, and is a Set, not a list.
Maybe just `executedScenes`, or `relativeScenes`?

server/lib/scene/scene.actions.js
```diff
@@ -103,7 +103,16 @@ const actionsFunc = {
       }
       setTimeout(resolve, timeToWaitMilliseconds);
     }),
-  [ACTIONS.SCENE.START]: async (self, action, scope) => self.execute(action.scene, scope),
+  [ACTIONS.SCENE.START]: async (self, action, scope) => {
+    if (scope.sceneCircularDependencyPreventionList.has(action.scene)) {
```
atrovato: That's nice to check circular execution, but how can we alert user in this case?
Maybe we should add a "warning" message in the scene editor to explain that each depend scenes will be run only once.

atrovato (commented):
Tested locally, looks fine.
2 things:
 - we may want to add a message in the UI to tell the user that scenes won't loop
 - execution conditions of the child scene are ignored (to execute it), maybe the user needs to know it too

what about a ℹ️ tooltip on scene selection box?

server/lib/scene/scene.execute.js
rob-mccann: Naming things is hard 😓 

`scenesAlreadyExecuted` perhaps

server/lib/scene/scene.actions.js
Pierre-Gilles: I think the box in the UI could have a little description explaining that circular execution is prevented ? Just before the select

server/lib/scene/scene.execute.js
atrovato: I know... naming... `alreadyExecutedScenes` instead.

rob-mccann: done

server/lib/scene/scene.actions.js
rob-mccann: I've added a message - any suggestions for copy and/or 🇫🇷 ?

atrovato: According to your message `"Scenes will only be called once in a chain, even if they're added multiple times. Trigger conditions will be ignored."`
"Chaque scène ne sera exécutée qu'une seule fois, et les conditions d'exécution seront ignorées."

@Pierre-Gilles ?

Pierre-Gilles: I would say in french: 

"Les conditions de déclenchements de la scène appelée seront ignorées. Si une scène appelle une autre scène, et vice-versa, Gladys ne tombera pas dans une boucle infinie: chaque scène ne sera exécutée qu'une seule fois. "

Pierre-Gilles (changes requested):
When I click on the start scene button: 

Screenshot 2021-04-22 at 12 20 09

I got this error: 

Screenshot 2021-04-22 at 12 20 30

I guess it's because `alreadyExecutedScenes` does not exist at this stage

Pierre-Gilles (approved):
Looks good to me this time! Thanks for the PR, great work 🙏

PR #1125, opened 2021-04-05
Remove extra console.log

rob-mccann:
### Pull Request check-list

### Description of change

Just removes a console.log that's currently causing a lint warning


front/src/actions/profile.js
```diff
@@ -175,7 +175,6 @@ function createActions(store) {
         });
         route('/dashboard/settings/user');
       } catch (e) {
-        console.log(e);
```
atrovato, line 178: Don't we prefer to log this has error (or warn)?

Pierre-Gilles: I think we could just replace this with a console.error() yes 

Because as Gladys is an open-source product without tracking, we sometimes ask users to show the browser logs to debug (no sentry, no analytics tool here), and console errors are useful

Pierre-Gilles: We try to avoid asking users for logs at all cost (not very user-friendly), but sometimes it's easier to find what's wrong ^^

rob-mccann: should this be `logger.error`?

atrovato: No, in front side, we don't have access to server logger, we use `console.error`

rob-mccann: Ah yes sorry - still new to the codebase. I'll change to console.error and add an eslint ignore and comment

atrovato (approved):

front/src/actions/profile.js
Pierre-Gilles: No need for an eslint ignore, console.error is allowed :)

Pierre-Gilles: Ah yes sorry, didn't see your change. Looks good to me

Pierre-Gilles (approved):

PR #1127, opened 2021-04-06
feat(zigbee2mqtt): Fix all lights

cicoub13:
After feedback from community

- Fix models
- Fix brightness feature (not read_only)


VonOx:
Oh good catch, that explain why slider is not available on dashboard.

atrovato (approved):

Pierre-Gilles (approved):
Looks good to me! Thanks for the fixes :)

PR #1082, opened 2021-02-28
Brightness device feature Dashboard

VonOx:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [x] If your changes affects code, did your write the tests?
- [x] Are tests passing? (`npm test` on both front/server)
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
~~- [ ] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)~~
~~- [ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~
~~- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~
- [x] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

This PR add brightness slider to dashboard

image

PR #1137, opened 2021-04-09
Fix incorrect attribute

atrovato:
Something weird in dashboard boxes.
Some use `feature.last_value`, some uses `feature.lastValue`...

Did I miss something?
Or this fix looks good?

(searched all `lastValue` occurrences and replace with `last_value`)


Pierre-Gilles:
I'm quite surprised, it seems you're right but I can't understand why it's working right now... (or maybe it's not working!)

@VonOx I think you coded the MultiLevel no? any idea?

Pierre-Gilles:
Indeed, @VonOx you coded the Brightness box: 

https://github.com/GladysAssistant/Gladys/pull/1082/files#diff-91ffef4861a07c96af49b03dbbba5c76e5f0d2826031cbd119642cec2eb5a3a3R13

Is it working or did you code it blindly?

Pierre-Gilles:
From my test, it seems you're right @atrovato, it's broken right now.

VonOx:
Yeah it works, but weirdly right now.

Slider is always at 50% when I switch light on and when I slide the button sometimes it stuck.

I have implemented brightness based on multi level but tested with my own devices.

I can't explain the behavior... Sorry

Pierre-Gilles:
It works to control devices, but the slider is always at 50% (because lastValue doesn't exist). I investigated, it never worked. 

but @atrovato, I don't think your PR solves the issue. I have a feeling the slider component never worked, even with the right variable. It needs a little bit of work to understand how to make the slider component work.

Thanks for finding this Alex!

VonOx:
Brightness on my lights physically change so this is weird.

Pierre-Gilles:
@VonOx nothing weird, the control works indeed, it's the display which is broken

VonOx:
> @VonOx nothing weird, the control works indeed, it's the display which is broken

Ok did'nt understand, my bad.

Pierre-Gilles:
@atrovato My bad, your PR solves the issue!

Pierre-Gilles (approved):
It works on my side! Thanks!

atrovato:
I just read all... quite funny :)

PR #1138, opened 2021-04-09
Single range box

atrovato:
Use single Range feature dashbox feature row.

Based on https://github.com/GladysAssistant/Gladys/pull/1137


front/src/components/boxs/device-in-room/DeviceRow.jsx
```diff
@@ -1,75 +1,32 @@
...
+  const elementType = ROW_TYPE_BY_FEATURE_TYPE[props.deviceFeature.type];
+
+  if (!elementType) {
+    // if no related components, we return nothing
+    return null;
   }
-  // if not, we return nothing
-  return null;
+
+  return createElement(elementType, props);
```
Pierre-Gilles, line 29: very nice, didn't know about createElement :) I learnt something today!

Pierre-Gilles (approved):
Tested locally, looks good to me

PR #1139, opened 2021-04-09
Television device and features

atrovato:
Based on https://github.com/GladysAssistant/Gladys/pull/1138

Add television category, features, icons, and some dashboard actions.


rob-mccann:
I was just working on something similar. I wonder if creating a generic "media player" device might allow us to consolidate audio players too?

Here's my work in progress for reference https://github.com/GladysAssistant/Gladys/compare/master...rob-mccann:media-device

front/src/components/boxs/device-in-room/EditDeviceInRoom.jsx
```diff
@@ -13,7 +13,8 @@ import actions from '../../../actions/dashboard/edit-boxes/editDevicesInRoom';
 const SUPPORTED_FEATURE_TYPES = [
   DEVICE_FEATURE_TYPES.LIGHT.BINARY,
   DEVICE_FEATURE_TYPES.LIGHT.COLOR,
-  DEVICE_FEATURE_TYPES.LIGHT.BRIGHTNESS
+  DEVICE_FEATURE_TYPES.TELEVISION.CHANNEL,
```
rob-mccann, line 17: Not sure if you want to include it but televisions also have a source (HDMI1, HDMI2, TV etc)  in addition to a channel

front/src/config/i18n/en.json
```diff
@@ -1489,6 +1489,36 @@
...
+      "guide": "Guide",
+      "menu": "Menu",
+      "volume-up": "Volume up",
+      "volume-mute": "Mute",
+      "volume-down": "Volume down",
+      "volume": "Volume",
+      "channel-up": "Channel up",
+      "channel-previous": "Previous channel",
+      "channel-down": "Channel down",
+      "tools": "Tools",
```
rob-mccann, line 1518: is this common to all TVs?

rob-mccann: ah I see below - perhaps this should be `Settings`?

atrovato:
I think we have to keep them separated, as tv, audio, or dvd player might not do the same... but they share 80% of their features...
Let ask to @GladysAssistant/contributors

rob-mccann:
Is that something we could cover off by having a device that is tagged with specific "features"?

Just to add a third type in there is a games console where the media might be a video game :)

atrovato:
Ok, you're work looks really good, so I only need more features related to TV device.
This one is a draft for now, I will notice forum that you're working on a more complete media service.

(this is PR is a split of a bigger one, to manage IR remote with broadlink, I can continue broadlink service without it for now)

atrovato:
After more reflexion, I think we need to have 'tv', 'game', 'music' categories to easily and clearly distinguish it.

rob-mccann:
I'm still feeling my way through the codebase so definitely appreciate ideas and support!

front/src/config/i18n/en.json
atrovato: On my Samsung remote, I see "TOOLS", but settings is acceptable.
I'll wait when this is ready for review.

Pierre-Gilles:
Hello! Just wanted to know the status of this PR ? :)

atrovato:
I close it for now.

PR #1143, opened 2021-04-11
WIP: Allow services to receive websocket events

rob-mccann:
### Pull Request check-list

### Description of change

This allows (future) services and other areas of code to listen for websocket events.

```
gladys.event.on(EVENTS.WEBSOCKET.RECEIVE, (event) => {
  // check event.type is the desired event and act accordingly; can reply using the below if desired:
  gladys.event.emit(EVENTS.WEBSOCKET.SEND, {
    type: WEBSOCKET_MESSAGE_TYPES.SOMETHING,
    userId: event.userId,
    payload: {}
  });
})
```

### DIFF
--- a/server/api/websockets/index.js
+++ b/server/api/websockets/index.js
@@ -122,7 +122,10 @@ function init() {
           }
           break;
         default:
-          logger.debug(`Message type not handled`);
+          this.gladys.event.emit(EVENTS.WEBSOCKET.RECEIVE, {
+            userId: user.id,
+            ...parsedMessage,
+          });
       }
     });
     setTimeout(() => {
--- a/server/test/websockets/index.test.js
+++ b/server/test/websockets/index.test.js
@@ -1,7 +1,8 @@
 const EventEmitter = require('events');
 const WebSocket = require('ws');
-const { assert, fake } = require('sinon');
+const { assert, fake, spy } = require('sinon');
 const WebsocketManager = require('../../api/websockets');
+const { WEBSOCKET_MESSAGE_TYPES, EVENTS } = require('../../utils/constants');
 
 describe('Websockets', () => {
   it('should send message to all users', () => {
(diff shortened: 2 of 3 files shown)


Pierre-Gilles:
Hi! Thanks for this PR. 

Why do you need to receive websocket event in a service? Can't a simple HTTP call be enough for front to service communication?

rob-mccann:
I'm building a service (LG TV) that relies on a socket being kept open with the device temporarily whilst authenticating (the user needs to accept a prompt on the TV). Using web socket from `front` makes encapsulating the open connection a lot easier.

Pierre-Gilles:
Ok! Why don't you just keep the open socket as a class attribute of your service and access it like `this.socket.` ?

I'm interested to see why using websockets could help here :)

rob-mccann:
No problem at all :) happy to try different approaches too if this doesnt make sense!

using `this.socket` I think would cause us to hit some concurrency issues where we keep one socket open for all devices/users which isn't what I wanted to do here (where I want one user to auth 1 device and another user may well want to auth a different device at the same time).

Here's the WIP implementation for the TV service and where I listen to the socket event:
https://github.com/rob-mccann/Gladys/blob/media-device/server/services/lgtv/lib/tv/promptListener.js#L35

Pierre-Gilles:
Why do you think it could cause concurrencies issues? I think it's even better to keep one connection and avoid connecting 10 times to the TV (I don't know how webOS tv works, just a guess)

I'm not sure this websockets workaround is the solution to your first problem. I think we want to keep a "stateless" REST API first for "Front -> Back" requests, as it would allow us in the future to let people use Gladys as a simple HTTP open API. 

Having different API would make it harder for people to integrate with Gladys and build third-party clients.

I'm still open to the discussion of course, just want to be sure that this is really necessary for your integration :)

rob-mccann:
Sure - rigour around this things is always helpful. I'll give it a try without using the socket and we can compare the two :)

PR #1148, opened 2021-04-17
Add ability to use request body response in scene

Pierre-Gilles:
### Pull Request check-list

### Description of change

The user is now able to use the try a request in scene, and use the API response in other boxes of the scene

Example with Bitcoin price:

https://user-images.githubusercontent.com/7365207/115101178-8cab6e80-9f74-11eb-8fb1-bde193514a1b.mp4

### DIFF
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -932,7 +932,10 @@
         "bodyPlaceholder": "Body",
         "headersLabel": "Headers",
         "headersKeyPlaceholder": "Name",
-        "headersValuePlaceholder": "Value"
+        "headersValuePlaceholder": "Value",
+        "responseData": "Response",
+        "tryButton": "Try request",
+        "requestError": "An error happened while making the request. Are you sure parameters are right?"
       }
     },
     "actions": {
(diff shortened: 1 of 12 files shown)


Pierre-Gilles:
Two questions:

- What do we need to do if internet is down?
- What do we need to do if API response changed compared to the first time?

atrovato:
Nice demo, 2 questions, really ?

> What do we need to do if internet is down?

Be able to catch default 4xx / 5xx errors, and maybe replace values by a magic i18n word `-- error --`?

> What do we need to do if API response changed compared to the first time?

If you use `get-value`, add a default magic value `-- data not found --`?

Pierre-Gilles:
> Be able to catch default 4xx / 5xx errors, and maybe replace values by a magic i18n word -- error --?
> If you use get-value, add a default magic value -- data not found --?

I don't think we want that, sending a message with a variable is just an example, but variables can be used in the "continue only if" block and compared with numbers, so it makes no sense no replace variables with default values.

I think we might want to do nothing for now, and see with usage if we want to add an option in the action box like "stop the scene if the request fails" 

Edit: See my message here about why it's not the default option: https://community.gladysassistant.com/t/utiliser-une-reponse-api-externe-dans-une-scene/6199/8?u=pierre-gilles

PR #1174, opened 2021-04-27
Replace moment by dayjs in caldav service

bertrandda:
### Pull Request check-list

### Description of change

I remove moment from server deps and use dayjs in caldav service.


Pierre-Gilles:
Thanks for the PR! :)

PR #1178, opened 2021-05-02
Support Chinese language

bulolo:
Support Chinese language


Pierre-Gilles:
Hi! 

Thank your very much for your interest in Gladys, and for your PR 🙏  

We discussed with other contributors, and we agree that at the moment we don't have the ressources to add more languages to Gladys Assistant. 

Adding a new language is not just translating the product once: it's maintaining this set of translations over the years, on a weekly basis, on the product & in the documentation.

It's a lot of work & it means that for each new features, we need to spend some time making sure it is translated & works in all languages. 

Right now, this open-source project is still small, we have very little ressources and we want to focus on adding new features.

On top of that, as none of us speak Chinese, I have the feeling that this language, if merged, will just be unmaintained afterwards, and Gladys in Chinese will just be slowly broken.

It's super hard for me to say "no" to such an amazing PR (I know it's a lot of work), but I think we need to stay focused on our mission. 

I hope you understand our position 🙂

Maybe you could maintain your own fork in Chinese language, and maybe in the future when we'll have more ressources, we'll merge the Chinese fork in Gladys? 🙂

Thanks again for your work 🙏

PR #1221, opened 2021-06-27
zWave Dimmer Integration

sescandell:
### Pull Request check-list

- ~[ ] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)~
- ~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~
- ~[ ] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~

### Description of change

Manage zWave Dimmer devices (devices supporting Command Class Switch MultiLevel (38)).


sescandell:
@Pierre-Gilles PR Ready dans le principe.

Basic objective is to correctly manage Dimmers. I made dev' and tested them with Fibaro Walli Dimmer. Specificity of these devices is they are working with the V4 spec' for multilevel switch. In V4: there is a new INDEX to manage the level of the device (there is the level and the target). This PR initially fix this. By the way I fixed and improved zWave tests.

Compared to the previous PR: no inheritance implementation used here. I'm convinced that's not a good idea, and we should use it for CommandClasses. Without it, we need to think about some mandatories functions and duplicate some of them (look at methofs getMinMax, getTransformedValue, ...). For now, with only 2 implementations, no inheritance is still fine. I'm wondering how it will evolve with time... let's see. We can still update this afterward.

"Why a new PR?": just to make it simpler instead of making a rebase given all modifications since (especially: multi-instances part).

Tell me.

_we can switch in French to make things easier if you want_

front/src/components/boxs/device-in-room/device-features/MultiLevelDeviceFeature.jsx
```diff
@@ -40,7 +40,7 @@ const MultiLevelDeviceType = ({ children, ...props }) => {
             value={props.deviceFeature.last_value}
             onChange={updateValue}
             class="form-control custom-range"
-            step="1"
+            step={props.deviceFeature.step}
```
Pierre-Gilles: You are adding a step attribute to the deviceFeature object, but I don't see it in the data model. It's not stored?

Those changes affects the whole Gladys, not just the Zwave integration.

server/package.json
```diff
@@ -12,8 +12,8 @@
     "test": "cross-env NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/**/*.test.js\" --exit",
     "test-service": "cross-env SQLITE_FILE_PATH=/tmp/gladys-test.db NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/services/$npm_config_service/*.test.js\" \"./test/services/$npm_config_service/**/*.test.js\" --exit",
     "coverage": "nyc npm test && nyc report --reporter=text-lcov > coverage.lcov",
-    "prettier-check": "prettier --check '**/*.js' '**/*.json'",
-    "prettier": "prettier --write '**/*.js' '**/*.json'",
+    "prettier-check": "prettier --check \"**/*.js\" \"**/*.json\"",
+    "prettier": "prettier --write \"**/*.js\" \"**/*.json\"",
```
Pierre-Gilles, line 16: Why did you change this? It doesn't seem related to the PR

server/services/zwave/lib/comClass/defaultCommandClass.js
```diff
@@ -0,0 +1,83 @@
+const { transformValue, normalizeValue } = require('../utils/valueBinders');
+
+const DefaultCommandClass = function DefaultCommandClass() {};
+
+DefaultCommandClass.prototype = {
```
Pierre-Gilles: Seems it's something remaining from the previous implementation

server/services/zwave/lib/comClass/switchMultilevelCommandClass.js
```diff
@@ -0,0 +1,142 @@
+const logger = require('../../../../utils/logger');
+const { COMMAND_CLASSES, INDEXES } = require('../constants');
+const DefaultCommandClass = require('./defaultCommandClass');
+
+const SwitchMultilevelCommandClass = function SwitchMultilevelCommandClass() {};
```
Pierre-Gilles: Same here

server/services/zwave/lib/utils/bindValue.js
```diff
@@ -1,27 +0,0 @@
...
- * @param {number} comclass - The comclass.
- * @param {number} index - The index.
- * @returns {number} Return the value adapted.
- * @example
- * const value = bindValue(6, 0x4501, 12, 1);
- */
-function bindValue(value, productId, comclass, index) {
-  // Zipato Mini Keypad RFID
-  if (
-    productId === '0x4501' &&
```
Pierre-Gilles, line 16: I know that this file was not used, but it's actually something useful and we talked about it recently on the forum:

https://community.gladysassistant.com/t/integration-zipato-mini-keypad/6325/18?u=pierre-gilles

Pierre-Gilles (changes requested):
Hi! Thank you very much for this PR, it's great work 🙏

I wrote a few feedbacks on the PR.

About the future of Z-Wave in Gladys, we talked about it recently on the forum, and the future of Z-Wave is probably in migrating to ZWaveJS/ZwaveJS2Mqtt. 

OpenZwave is not updated anymore, and it's buggy in many ways/painful to use/develop with.

If in the future you want to help us migrate to ZwaveJS2Mqtt, let us know in the forum! 🙂

And thanks again for the work done here!

server/package.json
sescandell: To make it work on Windows

server/services/zwave/lib/utils/bindValue.js
sescandell: I'll double check this.

Pierre-Gilles: It was not used, but we plan to use this! @tagcashdev worked on it

server/package.json
Pierre-Gilles: 👍 Ok I need to check that it still works on other platform too then

server/services/zwave/lib/comClass/defaultCommandClass.js
sescandell: What's the matter here? 🧐

Are you expecting something like this here:
```
function func1() {}
function func2() {}
function func3() {}

export default {
    func1,
    func2,
    func3
}
```
?

server/services/zwave/lib/utils/bindValue.js
sescandell: Ok, I'll bring back this code in "the new way" here

sescandell: Not sure where this was planned to be used... I supposed it was on valueChanged. The "code" is back at least. If we need to move it, we can.

front/src/components/boxs/device-in-room/device-features/MultiLevelDeviceFeature.jsx
sescandell: Reverted

server/services/zwave/lib/comClass/defaultCommandClass.js
sescandell: I removed the prototype part... let me know if the matter is somewhere else.

Pierre-Gilles: Yes! You are just coding little helper functions, no need to create classes for that (those are stateless). The simpler, the better.

server/test/services/zwave/lib/nodesExpectedResult.json
```diff
@@ -1770,7 +1771,8 @@
         "unit": "celsius",
         "has_feedback": true,
         "min": 0,
-        "max": 0
+        "max": 0,
+        "step": 1
```
Pierre-Gilles: There are still "step" attribute in this file

server/services/zwave/lib/comClass/alarmCommandClass.js
```diff
@@ -0,0 +1,41 @@
+const { COMMAND_CLASSES, INDEXES } = require('../constants');
+const DefaultCommandClass = require('./defaultCommandClass');
```
Pierre-Gilles: It's a little weird to have a capitalized require that is not a class.

```
const defaultCommandClass = require('./defaultCommandClass');
```

server/services/zwave/lib/comClass/defaultCommandClass.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {number} index - Index of the value.
+ * @param {number} instance - Instance of the value.
+ * @param {any} value - Value to transform.
+ *
+ * @returns {any} The normalized value.
+ *
+ * @example comClass.getNormalizedValue(node, 110, 0, 1, 50);
+ */
+function getNormalizedValue(node, comClass, index, instance, value) {
+  return normalizeValue(node.classes[comClass][index][instance].type, value);
```
Pierre-Gilles: Can I ask you a question, are you a Java developer? 😄 

I begin to see how you think: everything needs to be "class" (no offense), and it leads to lots of functions like this one (with no real code inside). 

I know that you've maybe worked like that for years so don't get offended, but I feel that the code is hard to read as  the user needs to jump from file to file, and most code is useless like this one.

All this could have been a one-line of not duplicate code in one helper file, no ?

I know that you already said that "the code will be complex next", but in my experience working 7 years on Gladys, other integrations are completely fine, and coded in a functional way.

Pierre-Gilles (changes requested):

server/services/zwave/lib/comClass/defaultCommandClass.js
sescandell: ahah, no worries, no offense ;)

No, I'm not a Java developper :) I'm using multiple languages (recently mainly on JS, C#, PHP), but I didn't play with Java for a long time.

> All this could have been a one-line of not duplicate code in one helper file, no ?

You could always do everything in one file. It's just a POV about separation and responsabilities. Is it the responsability to a CommandClass to know how to normalize a value (something linked to Gladys itself)? Probably not. The CommandClass needs to normalize a value, but doesn't need to know how. Putting this dedicated code in a dedicated file makes it easier to test (tests are focused on this class). Intention of the class/file is clearly defined. Yes, I can put this logic in the DefaulCommandClass, but it will make it bigger and knowing some concepts not really related to a CommandClass (so a file harder to read, harder to test as it should tests some things not directly related to CommandClass and so on). 

> the code is hard to read as the user needs to jump from file to file

It depends what you are looking for. If you need to deep dive in what's going on, yes you need one more file to open. But if you're clearly focused on CommandClass behavior, you don't need to deep dive in this part of code. It's out of the CommandClass concept and you can so assume this external file makes its job as it should. It's like the `emit` function. I don't need to know how it's done. i'm using it, assuming it's doing its job as expected so I can move forward. It makes it easier to read.

Finally, it might be more a habit issue than a real one. For example, when I firstly came on Gladys, as a regular JS developper, I've been very surprised about how the backend is structured: I think separating all functions of a same class in dedicated files doesn't help at all developpers (neither IDE). It's really hard and uncommon to navigate from functions to functions this way (Imagine moving on ES6+ syntax with real classes: you would not explode your functions in different files like it's actually done). But, finally, even like that, the product is working pretty well.

If your concern is this `valueBinders` file, I can bring it back to one file... I will have some troubles to sleep correctly, but it should be fine with time ;)

server/test/services/zwave/lib/nodesExpectedResult.json
sescandell: Fixed

server/services/zwave/lib/comClass/alarmCommandClass.js
sescandell: Fixed

server/services/zwave/lib/comClass/defaultCommandClass.js
Pierre-Gilles: I'm just concerned about consistency around the project, we don't work like that on other integrations :) 

This debate is hard as we are both opinionated in two opposite directions: you're mostly pro-class/inheritance, and I'm mostly pro-function and against inheritance.

It's an eternal debate in programming, and there are many articles on the pro/cons of each paradigm. Any of us won't change opinion I guess aha 😄 

To give an example of others integration that did a functional approach that worked pretty well: The Philips Hue integration or the Zigbee2mqtt integration

Both have a "models" folder (See [Philips Hue models](https://github.com/GladysAssistant/Gladys/tree/master/server/services/philips-hue/lib/models) or [zigbee2mqtt models](https://github.com/GladysAssistant/Gladys/tree/master/server/services/zigbee2mqtt/model) ) . 

Each file is a function that returns a formatted device. When needed, we use helper functions or a set of constants to get min/max, that kind of stuff.

It's just my opinion and I don't want to convert you, but I find this kind of implementation easier to read, code & maintain. 

When using inheritance, class are tightly coupled together and refactoring is harder when adding a new device. One new developer needs to know all the classes before adding a new device, it's prone to more refactoring errors. 

> For example, when I firstly came on Gladys, as a regular JS developper, I've been very surprised about how the backend is structured: I think separating all functions of a same class in dedicated files doesn't help at all developpers (neither IDE).

This one is another debate, and your point is definitely valid: it's harder to autocomplete with the "one file per function" approach. But on the other hand, it's cleaner IMO to split the functions in multiple files, make it's easier to access one function by typing cmd + p and the name of the function (ex: cmd + P + `device.setValue` ) and directly accessing the function (which you can't if you have one giant files with all functions). Believe me it's not a simple decision I took, I thought about it a lot when implementing this :p

PR #1248, opened 2021-07-23
Add the ability to vizualize device states on the dashboard

Pierre-Gilles:
### Pull Request check-list

### Description of change

Screenshot 2021-10-22 at 12 58 38


VonOx:
@pierre-gilles Can you provide dataset for testing?

Pierre-Gilles:
@VonOx I'm not sure if it's the goal, if I share my dataset, you will probably have the same results as I'm having locally: and it works fine on my side.

I want to see with your datasets!

VonOx:
Ok i will import my production data

front/src/components/boxs/chart/EditChart.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  render(props, { selectedDeviceFeaturesOption, deviceOptions, loading }) {
+    return (
+      <BaseEditBox {...props} titleKey="dashboard.boxTitle.chart">
+        <div class={loading ? 'dimmer active' : 'dimmer'}>
+          <div class="loader" />
+          <div class="dimmer-content">
+            <div class="form-group">
+              <label>
+                <Text id="dashboard.boxes.devicesInRoom.editRoomLabel" />
```
VonOx: Wrong label here

VonOx (changes requested):

VonOx:
Selecting device feature is filling Title box even if a value was already entered

image
image

I can't identify in which room theses values are


## Background tasks view

Relative time problem.
French/English mixed.
Duplicate entry.

As a user what I need to do to fix failed aggregation ?

image

## last 24 hours charts and last hour chart

image

Tooltip must indicate hours not date

## Ideas

* It would be usefull to combine multiple device feature on the same chart
* Hability to select a color
* Display min / max on periode or like this
image





PS: I will update this comment while testing

Pierre-Gilles:
> Selecting device feature is filling Title box even if a value was already entered

Doesn't it make sense to replace the title if the device feature selected changed?

> I can't identify in which room theses values are

You can put the name in the box title: 

Screenshot 2021-09-24 at 09 30 03

Screenshot 2021-09-24 at 09 30 10

Or maybe you were talking about the default? 

I was fearing putting device name + room name would make it too long by default.

> Relative time problem.
French/English mixed.

Ok, will fix. 

> As a user what I need to do to fix failed aggregation ?

You're right, we need to put somewhere the error message that happened. 

As a user, you won't be able to fix it, but just communicate the bug to us on the forum/Github. 

But I don't expect we'll see some fails, this error happened because it's still WIP, but in production it's like the rest of Gladys, everything is carefully tested and there is no reason it'll fail more than something else :) 

> It would be usefull to combine multiple device feature on the same chart 

Yes, definitely.

> Hability to select a color

Why not indeed, it makes even more sense with multi-devicefeature.

> Display min / max on periode or like this

Yes definitely.

> PS: I will update this comment while testing

Thanks for your test and precious feedback.

I'll spend some time on fixing this and will get back to you! 🙏

front/src/components/boxs/chart/EditChart.jsx
Pierre-Gilles: Fixed

front/src/components/boxs/chart/ApexChartComponent.jsx
```diff
@@ -0,0 +1,366 @@
...
+    if (this.props.chart_type === 'area') {
+      options = this.getAreaChartOptions();
+    } else if (this.props.chart_type === 'line') {
+      options = this.getLineChartOptions();
+    } else if (this.props.chart_type === 'stepline') {
+      options = this.getStepLineChartOptions();
+    } else if (this.props.chart_type === 'bar') {
+      options = this.getBarChartOptions();
+    } else {
+      options = this.getAreaChartOptions();
```
atrovato, line 112: This file is kind long, maybe we can extract each case in separated files.

front/src/components/boxs/chart/EditChart.jsx
```diff
@@ -0,0 +1,254 @@
...
+    } catch (e) {
+      console.error(e);
+      this.setState({ loading: false });
+    }
+  };
+
+  constructor(props) {
+    super(props);
+    this.props = props;
+    this.deviceFeatureBySelector = new Map();
```
atrovato, line 132: Are you sure you want to use `Map`? Object is not enough?

front/src/config/i18n/en.json
```diff
@@ -245,6 +246,29 @@
...
+        "defaultInterval": "Default interval",
+        "lastHour": "Last hour",
+        "lastDay": "Last 24 hours",
+        "lastSevenDays": "Last 7 days",
+        "lastThirtyDays": "Last 30 days",
+        "lastThreeMonths": "Last 3 months",
+        "lastYear": "Last year",
+        "noValue": "No values recorded on this interval",
+        "editNameLabel": "Enter the name of this box.",
+        "editDeviceFeaturesLabel": "Select the device you want to display here.",
```
atrovato: Ended with dot, but not other translations, is this expected?
(same on upper line)

front/src/config/i18n/fr.json
```diff
@@ -245,6 +246,29 @@
...
+      "chart": {
+        "defaultInterval": "Intervalle par défaut",
+        "lastHour": "Dernière heure",
+        "lastDay": "Dernière 24h",
+        "lastSevenDays": "Derniers 7 jours",
+        "lastThirtyDays": "Derniers 30 jours",
+        "lastThreeMonths": "Derniers 3 mois",
+        "lastYear": "Dernière année",
+        "noValue": "Pas de valeurs sur cet interval",
+        "editNameLabel": "Entrez le nom de cette box.",
```
atrovato: Same here for final dot (may be expected).

server/lib/device/device.calculateAggregate.js
```diff
@@ -0,0 +1,118 @@
...
+    const childProcess = spawn('node', [SCRIPT_PATH, JSON.stringify(params)]);
+
+    childProcess.stdout.on('data', async (data) => {
+      const text = data.toString();
+      logger.debug(`device.calculateAggregate stdout: ${data}`);
+      if (text && text.indexOf('updateProgress:') !== -1) {
+        const position = text.indexOf('updateProgress:');
+        const before = text.substr(position + 15);
+        const splitted = before.split(':');
+        const progress = parseInt(splitted[0], 10);
```
atrovato, line 87: Looks a bit confused here.
What is the expected `text` value?
If `text = "updateProgress:10:updateProgress"`, what for
`const [,progress] = text.split(':');`

server/package.json
```diff
@@ -10,6 +10,7 @@
     "postinstall": "node ./cli/install_service_dependencies.js",
     "pretest": "npm run eslint",
     "test": "cross-env NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/**/*.test.js\" --exit",
+    "test-without-lint": "cross-env NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/**/*.test.js\" --exit",
```
atrovato, line 13: Cheater! ;)

atrovato (commented):
Not sure all my comments request changes,
I didn't run it, Vonox did ;)
A bit complicated...

front/src/components/boxs/chart/ApexChartComponent.jsx
Pierre-Gilles: Every function contains the apexchart configuration, maybe I could export this in separate files you're right

front/src/components/boxs/chart/EditChart.jsx
Pierre-Gilles: Yes, I think it's always better to use those objects (Map, Set) when possible instead of using a bare object as a key value store, because Map expose some functions that are designed to work properly and that removes weird bugs.

When using an object, so many times some cases (null/undefined/0) are not well handled.

**Example:** 

```javascript
const myKeyValueStore = {};

keyValueStore['toto'] = 0
```

Later in the code: 

```javascript
if(keyValueStore['toto']) {
    //  0 is considered as "false" in Javascript, so it will not run
}
```

So what's the solution here?

You could do this: 

```javascript
if(keyValueStore['toto'] !== undefined) {
}
```

But then "null" is considered as "not existing" too, so we need to do:


```javascript
if(keyValueStore['toto'] !== undefined || keyValueStore['toto'] !== null) {
}
```

This is starting to become a bit long, and it's hard to grasp what we wanted to do.

Whereas, with a simple map: 

```javascript

const myKeyValueStore = new Map();
myKeyValueStore.set('key', 0);
if(myKeyValueStore.has('key')) {
// we have the guarantee that it handles all cases perfectly.
}
```

front/src/config/i18n/en.json
Pierre-Gilles: Nice catch, fixing

front/src/config/i18n/fr.json
Pierre-Gilles: Nice catch, fixing.

server/lib/device/device.calculateAggregate.js
Pierre-Gilles: > what for const [,progress] = text.split(':');

No it's not working, because I think we are receiving the data as blob of grouped logs, and sometimes it's grouped with other data.

Example, we can receive: 

```
const text = "SOME DATA RANDOM updateProgress:10:updateProgress OTHER DATA"
```

That's why I substr to remove the beginning of the string, then I split and remove the first part only.

server/package.json
Pierre-Gilles: My VSCode already run eslint on save + on paste, so it's really a waste of time to run lint every time I run tests :)

front/src/components/boxs/chart/ApexChartComponent.jsx
atrovato: This one is still pending.

Pierre-Gilles: Done

Pierre-Gilles:
To me the PR is ready 🚀🙂

I've updated the demo website: 

Screenshot 2021-10-22 at 12 58 38

Let me know if everyone is good to merge this!

PR #1253, opened 2021-08-01
Cypress coverage

atrovato:
Add cypress coverage on codecov report.

@Pierre-Gilles we have many work to do!


Pierre-Gilles:
Nice! Is there a way to better separate front coverage from back coverage?

The codecov message merge the two coverage and it looks like Gladys has a code coverage of only 60% without really knowing which coverage we are talking about :)

Screenshot 2021-08-09 at 10 23 42

atrovato:
Thank for feedbacks.
I'll take a look for this point.

atrovato:
@Pierre-Gilles 
I can propose this
image

Pierre-Gilles:
@atrovato Ok! Looks way better than before 🙂 On GitHub checks it'll be separate right ? 

So we still keep our rules on backend coverage change ?

atrovato:
I'll try removing some server tests to check for coverage check to fail.

atrovato:
Ok, you were right to ask for this point, coverage check is not applied anymore on server side.
Work again on it 😄

atrovato:
@Pierre-Gilles according to latest coverage, it looks good.

image

Once it's ok for you, I will revert "test deletion" commit.

Pierre-Gilles:
@atrovato 

The comment coverage still seems to mix both coverage:

Screenshot 2021-10-09 at 15 52 20

(I check, it's the coverage from 3 days ago)

You know if there is a way to get a notification message with only backend / or backend-front separated ?

atrovato:
Oww I didn't try this way.
Coverage are separated by flags, details on bottom table.

@VonOx can you help me to find a better way here?

VonOx:
@atrovato I'm working on

Maybe try with this codecov.yml

```yml
coverage:
  status:
    project:
      default:
        target: 90% #overall project/ repo coverage
      front:
        target: 60%
        flags:
          - front
      server:
        target: 90%
        flags:
          - server

# adding Flags to your `layout` configuration to show up in the PR comment
comment:
  layout: "reach, diff, flags, files"
  behavior: default
  require_changes: false  
  require_base: yes
  require_head: yes       
  branches: null

# New root YAML section = `flags:`
# This is where you would define every flag from your
# uploader, and update when new Flags added

flags:
  front:
    paths:
      - front
    carryforward: true
  server:
    paths:
      - server
    carryforward: true
```

atrovato:
Thanks @VonOx 
1st try currently building 🤞

atrovato:
@VonOx there is still only one report, one codecov comment... both front and server are still mixed.

VonOx:
I will merge your pr in my CI test repo, keep you updated

VonOx:
Codecov use only one comment, diff is global

But flags are working

https://github.com/VonOx/gladys-ci/pull/9#issuecomment-970697463

I think we need a master test to have a working diff ( codecov need a base to compare)

atrovato:
Thanls, I was already working with flags, but it looks it didn't please to Pierre-Gilles.

And do you think you can try with coveralls to compare multi-report / flag comment?

VonOx:
> And do you think you can try with coveralls to compare multi-report / flag comment?

Yep , no problem.

atrovato:
Not stale!

atrovato:
Stale...
It will be back...

Issue #1257, opened 2021-08-06
Zigbee2mqtt: Change a deprecated function to the device E1743 ( IKEA )


mikitouf:
Hi!

According to the FAQ ( [https://www.zigbee2mqtt.io/devices/E1743.html](url) ), the function click is deprecated , Can you change it  to the function action ( ON /OFF).

cicoub13:
It's already binded to `action` (feature button linked to field `action`). It doesn't work ?

https://github.com/GladysAssistant/Gladys/blob/master/server/services/zigbee2mqtt/model/IKEA.js#L31

VonOx:
I'm thinking the "click" action is deprecated not the topic itself

atrovato:
I think we are ok here. No click read for topic, only `action` can be found.

@mikitouf do you have any trouble with your device?

mikitouf:
yes , when I click , I see it on gladys8080 but it's do noting in my scene [https://ibb.co/fNtfd7L](url)
When I started it manually it's work but not with the device

VonOx:
So maybe a scene problem, not device.

Please post your intent in Gladys forum. It's not clear to me here.

mikitouf:
ok. i do it [https://community.gladysassistant.com/t/probleme-dans-lutilisation-de-linterrupteur-ikea-e1743-dans-une-scene-en-tant-que-declencheur/6525](https://community.gladysassistant.com/t/probleme-dans-lutilisation-de-linterrupteur-ikea-e1743-dans-une-scene-en-tant-que-declencheur/6525)

PR #1288, opened 2021-09-10
Change a deprecated function to the device E1743 ( IKEA )

VonOx:
#1257


VonOx:
@cicoub13 can you check this draft ? I'm not sure this is the right way to fix the issue.

PR #1294, opened 2021-09-11
Zigbee2mqtt devices

atrovato:
Zigbee2Mqtt devices from genetor tool.


VonOx:
Wow! You have scrapped z2m doc or something? 

Edit : I just see your post on forum

Pierre-Gilles:
Nice job, but I don't see the point of generating all files 😄

If we are able to generate all files, can't we do the mapping on the fly like on the Philips hue integration ? Make it dynamic ? 

We made it static because we weren't able to do it automatically, but if we are we can simplify the backend side and remove all those files :)

atrovato:
I close it since #1302 is ready for testing.

PR #1297, opened 2021-09-13
Add feature request about camera image rotation 180°

VonOx:
Add feature request about camera image rotation 180°

https://community.gladysassistant.com/t/pouvoir-tourner-une-image-de-camera-a-180/6508?u=vonox

Preview
image


VonOx:
@atrovato @Pierre-Gilles @cicoub13 , if any of you had some times to help me with front.

I'm struggle with switch in ui , I think it's working now because in DB value is updated.

image

But every time I save updated camera, Gladys loop in infinite loading.
image

Force refresh and problem is gone.

Where I'm wrong ?

Pierre-Gilles:
@VonOx Do you still need help with the UI ?

VonOx:
Yes please

atrovato:
For your information, demo mode doesn't work here. Parameter is missing in demo.json file.

atrovato:
And I think here
```
updateCameraRotation = e => {
    let newValue;
    if (e.target.checked) {
      newValue = 1;
      console.log('Je suis actif');
    } else {
      newValue = 0;
      console.log('Je suis inactif');
    }
    this.props.updateCameraRotation(this.props.cameraIndex, newValue);
  };
```
should set string var, not int.
```
updateCameraRotation = e => {
    const newValue = e.target.checked ? '1' : '0';
    this.props.updateCameraRotation(this.props.cameraIndex, newValue);
  };
```

VonOx:
Thank you @atrovato

But error is still here while saving ( toggle and save works anyway because value in DB is updated)

image

Corresponding line :

```js
checked={props.camera.cameraRotation.value === '1'}
```

atrovato:
There is a bug in `createOrUpdateCamera` and `saveCamera` methods (actions.js).

Device parameters are not mapped as it is in `getRtspCameraDevices` method.
I think we have the problem on "CAMERA_URL" too.

You should reuse this part (please introduce a new method instead of simply copy/paste).
```javascript
const cameraUrlParam = camera.params.find(param => param.name === 'CAMERA_URL');
if (cameraUrlParam) {
  camera.cameraUrl = cameraUrlParam;
}
const cameraRotationParam = camera.params.find(param => param.name === 'CAMERA_ROTATION');
if (cameraRotationParam) {
  camera.cameraRotation = cameraRotationParam;
}
```

VonOx:
> I think we have the problem on "CAMERA_URL" too.

Without my code , camera save works. It's on me but 🔽 

> Device parameters are not mapped as it is in `getRtspCameraDevices` method.

I don't understand that, I have duplicated the code stuff about `CAMERA_URL` . Is this index related ?

-------------------

image
cameraRotation => object name is the same as `getRtspCameraDevices`

image
And this part is for updating value in DB

The difference in `getRtspCameraDevices` is we are getting params by device not by param index ?
I'm right ?

image

atrovato:
Once we call save WS, we override the camera object, which has no more direct cameraRotation attribute.
But we set it right on load.
The property mapper should be called after we call save WS, before we update state.

That's nice you try yourself, and more, that's nice you ask for help, I appreciate.

VonOx:
Oh my god I think I got it

saveCamera return the state on newCamera object without param.
```js
async saveCamera(state, index) {
      const camera = state.rtspCameras[index];
      camera.features[0].name = camera.name;
      const newCamera = await state.httpClient.post(`/api/v1/device`, camera);
      const cameraUrlParam = newCamera.params.find(param => param.name === 'CAMERA_URL');
      if (cameraUrlParam) {
        newCamera.cameraUrl = cameraUrlParam;
      }
      const cameraRotationParam = newCamera.params.find(param => param.name === 'CAMERA_ROTATION');
      if (cameraRotationParam) {
        newCamera.cameraRotation = cameraRotationParam;
      }
      const rtspCameras = update(state.rtspCameras, {
        [index]: {
          $set: newCamera
        }
      });
      store.setState({
        rtspCameras
      });
    }
```
This works now, thank you @atrovato

front/src/routes/integration/all/rtsp-camera/actions.js
```diff
@@ -12,6 +12,17 @@ import createActionsIntegration from '../../../../actions/integration';
 function createActions(store) {
   const integrationActions = createActionsIntegration(store);
   const actions = {
+    async getRstpCameraParams(camera) {
```
atrovato: Not sure this should be called `getRstpCameraParams` but maybe `complete` or `map`...

server/services/rtsp-camera/lib/getImage.js
```diff
@@ -30,14 +37,18 @@ async function getImage(device) {
     );
     // we create a writestream
     const writeStream = fse.createWriteStream(filePath);
+    const outputOptions = [
+      '-vframes 1',
+      '-vf scale=640:-1', // resize the image with max width = 640
+      '-qscale:v 15', //  Effective range for JPEG is 2-31 with 31 being the worst quality.
+    ];
+    if (cameraRotationParam.value === '1') {
+      outputOptions.push('-vf hflip,vflip'); // Rotate 180
```
atrovato, line 46: Don't we plan to get 90° or only horizontal nor vertical rotation?

Pierre-Gilles: I'm not sure we can even do 90° no ? 

We display only landscape images

VonOx: Not sure about that because most of the camera have a good orientation ( spherical support ) and 90° rotation mean we are loosing aspect ratio. ( weird  things will happen in front )

front/src/routes/integration/all/rtsp-camera/actions.js
VonOx: I'm ok with "complete"

atrovato (dismissed):
Good enough to me :)

Pierre-Gilles (changes requested):
When I start this PR on an existing instance with existing cameras, the UI is broken: 

Screenshot 2021-10-18 at 13 20 02

This DEVICE_PARAM should be optional!

VonOx:
OK, what's the good way to solve this?

atrovato:
@VonOx you can use `get(props, 'camera.cameraRotation.value') === '1'` instead of `props.camera.cameraRotation.value === '1'`
Using `get` (alread imported) avoid 'undefined' properties.

atrovato (dismissed):
Looks fine with `get` usage.

VonOx:
Not yet, same error throw when saving ( because param not exist and it is not created), trying to figure out on my own 😉

VonOx:
@atrovato Need some help. I'm trying to create param if not exist in updateCameraRotation function ( mimic addCamera ):

```js
    updateCameraRotation(state, index, value) {
      let cameraRotationParamIndex = state.rtspCameras[index].params.findIndex(
        param => param.name === 'CAMERA_ROTATION'
      );
      if (!cameraRotationParamIndex === null) {
        const rtspCameras = update(state.rtspCameras, {
          [index]: {
            cameraRotation: {
              value: {
                $set: value
              }
            },
            params: {
              [cameraRotationParamIndex]: {
                value: {
                  $set: value
                }
              }
            }
          }
        });
        store.setState({
          rtspCameras
        });
      } else {
        const rtspCameras = update(state.rtspCameras[index], {
          $set: [
            {
              cameraRotation: {
                name: 'CAMERA_ROTATION',
                value: '0'
              },
              params: [
                {
                  name: 'CAMERA_ROTATION',
                  value: '0'
                }
              ]
            }
          ]
        });
        store.setState({
          rtspCameras
        });
      }
    },
```
No api available in front to add param on an existing device ? ( like state.rtspCameras[index].params.add(NAME,VALUE) )
I've tried $push / $set without success.

atrovato:
Why to you to create param?
Default behavior/value won't do the job?

```javascript
    async complete(camera) {
      const cameraUrlParam = camera.params.find(param => param.name === 'CAMERA_URL');
      /*
      if (cameraUrlParam) {
        camera.cameraUrl = cameraUrlParam;
      }
      */
      camera.cameraUrl = cameraUrlParam || ''; // Not sur for this one

      const cameraRotationParam = camera.params.find(param => param.name === 'CAMERA_ROTATION');
      /* Current state
      if (cameraRotationParam) {
        camera.cameraRotation = cameraRotationParam;
      }
      */
      // New state
      camera.cameraRotation = cameraRotationParam || '0';
      return camera;
    },
```

VonOx:
Nope I've error in console when toggle the switch

image

Logic error ( I think ) because cameraRotationParamIndex is null ( param doe'snt exist in DB )

```js
    updateCameraRotation(state, index, value) {
      let cameraRotationParamIndex = state.rtspCameras[index].params.findIndex(
        param => param.name === 'CAMERA_ROTATION'
      );
      const rtspCameras = update(state.rtspCameras, {
        [index]: {
          cameraRotation: {
            value: {
              $set: value
            }
          },
          params: {
            [cameraRotationParamIndex]: {
              value: {
                $set: value
              }
            }
          }
        }
      });
```

That's why I'm trying to create the param in the function updateCameraRotation() ( previous post )

atrovato:
Ok, that's only for backward compatibility. Param is missing on already created camera.
So, I think, you should set it up right on load (else part).

```javascript
    async complete(camera) {
      const cameraUrlParam = camera.params.find(param => param.name === 'CAMERA_URL');
      if (cameraUrlParam) {
        camera.cameraUrl = cameraUrlParam;
      }
      const cameraRotationParam = camera.params.find(param => param.name === 'CAMERA_ROTATION');
      if (cameraRotationParam) {
        camera.cameraRotation = cameraRotationParam;
      } else {
        // IF PARAM NOT FOUND, ADD IT
        const rotationParam = { name: 'CAMERA_ROTATION', value: '0' };
        camera.cameraRotation = rotationParam;
        camera.params.push(rotationParam);
      }
      return camera;
    }
```

VonOx:
Thank you @atrovato , I've understood

atrovato (approved):
Look great again :)

atrovato:
This one is ready to merge.

Pierre-Gilles (approved):
Just tested locally and it works fine, even with legacy devices ! Great work 👏

Sorry for taking so long to review this

PR #1334, opened 2021-10-21
Slugify with limax lib

### LINKED ISSUE #1110: Create house error when house name is in chinese
image

image
image

i found that if i create a house from /api/v1/house success,and then create room error.
the i save house again,the house name has created. i can't create

somethings else,is there support chinese room name or house name?

VonOx:
Fix #1110


atrovato:
@VonOx it looks like this new dependency breaks something.

Pierre-Gilles:
Yes, because the slugify function is used in the front as well.

Before, as the function was standalone, it was ok to import it in the front because there was no dependency needed.

Now, if we import this function in the front as well, the dependency needs to be in the front too (Do we want to ?)

Let's discuss this on the issue side.

PR #1335, opened 2021-10-21
Zigbee2mqtt update device

atrovato:
### Pull Request check-list

### Description of change

Detect if discovered device is already in Gladys, or ask for update.

### DIFF
--- a/server/utils/device.js
+++ b/server/utils/device.js
@@ -77,39 +77,40 @@ const matchParam = (params, param) => {
 };
 
 /**
- * @description Compare newDevice with exisingDevice on features and parameters
- * to tell if device can be updated.
+ * @description Compares both argument devices and check if any changes occurred on following attributes:
+ *  - features: check for added or deleted features (based only on external_id)
+ *  - params: check for added, updated or deleted (based on name and value)
  *
  * @param {Object} newDevice - New device.
  * @param {Object} existingDevice - Existing device.
- * @returns {boolean} Is new device is different on feature or parameters?
+ * @returns {boolean} Indicates if the new device is different from the existing one.
  * @example
- * isUpdatable({}, {})
+ * hasDeviceChanged({ features: [(3)], params: [ ... ]}, { features: [(0)], params: [ ... ]})
  */
-function isUpdatable(newDevice, existingDevice = {}) {
+function hasDeviceChanged(newDevice, existingDevice = {}) {
   const { features, params } = newDevice;
   const existingFeatures = existingDevice.features || [];
   const existingParams = existingDevice.params || [];
 
-  let updatable = existingFeatures.length !== features.length || existingParams.length !== params.length;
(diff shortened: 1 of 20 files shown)


atrovato:
This allow to see "update" button when device is changed.
image

And fix many client errors (seen in browser console).

server/utils/device.js
```diff
@@ -68,8 +68,48 @@ function getDeviceFeature(device, category, type) {
...
+ * @description Compare newDevice with exisingDevice on features and parameters
+ * to tell if device can be updated.
+ *
+ * @param {Object} newDevice - New device.
+ * @param {Object} existingDevice - Existing device.
+ * @returns {boolean} Is new device is different on feature or parameters?
+ * @example
+ * isUpdatable({}, {})
+ */
+function isUpdatable(newDevice, existingDevice = {}) {
```
Pierre-Gilles: what would make a device updatable or not ? Maybe the description is not self-explanatory, I don't understand why some device could be updated and some couldn't :)

Pierre-Gilles (commented):
Thanks for the PR! 

I wrote a small comment :) 

Do you recommend that I wait for this to be merged before deploying the new Zigbee2mqtt discovery ? 

Has this been tested by Zigbee2mqtt users ?

server/utils/device.js
atrovato: You're right, the name is not explicit enough,
this method can be used to "compare" device already stored in Gladys with newly discovored one.
It compares features on external_id and params on name and value.
If any is different, it returns true, so this can be used to display "update" button in discovery tab.

I'm open to any remarks on this point.

atrovato:
I'm not totally confortable about your comment, I agree your remark, but I feel empty of inspiration.

This has not been tested by community yet, I only checked this with demo mode.
This is not required with last Zigbbe2mqtt changes, it is an evolution.

I have problems on tests, not related to this:
```
  1) Device.calculateAggregate
       should calculate daily aggregate:

      AssertionError: expected 2900 to equal 3000
      + expected - actual

      -2900
      +3000
      
      at Context.<anonymous> (test/lib/device/device.calculateAggregate.test.js:101:43)

  2) Device.getDeviceFeaturesAggregates
       should return last day states:

      AssertionError: expected [ Array(71) ] to have a length of 72 but got 71
      + expected - actual

      -71
      +72
      
      at Context.<anonymous> (test/lib/device/device.getDeviceFeaturesAggregates.test.js:96:28)
```

server/utils/device.js
Pierre-Gilles: hasDeviceChanged?

atrovato: I agree, with more detailled comment to explain "changed on what".
Thanks ;)

Pierre-Gilles (approved):
Thanks for your PR, looks good to me now!

PR #1337, opened 2021-10-25
Add internal calendar sharing

bertrandda:
### Pull Request check-list

### Description of change

Added attribute in model `Calendars` to list Gladys users who can see the calendar.
In front calendar page display my calendars & calendars shared with me

image


Pierre-Gilles:
First, thanks for the PR ! 🙏

I don't really agree with the database modelization

First question, couldn't we just add a boolean like "shared: true/false", that just share the calendar with the whole instance? 

Do we really need this level of granularity?

bertrandda:
When it's a classic home usage maybe it's not necessary, but I though about someone with a deep usage of Gladys (ex: Terdious here https://community.gladysassistant.com/t/multi-utilisateurs/5510/5). Maybe he will want to share one calendar only with his wife and an other with his children or colleagues. Granularity was easy to set up in calendar so why not ?

I agree on modelisation, the easiest way was to just add an attribute that concatenate ids to not create specific joining table but I can create one if you prefer.

Pierre-Gilles:
I'm not sure we want to go for the most complex solution to solve the issue for all users, I'm sure most people just need a shared: true/false. 

In the case someone needs a very specific sharing, maybe he can just handle the sharing part in his own calendar (send invites), and then let each user connect the same calendar to Gladys (and in that case, we need to handle duplicate external_id)

I see two solutions: 

- Either we stick to the user calendar sharing in Gladys, but in that case I'm more in favor of a `t_calendar_user` table (the relational way, with enforced foreign keys)
- Either we simplify and move to a boolean option (shared: true/false)

What do you think ?

Pierre-Gilles:
I also see a potential problem: If you create a scene that works with a calendar (no the case right now, but it'll happen soon), if the user doesn't have access to the calendar, he won't be able to see the scene in the UI (and scene are cross-users)

If we create a notion of "public" calendars, we can potentially say that only public calendars can be used in scene

bertrandda:
And we do not plan to have "private" scenes?

If no, indeed public true/false calendar could be the best way

Pierre-Gilles:
No I don't think so.

Either the user is "admin" and he can see all scenes in the instance, or he is "user" and he is not authorized to create scenes.

bertrandda:
Ok then boolean should be enough, I change that

front/src/config/i18n/en.json
```diff
@@ -737,9 +737,11 @@
       "title": "CalDAV",
       "accountTab": "Account",
       "syncTab": "Synchronization",
+      "shareTab": "Sharing",
       "description": "Sync your CalDAV calendars.",
       "accountIntroduction": "To sync Gladys calendar using CalDAV, you first need an external compatible calendar. Then enter the configuration below.",
       "syncIntroduction": "Choose the calendars to synchronize.",
+      "shareIntroduction": "Choose the calendars you want to share with other Gladys users.",
```
atrovato: `Choose calendars` - no `the`

front/src/actions/calendar.js
```diff
@@ -9,9 +9,12 @@ function createActions(store) {
       try {
         const events = await state.httpClient.get('/api/v1/calendar/event', {
           from,
-          to
+          to,
+          shared: true
+        });
+        const calendars = await state.httpClient.get('/api/v1/calendar', {
+          shared: true
```
atrovato, line 16: Are you sure we need to explicitly request for shared calendars?
Shouldn't it the default behavior? "get mines AND shared"?

atrovato (commented):
Great code, some comments.
I don't try it for now, just review code.
I will ASAP.

front/src/actions/calendar.js
bertrandda: In my opinion, default should be yours only. For example in configuration we fetch only your calendars to select those to sync and It.
Discussion is open, I can change it

Pierre-Gilles: I agree with @bertrandda, what I see: 

- By default on the configuration page, you should only be able to change your calendars (not the shared one)
- On the calendar page, all calendars (yours + shared) are displayed
- In scenes, only shared calendars are displayed (not in this PR)

bertrandda:
Not stale, ready for reviews

Pierre-Gilles (approved):
I just tested the PR end-to-end locally, and it works great! 👏

Amazing work, sorry for taking so long to review !

PR #1353, opened 2021-11-10
Fix random tests

atrovato:
Sometimes, device aggration tests fail due to current date.
This fix current date to same date (taken from merge date).

PR #1354, opened 2021-11-10
Fix python dependency

atrovato:
Use `python3` to build docker image.


atrovato:
This one is blocked by https://github.com/GladysAssistant/Gladys/pull/1353

PR #1355, opened 2021-11-10
Fix demo mode

atrovato:
- Fix some bad paths on demo mode
 - Delete confusing `demo.json` file
 - Add timer on DemoHttpClient to see loaders.


Pierre-Gilles:
Oh shit I forgot to delete the demo.json file, my bad! 

Nice catch, I'm good for this!

front/src/utils/DemoHttpClient.js
```diff
@@ -1,5 +1,8 @@
 export class DemoHttpClient {
   getDemoFile = async () => {
+    // Set a timer to see loaders
+    await new Promise(r => setTimeout(r, 200));
```
Pierre-Gilles: I'm good with that (feels more realistic), but 200ms seems too slow on my side, it looks like the demo page is slower than the real Gladys :D 

Could you change to 100ms ?

Pierre-Gilles (changes requested):

front/src/utils/DemoHttpClient.js
atrovato: I will add this as an "env" config, maybe disable it by default for public mode, but with env var, I can enable it on my side.
What do you think?

Pierre-Gilles: I'm good with that 👍

atrovato: Done, disabled by default, can be overridden with 
```
DEMO_MODE=true
DEMO_REQUEST_TIME=1000
```

Pierre-Gilles: Works on my side!

Pierre-Gilles (approved):

PR #1361, opened 2021-11-12
Make installation of services mandatory to install Gladys

atrovato:
Make installation of services mandatory to install Gladys

PR #1362, opened 2021-11-12
Silent fail option on service install

atrovato:
Allow to silent fail service install with env var

.env
```
INSTALL_SERVICES_SILENT_FAIL=true
```

Default: build fails.


Pierre-Gilles:
Looks good to me!

atrovato:
Included in https://github.com/GladysAssistant/Gladys/pull/1361

PR #1364, opened 2021-11-12
Device room selector component

atrovato:
### Pull Request check-list

### Description of change

Next common component: device room selector.

### DIFF
--- a/front/src/components/device/form/DeviceRoomSelector.jsx
+++ b/front/src/components/device/form/DeviceRoomSelector.jsx
@@ -6,13 +6,20 @@ import cx from 'classnames';
 import actions from '../../../actions/house';
 import { RequestStatus } from '../../../utils/consts';
 
-@connect('houses,housesGetStatus', actions)
+@connect('housesWithRooms,housesGetStatus', actions)
 class DeviceRoomSelector extends Component {
   componentWillMount() {
     this.props.getHouses();
   }
 
-  render({ houses = [], housesGetStatus = RequestStatus.Getting, selectedRoomId, updateRoom, disabled, inputId }) {
+  render({
+    housesWithRooms = [],
+    housesGetStatus = RequestStatus.Getting,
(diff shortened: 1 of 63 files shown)


Pierre-Gilles (approved):

front/src/components/device/form/DeviceRoomSelector.jsx
```diff
@@ -0,0 +1,52 @@
+import { Component } from 'preact';
+import { Text } from 'preact-i18n';
+import { connect } from 'unistore/preact';
+import cx from 'classnames';
+
+import actions from '../../../actions/house';
+import { RequestStatus } from '../../../utils/consts';
+
+@connect('houses,housesGetStatus', actions)
```
Pierre-Gilles: I think there is an issue with the variable name `houses`

As the state is global, it's shared with all other componenents

With have some components that use the same key (`houses`) to store a list of houses without the rooms

It caused some issues in the past, that's why I renamed this key `housesWithRooms` when the room was included

Two ways of fixing this: 

- Rename the key to `housesWithRooms` here
- Stop using global variables (but we are losing the cache)

Pierre-Gilles (changes requested):

front/src/components/device/form/DeviceRoomSelector.jsx
Pierre-Gilles: Example here : https://github.com/GladysAssistant/Gladys/blob/master/front/src/components/boxs/weather/EditWeatherBox.jsx#L18

atrovato: Nice, thank for this point.
I will rename to `housesWithRooms` and (try to) check all other usages.

atrovato:
I will take it back and split into multiple smaller PRs, to migration integration pages one by one.
Keep tracking on https://github.com/atrovato/Gladys/pull/40

PR #1366, opened 2021-11-13
Check missing unit / feature translations

atrovato:
Add a build tool to check translations for units and features.
If a translation is missing, build will fail with:
image

### DIFF
--- a/front/cli/check_translations.js
+++ b/front/cli/check_translations.js
@@ -1,0 +1,49 @@
+const {
+  DEVICE_FEATURE_UNITS,
+  DEVICE_FEATURE_CATEGORIES,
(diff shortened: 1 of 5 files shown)


VonOx (dismissed):
Tested, works fine  in local ! But CI fail

atrovato:
Next step can be on feature icons, to check missings, but a lot are missing for now.

atrovato:
That's quite weird, a logger dependency is missing...

EDIT: this is because I use server logger, not front logger...

VonOx (approved):

Pierre-Gilles (approved):
Nice good idea! Looks good to me :)

PR #1383, opened 2021-11-27
Fix zigbee2mqtt set value

atrovato:
Use the device/feature property name to set value (instead of default "state" property).

Linked to: https://github.com/GladysAssistant/Gladys/issues/1333


Pierre-Gilles:
I'm waiting from more feedbacks from someone who can test this with lots of Zigbee2mqtt devices :)

atrovato:
@Pierre-Gilles before merging this one, what do you think about the storage of feature mapping values in device parameters.
For instance, here we use the property `feature_map_alarm = { 'ON': 1, 'OFF': 0, 1: 'ON', 0: 'OFF'}` to get mapping from Gladys to z2m (and vice-versa).

atrovato:
It may be more appropriated to store feature mapping values in feature directly, not in device.
And if it's well modelised, we may use this in front too, to display "text" according to int value.

```json
{
  "name": "feature",
  "min": 0,
  "max": 1,
  "mapping": {
    1: "ON",
    0: "OFF"
  }
```

atrovato:
Or maybe an array with index mapping
```json
{
  "name": "feature",
  "min": 0,
  "max": 1,
  "mapping": ["OFF", "ON"]
}
```

@Pierre-Gilles what do you think?

jparbel:
@Pierre-Gilles If possible, could you explain me what's the best way to test new commits. Thanks a lot.

Pierre-Gilles:
> It may be more appropriated to store feature mapping values in feature directly, not in device.

Why not, like a "value_mapping" attribute in the device_feature table? Could work.

How is it modelized in the data you can from Zigbee2mqtt ? Can you show me an example of a JSON you receive (full JSON of a device) ? 

> And if it's well modelised, we may use this in front too, to display "text" according to int value.

On this side I'm not sure we want that. Those data are "raw" english text and not translated. It should be in the i18n file, so it's translated in every language.

> Or maybe an array with index mapping

I prefer key value :D 

@jparbel Can you create a post on the forum so we can talk about it there? It depends on what you want to test, how you want to test it, it's not a simple answer :)

atrovato:
@Pierre-Gilles you will see an array of z2m devices here https://github.com/GladysAssistant/Gladys/blob/master/server/test/services/zigbee2mqtt/lib/payloads/mqtt_devices_get.json

You will see values for enum type, but for binary, I found value mappings on https://www.zigbee2mqtt.io/guide/usage/exposes.html#generic

Pierre-Gilles:
@atrovato So in that case "key" => "value" is best imo. 

I'm still wondering if it's a good idea to create a new column in the device_feature table, I'm wondering if this behavior is that generic. 

In most case here, I would have advocated to keep this mapping in the code. But here, for this precise case, it should be dynamic to keep the auto-discovery mechanism working.

But this is the only system I know that does that, so i'm really not sure about factoring that right now..

atrovato:
@Pierre-Gilles In the incoming Broadlink service, I will need to store an infrared capture according to each value (power on/off, buttons 0 to 9, volume up/down....). For now, I store all mappings in device params, but I need to be very strict on the naming.
In fact, this option is useful when the value is device/model specific.

jparbel:
>Can you create a post on the forum so we can talk about it there? It depends on what you want to test, how you want to test it, it's not a simple answer :)

@Pierre-Gilles I found the good answer on the forum (post from Psoy 2021 September <<_Lancer une image Gladys de test_>>)
It was exactly my question and you gave the good way to test. (I did not see it before)
But where can I find on Github the name of a specific image, for example, from Alex ?
Thank's a lot

atrovato:
@jparbel I didn't generate any docker image for this (yet).
We have an open discussion with @Pierre-Gilles.

And I think that I shouldn't store mapping in feature for z2m devices (only for captured devices).
For z2m I can map values with discovered payloads, as I already keep these in memory for the discovered tab in Gladys UI.

Let me some days to change this.
I draft this PR until it's done.

jparbel:
We have an open discussion with @Pierre-Gilles.

jparbel:
We have an open discussion with @Pierre-Gilles.

@atrovato excuse me for my posts, I did'nt know it was not the good place to send messages.
I understood it's a draft PR, but I just wanted to know how to test a PR.
I will ask the same question on the forum.

atrovato:
@jparbel don't worry, you're welcome in this discussion ;)
I changed my approach according the value mapping, and a new docker image is building (count 1h30 to get built).

You will found it with `atrovato/gladys:zigbee2mqtt` image name.

jparbel:
@atrovato Thank's for your work, I will try your image today.

> You will found it with `atrovato/gladys:zigbee2mqtt` image name.

I didn't found your image, Github is not very familiar for me.
I went on your Github : https://github.com/atrovato/Gladys. I don't know if I have to download the master branch or to go to actions/Build gladys dev images and after ...?

Pierre-Gilles:
@atrovato So we keep that in device param? Good for me :)

For broadlink, are you sure it's not something that should simply be in the code?

atrovato:
For Broadlnk, the concept is:
1. enable capture mode on IR device
2. press remote control button to copy IR code
3.  keep this code somewhere

As this captured code is related to a device, and more, to a feature, when need to keep device+feature information in database.
Each remote control provides differents IR codes. Today, on my git feature branch, I choose to store in device params, but device may have many many captured codes (power on/off, button from 0 to 9, volume ou/down, record, play, pause, stop...). So today I stored all of these as device params, and use an internal feature param naming to map device param to device.feature value.

I think we should talk about in appropriate topic.

---

Now for z2m, I reuse discovered device to map values, this ensure to be up-to-date according to z2m version.

Pierre-Gilles:
@atrovato I answered about Broadlink on the forum https://community.gladysassistant.com/t/v4-reflexion-integration-telecommande/5002/56?u=pierre-gilles

Is this PR ready for review?

Pierre-Gilles:
Btw, still waiting for testers on this PR ( @LM1LC3N7 ? )

atrovato:
I just have some feedbacks, on many devices, I forgot recursive search on `findMatchingExpose` method.
I am building a new image, for new session of tests.

atrovato:
@Pierre-Gilles tests look good according to https://community.gladysassistant.com/t/zigbee2mqtt-debug/6080/163 topic.
Ready to review ;)

Pierre-Gilles:
@atrovato If I understand correctly you're no longer storing the "expose" in database, only in RAM ? Looks better to me like this !

atrovato:
That's right, only RAM discovered devices. So we stay up to date with z2m version.

LM1LC3N7:
Sorry I have so much github notifications I did not noticed your mention 😅

Pierre-Gilles (approved):

Pierre-Gilles (approved):

PR #1385, opened 2021-11-28
Fix vibration sensor

atrovato:
Zigbee2mqtt vibration sensor is a binary, not a numeric sensor.


atrovato:
Change validated: 
https://community.gladysassistant.com/t/capteur-de-vibration-tuya-zigbee/6632/20?u=alextrovato

Pierre-Gilles (approved):

PR #1387, opened 2021-11-28
Remove useless TriggerPage

atrovato:
While I was refactor some common component, I see the TriggerPage.
But it looks it is a fake page, never used.
So, this PR removes it.


Pierre-Gilles (approved):
Indeed, I think it's some old parts, at first I wanted to separate the trigger page from the scene page, and at the end it's just the same view, and I forgot to clean this :) Thanks!

PR #1388, opened 2021-11-28
Nextcloud Talk service

bertrandda:
### Pull Request check-list

### Description of change

Added [Nextcloud Talk](https://nextcloud.com/talk/) integration to talk with Gladys throught Nextcloud Talk app.


front/src/routes/integration/all/nextcloud-talk/actions.js
```diff
@@ -0,0 +1,118 @@
...
+    store.setState({
+      nextcloudTalkGetSettingsStatus: NextcloudTalkStatus.Getting
+    });
+
+    let nextcloudUrl = '';
+    let nextcloudBotUsername = '';
+    let nextcloudBotPassword = '';
+    let nextcloudBotToken = '';
+
+    store.setState({
```
atrovato, line 30: Please call `store.setState` once, merging `nextcloudTalkGetSettingsStatus` and `nextcloud` vars

front/src/utils/consts.js
```diff
@@ -94,6 +94,12 @@ export const GetWeatherModes = {
   DailyForecast: 'dailyForecast'
 };
 
+export const NextcloudTalkStatus = {
```
atrovato: Not sure we want `NextcloudTalkStatus` in global `consts.js` file.

server/lib/user/index.js
```diff
@@ -8,6 +8,8 @@ const { getBySelector } = require('./user.getBySelector');
 const { getPicture } = require('./user.getPicture');
 const { getUserCount } = require('./user.getUserCount');
 const { getByTelegramUserId } = require('./user.getByTelegramUserId');
+const { getByNextcloudTalkToken } = require('./user.getByNextcloudTalkToken');
+const { getNextcloudTalkTokens } = require('./user.getNextcloudTalkTokens');
```
atrovato: Not we want Nextcloud as a part of Gladys core.
It should be a service, not changing Gladys core.

server/services/nextcloud-talk/lib/message.bot.js
```diff
@@ -0,0 +1,76 @@
...
+class NextcloudTalkBot extends EventEmitter {
+  constructor(gladys, serviceId, token) {
+    super();
+    this.gladys = gladys;
+    this.serviceId = serviceId;
+    this.token = token;
+    this.isPolling = false;
+  }
+
+  async startPolling() {
```
atrovato: We usually extract class methods in single files.

atrovato (changes requested):
I stop reading when I see you add a "nextcloud" token in Gladys core user model / table.
I am quite sure this is not the way to do this.
You may use service variables... or let's talk about point together.

VonOx:
@atrovato this is the way telegram is implemented, no?

atrovato:
@VonOx @Pierre-Gilles 
I don't think it's a good way to do.
I don't think we want to add information in user table each time we add a communication service.

bertrandda:
As @VonOx says, it's work like that with Telegram. I tried to reproduce something structurally close for Nextcloud Talk to maintain code more easily for both services.
In Telegram service there is a `telegram_user_id` in user model, it's used here https://github.com/GladysAssistant/Gladys/blob/dd2b5f0bd37ca6f7ce6ca2fa51c7df032462d8b1/server/lib/message/message.sendToUser.js#L35-L38 to know where send message for a specific user

server/services/nextcloud-talk/lib/message.bot.js
bertrandda: I think it's a bit particular, we will need to handle multiple bots in the same time, one for each user and they can have different status. If you have a better way to do it we can still change it

front/src/utils/consts.js
bertrandda: I am removing `NextcloudTalkStatus` and changing to existing `RequestStatus`

Pierre-Gilles:
Yeah about the Telegram implementation, I'm not sure we want to do this for all new services, I agree this was my mistake :) 

What about just using the variable with a user_id parameter? 

You can create variable in the t_variable table, it's scoped by service & by user so you can have a special key to store this value, one by user, and no need to change the database schema for this (it's key/value)

bertrandda:
Thank you @atrovato & @Pierre-Gilles. So I change the way to store the token. It now use variable (same as for username, password & url).
In Gladys core, I juste added `variable.getVariables.js` to fetch all existing variables for a same name and start correctly all user's Nextcloud bots. I also added variable lib inside `MessageHandler` to be able to retrieve token and send message through Nextcloud Talk
Is it better for you ?

bertrandda:
Not stale, ready for reviews

server/services/nextcloud-talk/lib/message/message.bot.js
```diff
@@ -0,0 +1,77 @@
+const EventEmitter = require('events');
+const get = require('get-value');
+const logger = require('../../../../utils/logger');
+
+class NextcloudTalkBot extends EventEmitter {
```
Pierre-Gilles: We don't really use this pattern in Gladys (class with inheritance in one big file)

To stay consistent with the rest of the project, could you do it with a prototype and 1 function per file like on other services?

server/services/nextcloud-talk/lib/message/message.disconnect.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @description Disconnect Nextcloud Talk
+ * @returns {Promise} - Resolve.
+ * @example
+ * disconnect();
+ */
+async function disconnect() {
+  if (this.bots.length > 0) {
+    Promise.all(this.bots.map((bot) => bot.stopPolling()));
```
Pierre-Gilles: Isn't that too aggressive? Use Promise.each

server/services/nextcloud-talk/lib/message/message.bot.js
```diff
@@ -0,0 +1,77 @@
...
+    if (
+      lastKnownMessageId &&
+      newMessages &&
+      newMessages[newMessages.length - 1].actorId !== this.NEXTCLOUD_BOT_USERNAME
+    ) {
+      this.emit('message', newMessages.pop());
+    }
+
+    if (this.isPolling) {
+      setTimeout(() => this.poll(result.headers['x-chat-last-given'] || lastKnownMessageId), 50);
```
Pierre-Gilles: Are you polling every 50ms forever? It's a bit extreme no?

server/lib/message/message.sendToUser.js
```diff
@@ -30,11 +30,26 @@ async function sendToUser(userSelector, text, file = null) {
     payload: messageCreated,
   });
   // We send the message to the telegram service
-  const service = this.service.getService('telegram');
+  const telgramService = this.service.getService('telegram');
   // if the service exist and the user had telegram configured
-  if (service && user.telegram_user_id) {
+  if (telgramService && user.telegram_user_id) {
```
Pierre-Gilles: Typo

Pierre-Gilles (changes requested):
Thanks for this PR ! A few feedbacks :)

server/services/nextcloud-talk/lib/message/message.bot.js
bertrandda: I rewrite full process, it should be better

server/services/nextcloud-talk/lib/message/message.bot.js
bertrandda: It's not really that, it requests for new message here https://github.com/GladysAssistant/Gladys/pull/1388/files#diff-388fb960051d373b1c5591fe8a071c088ff028c5d4628956265ebbbadc6b6b3bR16-R20
There is a timeout define in the request query parameters to 15, the request end on new message response OR on timeout, then retry if we polling is enable, I added a delay of 50ms to possibly reduce load but it's not really necessary.

Pierre-Gilles: But ` this.bots[userId].isPolling = true` right ?

So it's always doing a request, wait 50ms, one other request?

Considering the request takes 100ms, it means we are requesting every 150ms forever

It means 6.6 req/seconds = 2160 req/min = 129k requests per hour = 3 millions requests per day = 90 millions req/month

It's wayyy to much 😅

Pierre-Gilles: Ok makes sense, I went reading the code and indeed!

front/src/config/i18n/en.json
```diff
@@ -319,6 +319,22 @@
...
+      "introduction": "To speak through Nextcloud Talk, you need a specific Nextcloud account for your bot.",
+      "url": "Nextcloud host",
+      "urlInfo": "Enter your Nextcloud base URL.",
+      "username": "Username Nextcloud bot account",
+      "usernameInfo": "Enter Nextcloud bot account username.",
+      "password": "Password Nextcloud bot account",
+      "passwordInfo": "Enter Nextcloud bot account password.",
+      "token": "Bot conversation token",
+      "tokenInfo": "Enter conversation token between your Nextcloud account and the bot account.",
+      "configurationError": "Failed to save configration.",
```
Pierre-Gilles: *configuration

Pierre-Gilles (changes requested):
Thank you for making the necessary changes, I appreciate it the PR is really nice now 🙏

Code is clean, and well designed. Great work 👏

I spotted a super small typo, but nothing really serious.

If the PR is properly working for users, I'm good to merge! 👍

server/services/nextcloud-talk/index.js
```diff
@@ -0,0 +1,41 @@
...
+   * @public
+   * @description This function stops the Nextcloud Talk Service
+   * @example
+   * gladys.services.nextcloudTalk.stop();
+   */
+  async function stop() {
+    logger.info('stopping Nextcloud Talk service');
+    await messageHandler.disconnect();
+  }
+
```
Pierre-Gilles, line 35: Oh, and just one other feedback, can you implement the new function `isUsed` ?

It should return true/false if the user is actively using the service :) 

Like: 

https://github.com/GladysAssistant/Gladys/blob/master/server/services/zigbee2mqtt/index.js#L38

server/services/nextcloud-talk/lib/message/message.connect.js
```diff
@@ -0,0 +1,35 @@
...
+ * @example
+ * connect([
+ *  { value: 'token1', user_id: 'f0de00a8-8ba7-4a4e-8f5a-7a21e94f36a8' },
+ *  { value: 'token2', user_id: '7f53d6ff-714f-4f08-a17b-df361a746d81' },
+ * ]);
+ */
+async function connect(tokens) {
+  // if a bot exist, we disconnect
+  if (Object.keys(this.bots).length > 0) {
+    await this.disconnect();
```
Pierre-Gilles, line 15: This use case seems not tested according to codecov?

server/test/services/nextcloud-talk/index.test.js
```diff
@@ -0,0 +1,45 @@
+const { expect } = require('chai');
+const NextcloudTalkService = require('../../../services/nextcloud-talk');
+
+describe('nextcloud-talk', () => {
+  it('should start service', async () => {
+    const gladys = {
+      variable: {
+        getVariables: () => Promise.resolve([{ user_id: 'f0de00a8-8ba7-4a4e-8f5a-7a21e94f36a8', value: 'testToken' }]),
+        getValue: () => Promise.resolve('testvalue'),
```
Pierre-Gilles: Why don't you use fake for this?

fake.resolve() 

?

It let's you test that the function was indeed called

Example:

https://github.com/GladysAssistant/Gladys/blob/ee812a2afae3aeeabce870d3e607975a7a026d63/server/test/services/mqtt/lib/installContainer.test.js#L81

Pierre-Gilles (changes requested):
Thanks for your changes, it very nice! :) Great PR 🙏 

I have some super small comments but nothing too serious.

server/services/nextcloud-talk/lib/message/message.connect.js
bertrandda: It should be covered now

Pierre-Gilles:
@bertrandda Is this ready for review ? Ping me if yes :)

bertrandda:
@Pierre-Gilles yes code is ready, just need user real life tests

Pierre-Gilles:
@bertrandda Was there some real-life tests since then?

bertrandda:
> @bertrandda Was there some real-life tests since then?

Except mine no new tests

Pierre-Gilles:
@bertrandda Ok, if we don't have news on the forum in 1 week, and you tested carefully on your side, I'm ok to merge it to master :) We can't wait forever for testers ! (Don't hesitate to ping me if I forget)

Pierre-Gilles:
@bertrandda Sorry I did a few Eslint changes that needs to be applied to this PR

I'm ok to merge after, no need to wait for more testers :)

PR #1416, opened 2022-01-19
Allow to use "long" event for sonoff-snzb-01

LM1LC3N7:
As requested here:
https://community.gladysassistant.com/t/bouton-sonoff-snzb-01/6812/8

### Description of change

Adding another case for devices with "long" events (equivalent to existing "hold" event).

### DIFF
--- a/server/services/zigbee2mqtt/utils/convertValue.js
+++ b/server/services/zigbee2mqtt/utils/convertValue.js
@@ -35,6 +35,9 @@ function convertValue(feature, value) {
         case 'hold':
           result = BUTTON_STATUS.LONG_CLICK;
           break;
+        case 'long':
+          result = BUTTON_STATUS.LONG_CLICK;
+          break;
         default:
           result = value;
       }


server/services/zigbee2mqtt/utils/convertValue.js
```diff
@@ -35,6 +35,9 @@ function convertValue(feature, value) {
         case 'hold':
           result = BUTTON_STATUS.LONG_CLICK;
           break;
+        case 'long':
+          result = BUTTON_STATUS.LONG_CLICK;
+          break;
```
bertrandda, line 40:
```suggestion
        case 'hold':
        case 'long':
          result = BUTTON_STATUS.LONG_CLICK;
          break;
```
Maybe better?

cicoub13 (changes requested):
OK for me. You need to complete unit tests in `/server/test/services/zigbee2mqtt/utils/convertValue.test.js`

Issue #1426, opened 2022-02-03
[Zigbee2Mqtt] Wrong value in MQTT message causes Service crash


cicoub13:
**Describe the bug**

When a device sends a value not compatible with Gladys Zigbee2Mqtt, the zigbee2mqtt service crash

https://community.gladysassistant.com/t/dashboard-evolutions-bugs-etc/5150/113#ajout-des-logs-rseau-1

**Logs**

```
2022-02-03T07:36:35+0100 <warn> handleMqttMessage.js:96 () Zigbee2mqtt device RadiateurSalleaManger, feature boost_heating not configured in Gladys.
2022-02-03T07:36:35+0100 <warn> handleMqttMessage.js:96 () Zigbee2mqtt device RadiateurSalleaManger, feature boost_heating_countdown not configured in Gladys.
2022-02-03T07:36:35+0100 <warn> handleMqttMessage.js:96 () Zigbee2mqtt device RadiateurSalleaManger, feature boost_heating_countdown_time_set not configured in Gladys.
2022-02-03T07:36:35+0100 <error> index.js:20 (process.<anonymous>) uncaughtException catched: uncaughtException
2022-02-03T07:36:35+0100 <error> index.js:21 (process.<anonymous>) Error: Zigbee2mqqt don't handle value "LOCK" for feature "mode".
    at convertValue (/src/server/services/zigbee2mqtt/utils/convertValue.js:55:17)
    at /src/server/services/zigbee2mqtt/lib/handleMqttMessage.js:92:24
    at Array.forEach (<anonymous>)
    at Zigbee2mqttManager.handleMqttMessage (/src/server/services/zigbee2mqtt/lib/handleMqttMessage.js:86:41)
    at MqttClient.<anonymous> (/src/server/services/zigbee2mqtt/lib/connect.js:53:12)
    at MqttClient.emit (events.js:400:28)
    at MqttClient._handlePublish (/src/server/services/zigbee2mqtt/node_modules/mqtt/lib/client.js:1277:12)
    at MqttClient._handlePacket (/src/server/services/zigbee2mqtt/node_modules/mqtt/lib/client.js:410:12)
    at work (/src/server/services/zigbee2mqtt/node_modules/mqtt/lib/client.js:321:12)
    at Writable.writable._write (/src/server/services/zigbee2mqtt/node_modules/mqtt/lib/client.js:335:5)
[...]

PR #1427, opened 2022-02-03
Fix 1426: Fix critical bug in Zigbee2mqtt integration that could slow down the Gladys instance in case a message is not converted

cicoub13:
### Pull Request check-list

### Description of change

Fix the issue. Didn't test with real device


cicoub13:
I don't know why the coverage is lower on `scene.addScene.js` 🤔

cicoub13:
Fixes #1426 (I forgot in my commit 😞)

Pierre-Gilles (approved):
Looks good to me! 

The addScene issue seems to be a mistake from codecov, I don't see why it could be linked.

Thanks for the quick PR :)

PR #1434, opened 2022-02-16
Docker containerId - cgroupv1 & v2 compatible

VonOx:
Rework on the getNetworkMode, more precisely getting the container Id:

* Getting container id from cid file if exist
* Updateing command to be cgroups v2 compliant

⚠️ If host system is using cgroupv2 , gladys container must be created with specific argument
```
--cgroupns=host 
-v /sys/fs/cgroup:/sys/fs/cgroup:rw
```
So documentation and rpi image need to be updated too and cgroup "usage" deprecated


VonOx:
@Pierre-Gilles eslint complain about detructuring but I'm not sure how to fix my ugly 'IF' 😓

Pierre-Gilles:
Do we agree that this PR is no longer up to date after what we said on friday?

VonOx:
Yes you are right, I need to push the part concerning cidfile

VonOx:
@Pierre-Gilles @atrovato , I don't really understand why test fail on bluetooth here, any clue ?

server/lib/system/system.getNetworkMode.js
```diff
@@ -14,9 +15,15 @@ async function getNetworkMode() {
   }
 
   if (!this.networkMode) {
-    const cmdResult = await exec('head -1 /proc/self/cgroup | cut -d/ -f3');
-    const [containerId] = cmdResult.split('\n');
-    const gladysContainer = this.dockerode.getContainer(containerId);
+    if (fs.existsSync(`/var/lib/gladysassistant/containerId`)) {
+      this.containerId = fs.readFileSync('/var/lib/gladysassistant/containerId', 'utf8');
```
Pierre-Gilles: Could you use the async version of those functions? 
Otherwise it blocks the whole Node.js process during the file read, which is not ideal when the server is used :) 

it's ok to use those when the code is only running at startup, not when it's a route called during usage

Pierre-Gilles (changes requested):

Pierre-Gilles:
No idea for the bluetooth, I just restarted the build because sometimes the bluetooth fails for no reason (noble is not very stable)

atrovato:
@VonOx 
This comes from your test, and stub on `fs`.
You should better fake and mock fs instead of using real one.

See changes in patch here ;) 

[fix-test-patch.txt](https://github.com/GladysAssistant/Gladys/files/8146507/fix-test-patch.txt)

VonOx:
Thank you @atrovato , but I'm not sure to understand how my test can have an impact on bluetooth test. ( on my view completly independant , in reality not :frowning: )

server/lib/system/system.getNetworkMode.js
```diff
@@ -14,9 +15,17 @@ async function getNetworkMode() {
...
-    const [containerId] = cmdResult.split('\n');
-    const gladysContainer = this.dockerode.getContainer(containerId);
+    fs.access(`/var/lib/gladysassistant/containerId`, fs.constants.F_OK, (err) => {
+      if (err) {
+        // eslint-disable-next-line no-useless-escape
+        this.containerId = exec(
+          'cat /proc/self/cgroup | grep -o -e "docker.*" | head -n 1 | sed -e "s/.scope//g;s/docker-//g;s!docker/!!g"',
+        );
+      } else {
+        this.containerId = fs.readFile('/var/lib/gladysassistant/containerId', 'utf8');
```
Pierre-Gilles, line 25: This won't work, you are using a callback (async) function, the result won't be there.

You need to use async/await.

Two possibilities:

- Promisify fs (for example with bluebird), because fs doesn't return promise by default.
- Use fs-extra, a lib we already use in Gladys: https://github.com/jprichardson/node-fs-extra

In Gladys 4 we don't use the callback pattern (because of the famous callback hell issue), and we rely only on async/await.

server/package.json
```diff
@@ -10,6 +10,7 @@
     "postinstall": "node ./cli/install_service_dependencies.js",
     "pretest": "npm run eslint",
     "test": "cross-env NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/**/*.test.js\" --exit",
+    "test-system": "cross-env NODE_ENV=test ./node_modules/mocha/bin/mocha --recursive ./test/bootstrap.test.js \"./test/lib/system/*.test.js\" --exit",
```
Pierre-Gilles, line 13: No need for that kind of thing, you can put `describe.only` or `it.only` during testing to only run one test.

server/test/lib/system/system.getNetworkMode.test.js
```diff
@@ -4,12 +4,20 @@ const sinon = require('sinon');
...
 const proxyquire = require('proxyquire').noCallThru();
-
+const fs = require('fs');
 const { PlatformNotCompatible } = require('../../../utils/coreErrors');
 const DockerodeMock = require('./DockerodeMock.test');
 
+let fsError = false;
+const fsMock = {
+  ...fs,
+  access: (file, options, callback) => {
```
Pierre-Gilles, line 14: You'll need to mock this as a async function

You can use fake.resolves() to do this (search in the codebase for `fake.resolves` and you will see some examples :)

Pierre-Gilles (changes requested):
I wrote a few feedbacks regarding the callback pattern used here.

Let me know if you have any questions :)

Pierre-Gilles:
> Thank you @atrovato , but I'm not sure to understand how my test can have an impact on bluetooth test. ( on my view completly independant , in reality not 😦 )

You stubbed some functions globally in your test, so it affected other tests too.

server/lib/system/system.getNetworkMode.js
VonOx: OK understood

server/package.json
VonOx: Yeah bad file commit, I was bored to run all tests to seen mine failed, thanks for the tips

server/test/lib/system/system.getNetworkMode.test.js
VonOx: Hard to learn, is Docker test use the same

https://github.com/GladysAssistant/Gladys/blob/master/server/test/lib/system/system.isDocker.test.js#L10

server/lib/system/system.getNetworkMode.js
VonOx: @Pierre-Gilles  what about `const { promises: fs } = require("fs");` ?

Pierre-Gilles: Indeed! I didn't know about this one, I learned something :D 

You'll need to change the access too btw

server/test/lib/system/system.getNetworkMode.test.js
Pierre-Gilles: No! 

One example: https://github.com/GladysAssistant/Gladys/blob/dc25cf7ed3ac36a2962e899e4c152ffe10c2c6ae/server/test/services/mqtt/lib/saveConfiguration.test.js#L23

Pierre-Gilles:
I improved this PR here: https://github.com/GladysAssistant/Gladys/pull/1459

PR #1479, opened 2022-03-22
[CORE] Add more units

VonOx:
### Pull Request check-list

- ~~[ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.~~

### Description of change

Just adding some units asked by community and needed for future integrations ( Z2M Lixee TIC 👀 )


atrovato (dismissed):

front/src/config/i18n/en.json
```diff
@@ -1556,36 +1556,76 @@
   "deviceFeatureUnit": {
```
Pierre-Gilles, line 1556: In this file, the `deviceFeatureCategory` attribute should be filled for new category

front/src/config/i18n/fr.json
```diff
@@ -1556,36 +1556,76 @@
   "deviceFeatureUnit": {
```
Pierre-Gilles, line 1556: Same in this file !

Pierre-Gilles (changes requested):
Nice PR, thanks @VonOx ! 

There are some translations missing in the i18n files, so those attributes could be used in the UI (in the MQTT integration for example)

front/src/config/i18n/en.json
Pierre-Gilles: https://github.com/GladysAssistant/Gladys/blob/master/front/src/config/i18n/en.json#L1608

VonOx (commented):
@Pierre-Gilles deviceFeatureCategory is already filled 😕 

image

Pierre-Gilles:
@VonOx Weird, because I'm not able to use those devices in the MQTT integration. Is it working on your side ?

VonOx:
I miss some mapping in front/server, it's now better

VonOx:
@Pierre-Gilles Can you look again please ?

Terdious:
Hello @VonOx,
To see how to turn the thing, but in connection with the heating, could you add a category "Time" which will be a command and return of state. The goal is to be able to display it either in date+hour+minute or in duration (declared in minutes for example) if > 60 then hours are displayed, otherwise minutes.
So for types:
- a duration in integer I think (the goal here being to be able to display the remaining duration, to increase it or decrease it)
- dateTimeUTC (format 1651346855 for example - the goal here being to display the end date without possible command? or add this one in the time-sensor)
No extra unit I guess.
Tell me what you think of it.
to be able to display something like this:
image

VonOx:
Hi @Terdious 

* The first one is already added ( I think ? ) , but duration seems to be a better term 
```json
"time-sensor": {
      "shortCategoryName": "Time sensor",
      "decimal": "Time (decimal)",
      "integer": "Time (integer)"
    }
```
Combined with unit of your choice

```json
{
    "microseconds": "μs",
    "milliseconds": "ms",
    "seconds": "s",
    "minutes": "min",
    "hours": "h",
    "days": "j",
    "weeks": "sem",
    "months": "m",
    "years": "an."
}
```
* The second one must be calculated I think (dayjs will do the math) , there is no device feature for that ?

front/src/config/i18n/fr.json
```diff
@@ -1760,6 +1828,41 @@
...
+      "index": "Index"
+    },
+    "volume-sensor": {
+      "shortCategoryName": "Capteur de volume",
+      "decimal": "Volume (décimale)",
+      "integer": "Volume (entier)"
+    },
+    "currency": {
+      "shortCategoryName": "Monnaie",
+      "cost": "Coût"
```
Pierre-Gilles: "Coût" is weird, I tried to use this feature without looking at this file, and I couldn't find it. 

Maybe the category should be like "Bourse" ou "Devise", with inside "Monnaie" 

Looks like this now:

Screenshot 2022-05-03 at 09 51 47

Pierre-Gilles: Also, it should be a decimal

front/src/config/i18n/fr.json
```diff
@@ -1760,6 +1828,41 @@
...
+    },
+    "precipitation-sensor": {
+      "shortCategoryName": "Capteur de précipitations",
+      "decimal": "Précipitation"
+    },
+    "uv-sensor": {
+      "shortCategoryName": "Capteur d'UV",
+      "integer": "Indice UV"
+    },
+    "time-sensor": {
```
Pierre-Gilles: "time-sensor" is weird, duration is better!

front/src/config/i18n/en.json
```diff
@@ -1760,6 +1828,41 @@
...
+      "index": "Index"
+    },
+    "volume-sensor": {
+      "shortCategoryName": "Volume sensor",
+      "decimal": "Volume (decimal)",
+      "integer": "Volume (integer)"
+    },
+    "currency": {
+      "shortCategoryName": "Currency",
+      "cost": "Cost"
```
Pierre-Gilles: Same here, should be decimal

"Trading" or "Money" as category name

and inside "Currency"

front/src/config/i18n/fr.json
```diff
@@ -1556,36 +1556,104 @@
...
+    "uv-index": "Indice UV",
+    "microseconds": "μs",
+    "milliseconds": "ms",
+    "seconds": "s",
+    "minutes": "min",
+    "hours": "h",
+    "days": "j",
+    "weeks": "sem",
+    "months": "m",
+    "years": "an."
```
Pierre-Gilles: why the "." ?

front/src/config/i18n/fr.json
```diff
@@ -1760,6 +1828,41 @@
       "stop": "Stop",
       "record": "Enregistrer"
     },
+    "energy-sensor": {
+      "shortCategoryName": "Capteur d'énergie",
```
Pierre-Gilles, line 1832: capteur d'énergie? c'est quoi ?

Pierre-Gilles (changes requested):
Great PR! It'll really help users :) 

I have a few changes.

Don't hesitate to open Gladys when doing this, then going to the MQTT integration, and see how it's looking in the UI, it should be easily searchable and understandable.

Pierre-Gilles:
@Terdious I don't think we want to add this in this PR. It looks like there is some work to think about on heating devices, and if we include this development in this PR, this PR will never go out now!

front/src/config/i18n/fr.json
VonOx: Des devices spécifiques à la mesure de l'énergie

Exemple qui arrive https://www.zigbee2mqtt.io/devices/ZLinky_TIC.htmle

front/src/config/i18n/fr.json
VonOx: C'est l'abréviation de année(s)

front/src/config/i18n/fr.json
VonOx: I agree

Pierre-Gilles (approved):
Ca me parait bien !

PR #1488, opened 2022-03-27
Upgrade cypress

atrovato:
Upgrade Cypress with v9.5.2
And avoid future 429 too many requests on login step.

--- 

In local mode, using cypress:open, Cypress keeps same browser, so local storage memory is kept.
In CI, using cypress:run, Cypress opens a new browser for each spec file (test file), so we loose local storage between each.
We have to use a plugin, which can keep memory at Node level to store shared memory between tests, as access_token.
 
---

The problem will be soon visible, on next cypress tests. I get it on incoming Broadlink service integration.

### DIFF
--- a/front/cypress/integration/routes/0-signup/SignUp.js
+++ b/front/cypress/integration/routes/0-signup/SignUp.js
@@ -121,9 +121,6 @@ describe('Sign-up', () => {
       req => {
         req.reply(res => {
           window.localStorage.setItem('user', JSON.stringify(res.body));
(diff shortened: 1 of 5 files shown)


atrovato:
@Pierre-Gilles without this one, we won't be able to create Cypress tests. Cypress will fail on CI due to "too many request" on login..

Pierre-Gilles (approved):
Looks good to me! Thanks for this PR :)

PR #1496, opened 2022-03-31
Add API request states features

Terdious:
### Pull Request check-list

- [x] Did you test this pull request in real life? With real devices?

### Description of change

The goal of this PR would be to have a local Gladys APIRest request to retrieve all device_features states with a date/time to date/time range.

The intended use would be, in particular, to be able to recover the energy consumed or produced in 1 hour to be able to perform calculations via Node Red. Indeed to date Gladys cannot make calculations, and Node Red does not contain a database (why also store the data 2 times) but we can make calculations there.

Linked to the feature request [Gladys Community](https://community.gladysassistant.com/t/ajout-dune-requete-apirest-gladys-permettant-de-recuperer-les-etats-des-features-sur-une-plage/7127)


Terdious:
@Pierre-Gilles, 
Do you have a link or an explanation to modify the REST API documentation, or is it up to you?

Terdious:
Rendering example of a query on 2 features with Insomnia
image

And same query with max_states at 3
image

Terdious:
Following the recommendations of the feature request, switch to data raw rather than downsampling :
image

Terdious:
@Pierre-Gilles, have you been able to start reviewing this PR? Are there any gaps or anything else?

server/test/lib/device/device.getDeviceFeaturesStates.test.js
```diff
@@ -0,0 +1,168 @@
...
+    };
+    const dateState = `${now.getUTCFullYear()}-${`0${now.getUTCMonth() + 1}`.slice(-2)}-${`0${now.getUTCDate()}`.slice(
+      -2,
+    )}`;
+    const deviceInstance = new Device(event, {}, stateManager, {}, {}, variable);
+    const states = await deviceInstance.getDeviceFeaturesStates('test-device-feature', {
+      from: new Date(`${dateState} 00:00:00.000`).toISOString(),
+      to: new Date(`${dateState} 23:59:59.999`).toISOString(),
+    });
+    expect(states).to.have.lengthOf(23 * 60 + 59 - 1);
```
Pierre-Gilles: why -1 ?

Terdious: Thx @Pierre-Gilles,
If I'm counting correctly: state[0] to state[23*60+59=1439] in total, but the base date is `const now = new Date('2000-06-15 23:59:00.000'); `. So the 1st value is created the day before '2000-06-14 23:59:00.000'
I did this on purpose to test the consideration of the previous day. So in the end we only have to recover from the 2nd state if I'm not being stupid, i.e. state[1] to state[1439] => state[0] to state[1438].

I am wrong ?

Pierre-Gilles: You're not wrong, but it's not clear in the test what you are testing! 

Can you maybe create one event explicitly the day before, and create this series of events during the day explicitely

Also, I'm not sure you need that much events for your test? 

I guess it's because you copied the other tests, and we needed to test the downsampling, and here you are not downsampling, so maybe just inserting 10 events is fine (and will take less time)

Terdious: Hey @Pierre-Gilles,
I reduced the number of states created in the tests and reviewed the test (3h * 60 min). For the rest sorry I'm not sure I understood correctly, you want me to delete one of the following tests:
```
  - it('Should return the full 24h existing state of the device feature - with queries "from" and "to" in GMT')
  - it('should return states between 00:10 and 01:10 with a target between 2000-06-15 00:10 and now using take and skip , only values')
  - it('should return error, device feature does not exist')
  - it('should return error, start date missing')
```

I thought I had to test every possibility at the option level.

Terdious:
Hey @Pierre-Gilles, I hope you're ok!!
Will you have some time to watch before the next Gladys release? If it's good for you, I'd be interested in being able to use it in production ^^

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+async function getDeviceFeaturesStates(selector, options) {
+  const deviceFeature = this.stateManager.get('deviceFeature', selector);
+  if (deviceFeature === null) {
+    throw new NotFoundError('DeviceFeature not found');
+  }
+  if (options.from === undefined) {
+    throw new NotFoundError('Start date missing');
+  }
+
+  // Default from date is one week ago
```
Pierre-Gilles: Might be an old comment, there are no default here

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+  skip: 0,
+  order_dir: 'ASC',
+  order_by: 'created_at',
+};
+
+/**
+ * @description Get all features states aggregates.
+ * @param {string} selector - Device selector.
+ * @param {Object} [options] - Options of the query.
+ * @param {string} [options.from] - Start date in UTC format "yyyy-mm-ddThh:mm:ss:sssZ"
```
Pierre-Gilles: I think this parameter is not optional, so shouldn't be between brackets []

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+ * or "yyyy-mm-dd hh:mm:ss:sss" (GMT time).
+ * @param {string} [options.to] - End date in UTC format "yyyy-mm-ddThh:mm:ss:sssZ"
+ * or "yyyy-mm-dd hh:mm:ss:sss" (GMT time).
+ * @param {number} [options.take] - Number of elements to return.
+ * @param {number} [options.skip] - Number of elements to skip.
+ * @param {string} [options.attributes] - Possible values (separated by a comma ',' if several): 'id',
+ * 'device_feature_id', 'value', 'created_at' and 'updated_at'. Leave empty to have all the columns.
+ * @returns {Promise<Object>} - Resolve with an array of data.
```
Pierre-Gilles:
```
Promise<Array>
```

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+ * @example
+ * device.getDeviceFeaturesStates('test-device', [from: '2022-03-31T00:00:00.000Z', to: '2022-03-31T23:59:59.999Z',
```
Pierre-Gilles: example if not correct, options is not an array

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+ * take: 100, skip: 10, attributes: 'id,value,created_at']);
+ */
+async function getDeviceFeaturesStates(selector, options) {
+  const deviceFeature = this.stateManager.get('deviceFeature', selector);
+  if (deviceFeature === null) {
+    throw new NotFoundError('DeviceFeature not found');
+  }
+  if (options.from === undefined) {
+    throw new NotFoundError('Start date missing');
```
Pierre-Gilles: validation error shouldn't be 404, but 400 bad requests

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,73 @@
...
+    queryParams.limit = optionsWithDefault.take;
+  }
+
+  if (optionsWithDefault.attributes !== undefined) {
+    queryParams.attributes = optionsWithDefault.attributes.split(',');
+  }
+
+  const states = await db.DeviceFeatureState.findAll(queryParams);
+
+  return states;
```
Pierre-Gilles: In that case, this is fine: 

```
return db.DeviceFeatureState.findAll(queryParams);
```

server/test/lib/device/device.getDeviceFeaturesStates.test.js
```diff
@@ -0,0 +1,136 @@
...
+    };
+    const stateManager = {
+      get: fake.returns({
+        id: 'ca91dfdf-55b2-4cf8-a58b-99c0fbf6f5e4',
+        name: 'my-feature',
+      }),
+    };
+    const dateStateFrom = `${now.getUTCFullYear()}-${`0${now.getUTCMonth() + 1}`.slice(-2)}-${`0${now.getUTCDate() -
+      1}`.slice(-2)}`;
+    const dateStateTo = `${now.getUTCFullYear()}-${`0${now.getUTCMonth() + 1}`.slice(
```
Pierre-Gilles: this will break when we'll reach months with 2 digits.

Why not use some static data for those tests?

server/test/controllers/device/device.controller.test.js
```diff
@@ -107,6 +107,38 @@ describe('GET /api/v1/device_feature/aggregated_states', () => {
   });
 });
 
+describe('GET /api/v1/device_feature/:device_feature_selector/states', () => {
```
Pierre-Gilles, line 112: Overall, same feedbacks as before.

This test files looks very complex for a very simple feature.

It's complex because it's a copy paste from another test suite that had a complete different purpose, I think it might be great to rewrite those tests in a way simpler way.

Pierre-Gilles (changes requested):
Just reviewed this PR ! Thanks for your work :) 

There is still some work on the PR, I feel that a lot of the complexity of this PR is coming from the fact that the code has been copy-pasted from the get aggregate feature, which is a complete different feature with different goals.

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Fixed

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Fixed

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Fixed

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Fixed

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Fixed

server/lib/device/device.getDeviceFeaturesStates.js
Terdious: Sorry, I'm not sure I understood everything about this request, are you asking me here for 2 separate things or only the modification of the [] to {} for the option which is an object and not an array?
- [x] Modified the brackets [...] by braces{...} to signify an object and not an array: 
- [ ] Add an example with a 400 bad request return? (FYI if it's about that, I didn't do it because I haven't seen any other example of failure in the current Rest API documentation. But I can add it if needed ^^)

server/test/lib/device/device.getDeviceFeaturesStates.test.js
Terdious: I found this formula on the net, it will never be the case because if 1 digit `now.getUTCMonth() + 1 = 1` then `(01).slice(-2) = 01` and if 2 digit ` now.getUTCMonth() + 1 = 10` then `(010).slice(-2) = 10`.
But you're right, you might as well register it in static. I modify that.

Terdious: Fixed

server/test/controllers/device/device.controller.test.js
Terdious: Hello @Pierre-Gilles,
Indeed it is mainly a copy/paste, but with modifications. The basis of the "states" query being the same. I've read, mostly understood, and edited accordingly to the need I thought (most certainly visibly wrong).
I thought that we really had to test everything to be sure that in the future, even if by mistake a modification of the DB or other happened, we could transcribe it. But in fact I certainly do too much.
So I reduced to the best of my understanding on your vision. Including property checks in this file.
In the file `server/test/lib/device/device.getDeviceFeaturesStates.test.js` I removed the 1st test which effectively is the repetition of this one.
> Pierre-Gilles 2 days ago
> Overall, same feedback as before.

However for this part, I don't see how to put the date in hard since I use the `insertStates` already created in line 10 which uses the date and time of the day. I was able to perform the tests with data for November / December, it did not pose any problems.

> It's complex because it's a copy paste from another test suite that had a complete different purpose, I think it might be great to rewrite those tests in a simpler way.

For this part, I'm sorry, the writing of the tests being already very complex for me, I don't see how to simplify these more after this revision. If that still doesn't fit, do you want me to ask Alex as well if he has some time.

Terdious:
@Pierre-Gilles or @atrovato,
I don't understand, the cypress test failed on a Bluetooth file when nothing has changed and the code has no impact on it. So I guess to a random error.
image
Is there a way to rerun a check on github without updating the PR branch?

Terdious:
@Pierre-Gilles

atrovato:
@Terdious if you rebase with master branch, the failing test will be skiped!

Terdious:
Hello @Pierre-Gilles, would it be possible for you to do a new review of this PR?
Thanks in advance.

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,69 @@
...
+ * device.getDeviceFeaturesStates('test-device', {from: '2022-03-31T00:00:00.000Z', to: '2022-03-31T23:59:59.999Z',
+ * take: 100, skip: 10, attributes: 'id,value,created_at'});
+ */
+async function getDeviceFeaturesStates(selector, options) {
+  const deviceFeature = this.stateManager.get('deviceFeature', selector);
+  if (deviceFeature === null) {
+    throw new NotFoundError('DeviceFeature not found');
+  }
+  if (options.from === undefined) {
+    throw new Error400('Start date missing');
```
Pierre-Gilles, line 35: In the core, we try to avoid using HTTP errors (specific to the API), and instead use core errors. In that case, I would use the "BadParameters" error in the `require('../../utils/coreErrors');`

server/lib/device/device.getDeviceFeaturesStates.js
```diff
@@ -0,0 +1,69 @@
...
+    offset: optionsWithDefault.skip,
+    order: [[optionsWithDefault.order_by, optionsWithDefault.order_dir]],
+  };
+
+  // take is not a default
+  if (optionsWithDefault.take !== undefined) {
+    queryParams.limit = optionsWithDefault.take;
+  }
+
+  if (optionsWithDefault.attributes !== undefined) {
```
Pierre-Gilles, line 61: I think the default should be to have only those attributes returned: 

```
	{
    		created_at: '2022-04-06 10:00:09.225 +00:00',
    		value: 139,
    		id: '30c43c01-0718-40cd-84e0-b975894bd5af'
    }
 ```

(And the user can modify that if needed)

What do you think of that?

(And don't forget to add a test to it)

File: server/test/controllers/device/device.controller.test.js [hunk 12]
```diff
@@ -109,6 +109,34 @@ describe('GET /api/v1/device_feature/aggregated_states', () => {
...
+  beforeEach(async function BeforeEach() {
+    this.timeout(1000);
+    await insertStates(1);
+  });
+  it('should get device feature states by selector', async () => {
+    const now = new Date();
+    const dateState = `${now.getUTCFullYear()}-${`0${now.getUTCMonth() + 1}`.slice(-2)}-${`0${now.getUTCDate()}`.slice(
+      -2,
```
Pierre-Gilles, line 120: We use dayjs everywhere in Gladys, it might be simpler/more readable just to use a simple:


```
dayjs().format('YYYY-MM-DD')
```

No?

server/test/lib/device/device.getDeviceFeaturesStates.test.js
```diff
@@ -0,0 +1,91 @@
...
+};
+
+describe('Device.getDeviceFeaturesStates', function Describe() {
+  this.timeout(5000);
+
+  afterEach(async () => {
+    const queryInterface = db.sequelize.getQueryInterface();
+    await queryInterface.bulkDelete('t_device_feature_state');
+  });
+  it('should return states between 01:10 and 02:09 with a target between 2000-06-15 00:10 and now using take and skip , only created_at and values', async () => {
```
Pierre-Gilles, line 38: The test assertion seems to test a very specific use case, and I don't see anything in the test code that checks that it's working as advertised?

server/test/lib/device/device.getDeviceFeaturesStates.test.js
```diff
@@ -0,0 +1,91 @@
...
+const Job = require('../../../lib/job');
+
+const event = new EventEmitter();
+const job = new Job(event);
+
+const now = new Date('2000-06-15T03:59:00.000Z');
+const insertStates = async () => {
+  const queryInterface = db.sequelize.getQueryInterface();
+  const deviceFeatureStateToInsert = [];
+  const statesToInsert = 3 * 60;
```
Pierre-Gilles, line 16: I think this is still based from the tests from the aggregated states, but it doesn't seem to be really useful to have that much states ?

IMO, you're doing a simple function, you can write a simple test

In your case, you could even hardcode some states here (like 4 states) with fixed dates. 

Then, in your test, you write a query between fixed dates and make sure that only 2 of the 4 states are returned ? (for example)

If you want to test the ?take, you could add another test and do it with ?take=1 for example

server/test/lib/device/device.getDeviceFeaturesStates.test.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+  await queryInterface.bulkInsert('t_device_feature_state', deviceFeatureStateToInsert);
+};
+
+describe('Device.getDeviceFeaturesStates', function Describe() {
+  this.timeout(5000);
+
+  afterEach(async () => {
+    const queryInterface = db.sequelize.getQueryInterface();
+    await queryInterface.bulkDelete('t_device_feature_state');
```
Pierre-Gilles, line 36: This is not useful, database is automatically purged between each tests

server/test/controllers/device/device.controller.test.js
```diff
@@ -109,6 +109,34 @@ describe('GET /api/v1/device_feature/aggregated_states', () => {
...
+    await authenticatedRequest
+      .get('/api/v1/device_feature/test-device-feature/states')
+      .query({
+        from: new Date(`${dateState}T00:00:00.000Z`),
+        to: new Date(`${dateState}T23:59:59.999Z`),
+      })
+      .expect('Content-Type', /json/)
+      .expect(200)
+      .then((res) => {
+        expect(res.body).to.have.lengthOf(2000);
```
Pierre-Gilles, line 132: It's a bit heavy to test returning 2000 state. The goal of inserting 2000 states for aggregated tests was to verify that only 100 of the 2000 were returned in the JSON response.

Here, testing that 10, 200, or 2000 are returned doesn't change anything, it'll just slow the test down on every CI run.

I would suggest testing only with like 50-100 states, it should be enough !

server/api/controllers/device.controller.js
```diff
@@ -106,6 +106,83 @@ module.exports = function DeviceController(gladys) {
...
+   * @api {get} /api/v1/device_feature/:device_feature_selector/states getDeviceFeaturesStates
+   * @apiName getDeviceFeatureStates
+   * @apiGroup Device
+   * @apiParam {string} from - Start date in UTC format "yyyy-mm-ddThh:mm:ss:sssZ"
+   * or "yyyy-mm-dd hh:mm:ss:sss" (GMT time).
+   * @apiParam {string} [to="now"] - End date in UTC format "yyyy-mm-ddThh:mm:ss:sssZ"
+   * or "yyyy-mm-dd hh:mm:ss:sss" (GMT time).
+   * @apiParam {number} [take] - Number of elements to return.
+   * @apiParam {number} [skip=0] - Number of elements to skip.
+   * @apiParam {number} [attributes] - Possible values (separated by a comma ',' if several): 'id',
```
Pierre-Gilles, line 119: *string

Pierre-Gilles (changes requested):
Thanks for the PR! A few feedbacks

PR #1499, opened 2022-04-06
Upgrade critical dependencies

atrovato:
We get an error on check dependency phase at build time.
I ran `npm audit --fix` to fix it.

⚠️ `set-value` dependency has been upgraded to 4.0.1, and requires changes in scene code.


Pierre-Gilles:
Thanks for your PR! I added more tests to verify that scenes were ok, and merged this in https://github.com/GladysAssistant/Gladys/pull/1501

PR #1502, opened 2022-04-12
CI - Codecov token

VonOx:
This prevent codecov action to fail according to codecov dev. ( Even if optional for public repo )

CODECOV_TOKEN secret need to be added obviously


Pierre-Gilles:
I don't think this is the solution to @Terdious problem, this will fail even more because PR don't have access to env variable from the repo (for obvious security reason)

On @Terdious repo it was just a random fail from Codecov no? (their platform is always very buggy)

VonOx:
@Pierre-Gilles it fail on many PR
Workaround is in the output
https://github.com/GladysAssistant/Gladys/runs/5974444144?check_suite_focus=true#step:13:30

PR #1509, opened 2022-04-19
Add clock box for dashboard

callemand:
### Pull Request check-list

### Description of change

The purpose of this PR is a proposal to add box on dashboard to display a clock.

Screenshot 2022-05-08 at 12 41 07
Screenshot 2022-05-08 at 12 41 29

Settings: 
Screenshot 2022-05-08 at 12 42 20
Screenshot 2022-05-08 at 12 43 43

### DIFF
--- a/front/src/components/boxs/clock/Clock.jsx
+++ b/front/src/components/boxs/clock/Clock.jsx
@@ -22,7 +22,7 @@ const Clock = ({ children, ...props }) => (
           <div class={style.analogSmallText}>{props.year}</div>
         </div>
         <div class={style.analogClock}>
-          <ReactClock className='react-clock' value={props.time} renderSecondHand={props.displaySecond} />
+          <ReactClock className="react-clock" value={props.time} renderSecondHand={props.displaySecond} />
         </div>
       </div>
     )}
--- a/front/src/components/boxs/clock/Clock.jsx
+++ b/front/src/components/boxs/clock/Clock.jsx
@@ -9,6 +9,7 @@ import { CLOCK_TYPES } from './ClockTypes';
 import { Text } from 'preact-i18n';
 import style from './style.css';
 import get from 'get-value';
+import cx from 'classnames';
 
 const Clock = ({ children, ...props }) => (
   <div class="card">
@@ -28,7 +29,7 @@ const Clock = ({ children, ...props }) => (
     )}
     {props.clockType === CLOCK_TYPES.DIGITAL && (
       <div class={style.padding}>
-        <div class={style.digitalTime}>{props.time}</div>
+        <div class={cx(style.digitalTime, 'text-monospace')}>{props.time}</div>
         <div class={style.digitalDate}>
           <Text
             id="dashboard.boxes.clock.date"
--- a/front/src/components/boxs/clock/Clock.jsx
+++ b/front/src/components/boxs/clock/Clock.jsx
@@ -22,7 +22,7 @@ const Clock = ({ children, ...props }) => (
           <div class={style.analogSmallText}>{props.year}</div>
         </div>
         <div class={style.analogClock}>
-          <ReactClock className={'react-clock'} value={props.time} renderSecondHand={props.displaySecond} />
+          <ReactClock className='react-clock' value={props.time} renderSecondHand={props.displaySecond} />
         </div>
       </div>
     )}
--- a/front/src/components/boxs/clock/Clock.jsx
+++ b/front/src/components/boxs/clock/Clock.jsx
@@ -1,0 +1,50 @@
+import { Component } from 'preact';
+import { connect } from 'unistore/preact';
+import dayjs from 'dayjs';
+import localizedFormat from 'dayjs/plugin/localizedFormat';
+dayjs.extend(localizedFormat);
+
+const Clock = ({ children, ...props }) => (
+  <div class="card">
+    <div className="card-header">
+      <h3 className="card-title">{props.date}</h3>
+    </div>
+    <div
+      class="card-body o-auto"
+      style={{
+        textAlign: 'center',
+        fontSize: '30px'
+      }}
+    >
+      {props.time}
+    </div>
+  </div>
+);
+
+class ClockComponent extends Component {
+  refreshTime = () => {
+    const date = dayjs()
+      .locale(this.props.user.language)
+      .format('LL');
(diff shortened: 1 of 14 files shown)


front/src/config/i18n/en.json
```diff
@@ -276,6 +277,12 @@
         "no": "No",
         "preview": "Preview",
         "showPreviewButton": "Show preview"
+      },
+      "clock": {
+        "analog": "Analog",
+        "digital": "Digital",
+        "date": "{{day}}, {{month}} {{dayNumber}}, {{year}}",
```
atrovato, line 298: Can't you use dayjs format according to user language instead?

front/src/components/boxs/clock/Clock.jsx
```diff
@@ -0,0 +1,152 @@
...
+  paddingTop: '10px',
+  paddingBottom: '10px'
+};
+
+const Clock = ({ children, ...props }) => (
+  <div class="card">
+    {props.clockType === CLOCK_TYPES.ANALOG && (
+      <div
+        style={{
+          display: 'flex',
```
atrovato: There is too many style, why don't you store element styling in css file?

atrovato (changes requested):
I tried on my computer, but analog clock doesn't work well (left element).

clock


And I think we may have to choice to display seconds or not, it is stressing me :D

front/src/config/i18n/en.json
callemand: Yes, but we cannot have the day without the hours : https://day.js.org/docs/en/display/format#list-of-localized-formats

front/src/components/boxs/clock/Clock.jsx
callemand: Yes you have right, I'm fixing that :)

callemand:
> display seconds or not, i



> I tried on my computer, but analog clock doesn't work well (left element).
> 
> clock
> 

I found the error, It's the default value of the edit box send an undefined, fixing in progress :)

> And I think we may have to choice to display seconds or not, it is stressing me :D

Yes, It's a good idea 👍

callemand:
@atrovato : I have added an options to display or not the second, I have updated the screenshot

front/src/config/i18n/en.json
Pierre-Gilles: You can format as you want: https://day.js.org/docs/en/display/format

atrovato:
I will re-try this, sorry for the delay.

front/src/components/boxs/clock/Clock.jsx
```diff
@@ -0,0 +1,100 @@
...
+      <div class={style.analogRow}>
+        <div class={style.analogCol}>
+          <div class={style.analogSmallText}>{props.day}</div>
+          <div class={style.analogBigText}>
+            <Text id="dashboard.boxes.clock.smallDate" fields={{ month: props.month, dayNumber: props.dayNumber }} />
+          </div>
+          <div class={style.analogSmallText}>{props.year}</div>
+        </div>
+        <div class={style.analogClock}>
+          <ReactClock className={'react-clock'} value={props.time} renderSecondHand={props.displaySecond} />
```
atrovato: You don't need to wrap class between`{'...'}` as you don't use any variable.

front/src/components/boxs/clock/style.css
```diff
@@ -0,0 +1,46 @@
+.padding {
+  padding: 10px 20px 10px 20px;
+}
+
+
```
atrovato: One space is enough here :)

front/src/style/index.css
```diff
@@ -233,3 +233,72 @@ body {
 .device-features {
   min-height: 70px;
 }
+
+
```
atrovato: One space is enough here :)

front/src/style/index.css
```diff
@@ -233,3 +233,72 @@ body {
...
+  border-radius: 50%;
+}
+.react-clock__face {
+  border: 1px solid white;
+}
+.react-clock__second-hand {
+  transition: transform cubic-bezier(0.68, 0, 0.27, 1.55) 0.2s;
+}
+
+
```
atrovato: One space is enough here :)

atrovato (approved):
Ok, it is quite better.
Look's good to me.

atrovato (changes requested):
Please fix conflicts and handle all comments :)

front/src/config/i18n/en.json
callemand: I don't find a good way to use dayjs directly because the `LocalizedFormat` proposed by this library don't have the good display. 
There is no way to add second on the time or have the day without the time. 
If you have any idea :)

Pierre-Gilles: Oh right, I understand. I agree!

server/utils/constants.js
```diff
@@ -835,6 +836,11 @@ const JOB_ERROR_TYPES = {
   UNKNOWN_ERROR: 'unknown-error',
 };
 
+const CLOCK_TYPES = {
+  ANALOG: 'analog',
+  DIGITAL: 'digital',
+};
+
```
Pierre-Gilles: Is it something that really needs to be in the server constants? It seems it's something front only right?

It could just be in the components in the front imo :)

Pierre-Gilles (changes requested):
Thanks for taking the time to update this PR 🙏

The code is good to me expect one little detail (nothing serious)

Functionally, the design of the month seems a bit big to me, I guess it's because it was designed in May when the month was small, and now we are in December 😂

What about making the "17 décembre" smaller, and a bit more compact (less spacing between the two) ? 

Screenshot 2022-12-17 at 15 00 44

server/utils/constants.js
callemand: Yes, it's not used in backend part, I have moved this constant into component

Pierre-Gilles (changes requested):
Thanks for the changes 🙏

Good for me for the code, just can you add the text-monospace class to the numeric time so that there is no "glitch"

See my comment here: https://community.gladysassistant.com/t/afficher-lheure-sur-le-dashboard/7071/38?u=pierre-gilles

Pierre-Gilles (approved):
Good for me now! Thanks for this PR 🙏

PR #1542, opened 2022-05-25
Push docker images to GHCR

VonOx:
Updating CI to push image to Github Container Registry
Good bandwidth and no request limitation.

For those using watchtower with a lot of container its a good choice.

Tested here => https://github.com/VonOx/gh-test-ci/pkgs/container/gladys

### DIFF
--- a/.github/workflows/docker-release-build.yml
+++ b/.github/workflows/docker-release-build.yml
@@ -148,7 +148,7 @@ jobs:
         id: docker_meta
         uses: docker/metadata-action@v3
         with:
-          images: ${{ secrets.DOCKERHUB_REPO }}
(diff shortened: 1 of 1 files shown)


atrovato:
Does it mean that we have to declare a `GITHUB_TOKEN` or Github knows itself?

It looks really clean ;)

VonOx:
Token is available in CI, nothing to do.

Pierre-Gilles:
Very nice, good idea! Do you confirm that we still push to DockerHub, it's just a second destination?

VonOx:
Yes dockerhub stuff still here, just another destination.

Pierre-Gilles (approved):

PR #1343, opened 2021-10-30
Zigbee2mqtt lqi

atrovato:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [ ] If your changes affects code, did your write the tests?
- [ ] Are tests passing? (`npm test` on both front/server)
- [ ] Is the linter passing? (`npm run eslint` on both front/server)
- [ ] Did you run prettier? (`npm run prettier` on both front/server)
- [ ] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- [ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.
- [ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).
- [ ] Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

Please provide a description of the change here. It's always best with screenshots, so don't hesitate to add some!

PR #1546, opened 2022-05-29
Signal dashboard box

atrovato:
### Pull Request check-list

- ~[ ] If your changes affects code, did your write the tests?~
- ~[ ] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)~
- ~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~

### Description of change

Signal quality dashbroad box.

image

It computes a ratio between min/max/value of the device feature to match 0 to 5 int value.


VonOx:
@atrovato I'm not able to test right now. I will ASAP next week.

In your screenshot icon look a bit darker.

Pierre-Gilles (commented):
Nice PR ! :)

A few feedbacks: 

- Icon looks indeed a bit darker than the rest of the UI
- So in the end we are doing the calcul on the frontend side? If we want to use this feature for bluetooth as well, does it work?

VonOx:
Actual icon is webfont, so css color do the job.

Svg use another property `fill` that why icon appear darker

atrovato:
It's not so simple to manage color and size from CSS with this method (css background svg).
I am looking for something else... maybe jsx svg with css color and size.

atrovato:
Second try, with tabler-icons sprite (all icons in 1 svg).
The single SVG is size of 107kb (a bit heavy), but we may add only icons we want, instead of all.
So this way, it allows to override color and size from wrapper "div" with css classes.

image

@Pierre-Gilles I'm not sure about "backend or frontend should compute the ratio", so here, it matches both, as frontend compute the ratio by default.

Pierre-Gilles:
@atrovato We talked a lot about it on the forum :D In this PR, the feature (signal quality) is calculated on the frontend side, and suppose that the value stored in it is a linear scale. If we have another technology that send signal quality data but in another type of scale (log scale?), the frontend calcul doesn't work

Also, in general in Gladys 4, we standardized the format stored in the DB (Ex: color always converted to the same format, etc..) It's not the case here!

Pierre-Gilles:
Discussion on the forum: https://community.gladysassistant.com/t/nouvelle-feature-signal-strength-ou-etat-de-la-connexion/6165/17?u=pierre-gilles

front/src/components/icons/SvgIcon.jsx
```diff
@@ -0,0 +1,17 @@
+const spritePath = '/assets/icons/icon-sprite.svg';
```
Pierre-Gilles: why don't you import the svg here? 

Like

import sprite from `../.../.svg` ?

atrovato: Done, but with absolute path!

VonOx (dismissed):
Looking good after testing

Pierre-Gilles (approved):

VonOx:
@atrovato testing #1343 but for some reason feature always display off

image

image

PR #1557, opened 2022-06-11
Integration sort

atrovato:
According to test on non admin user: https://github.com/GladysAssistant/Gladys/pull/1395#issuecomment-1152109985

I fixed the counters.

1st commit is same as the 1st PR.
2nd commit is the fix.

I miss to get only the length.

### DIFF
--- a/front/src/actions/integration.js
+++ b/front/src/actions/integration.js
@@ -27,7 +27,7 @@ const actions = store => ({
         i => HIDDEN_CATEGORIES_FOR_NON_ADMIN_USERS.indexOf(i.type) === -1
       );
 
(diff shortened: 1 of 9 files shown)


Pierre-Gilles (changes requested):
It seems there is a bug on first load, everytime I test it, it doesn't work on first load but only on second: 

https://user-images.githubusercontent.com/7365207/173329280-c07882b9-48b0-496d-9f9c-dd23162c6809.mov

VonOx:
Reproduced the bug here too

atrovato:
Thank you for feedbacks, I forgot the async on action method.
And I add the "admin system" page management too.
All should be fixed now.

Pierre-Gilles (approved):
Thanks for your fixes, looks fine to me!

PR #1462, opened 2022-03-12
Open/close on scenes #1423

atrovato:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- ~~[ ] If your changes affects code, did your write the tests?~~
- [x] Are tests passing? (`npm test` on both front/server)
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [x] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- ~~[ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.~~
- ~~[ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~
- ~~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~
- [x] Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

Allow to override default On/Off label on scene page according to binary category.

scenes-binary-sensors

Fixes #1423

PR #1578, opened 2022-07-24
Scenes binary sensor labels

atrovato:
Please look at old PR for history #1462 
I fixed small screen view.

### DIFF
--- a/front/src/config/demo.js
+++ b/front/src/config/demo.js
@@ -768,6 +768,24 @@ const data = {
               last_value_changed: '2019-02-12 07:49:07.556 +00:00'
             }
           ]
+        },
+        {
+          id: '284d8f68-220c-45fd-a73a-eccb547aff24',
+          name: 'Window sensor',
+          selector: 'opening-sensor',
+          features: [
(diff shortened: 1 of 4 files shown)


Pierre-Gilles:
@atrovato 

Better, but there is still an alignment problem with the buttons! 

Before:

Screenshot 2022-08-04 at 11 15 37

After:
Screenshot 2022-08-04 at 11 14 44

atrovato:
I think I will take more time on this, the page is not fully responsive.

image

Maybe we need to have one line per form part in mobile view...
I will play with and keep you in touch.

atrovato:
And now? @Pierre-Gilles 

image

Pierre-Gilles:
What about putting the arrow vertically when in mobile ? 

More classic pattern: 

Screenshot 2022-11-25 at 15 48 39

atrovato:
@Pierre-Gilles done ;)

image

Pierre-Gilles (approved):
So much better now !! Love this design on mobile 😄

PR #1576, opened 2022-07-24
Shutter/curtain device

atrovato:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [x] If your changes affects code, did your write the tests?
- [x] Are tests passing? (`npm test` on both front/server)
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [x] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- [ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.
- ~~[ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~
- ~~[ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~
- [x] Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

Add curtain/shutter feature.
Not used by services for now.

PR #1581, opened 2022-08-03
Z2M Add support for smart valve

VonOx:
Related to => https://community.gladysassistant.com/t/vanne-zigbee2mqtt-absence-dun-parametre/7451

image


atrovato:
That's a good question, I have same the problem according to your comment on the cover feature.
I keep thinking about it... I will let you know once I find the solution.

atrovato:
@VonOx do you know how to handle this 3 states feature? In dashboard view?

> ### Timer_state (enum)
> Value can be found in the published state on the `timer_state` property. It's not possible to read (`/get`) this value. To write (`/set`) a value publish a message to topic `zigbee2mqtt/FRIENDLY_NAME/set` with payload `{"timer_state": NEW_VALUE}`. The possible values are: `disabled`, `active`, `enabled`.

You can add the mapping on `disabled`, `active`, `enabled` values in `enumType.js` file.

VonOx:
OK got it , I was afraid of some side effect.

In front I think I will mimic your cover component.

VonOx:
@atrovato @Pierre-Gilles 

What do you think about that ?

image
image
Color change if timer is active

Pierre-Gilles:
@VonOx Nice! Have you tried putting the indicator on the left ?

182862163-d018b135-77a6-4d42-88c1-66fdea36dacc

VonOx:
> @VonOx Nice! Have you tried putting the indicator on the left ?

image

Left of the button group

Pierre-Gilles:
Looks cleaner to me! what do you think of this change?

VonOx:
Based on your comment on #1576  => https://github.com/GladysAssistant/Gladys/pull/1576#issuecomment-1204999633

For button feedback while updating value, what about a spinner that replace the text ? ( I'm waiting for @atrovato feedback too )

Pierre-Gilles:
Spinner inside the button could work!

atrovato:
I'm not sure this matches the feature actions.

The definition in z2m says you can set 1 of this 3 values: `disabled`, `active`, `enabled`.

Where is the `active` value to set? Is the green icon an action button or only an indicator?
Did I miss something?

EDIT: but I prefer the 2nd try, on the left, and I like it ;)

Pierre-Gilles:
@VonOx Is this ready for review? Tested in real life?

VonOx:
Only in demo mode.

Pierre-Gilles:
Ok, can you post a message here to get some testers? https://community.gladysassistant.com/t/vanne-zigbee2mqtt-absence-dun-parametre/7451/11

VonOx:
@Pierre-Gilles there is only one user with this device, z2m stuff is not user friendly for testing. I will ask anyway.

Pierre-Gilles:
@VonOx Let me know when you have news on this ^^

PR #1586, opened 2022-08-12
fix(brain): Improve french camera questions

cicoub13:
### Description of change

Add question `"Montre moi la caméra de la %room%"` to 🇫🇷 camera brain

See https://community.gladysassistant.com/t/releve-de-bugs-divers/6271/138


Pierre-Gilles (approved):
Tested locally and it works fine! Thanks for this PR 🙏

PR #1615, opened 2022-10-10
Add new thermostat temperature device

Pierre-Gilles:
### Pull Request check-list

### Description of change

Kapture 2022-10-14 at 13 29 54


front/src/components/boxs/device-in-room/device-features/ThermostatDeviceFeature.jsx
```diff
@@ -0,0 +1,62 @@
+import get from 'get-value';
+import { Text } from 'preact-i18n';
+
+import { getDeviceName } from './utils';
+import { DeviceFeatureCategoriesIcon } from '../../../../utils/consts';
+
+const NumberDeviceFeature = ({ children, ...props }) => {
```
atrovato: Bad name according to filename.

An another option would be to merger with NumberDeviceFeature component, adding 2 options (step as number and displayUnit = true/false)

atrovato (changes requested):

front/src/components/boxs/device-in-room/device-features/ThermostatDeviceFeature.jsx
Pierre-Gilles: I think we want to keep this as a separate component, as I changed the code to make it more specific :)

VonOx (changes requested):
Component look fine, but I think we need to implement 2 features for thermostats.

* Temperature ( Actual measured température ) `DEVICE_FEATURE_TYPES.THERMOSTAT.TEMPERATURE`
* Target Temperature `DEVICE_FEATURE_TYPES.THERMOSTAT.TARGETTEMPERATURE` 

Display both on the same components is a must have but can be implemented later.

So I suggest renaming to TARGETTEMPERATURE

Pierre-Gilles:
@VonOx Thanks for the review! Indeed, TARGET_TEMPERATURE is more clear

Pierre-Gilles:
@VonOx I updated with the requested changes

PR #1641, opened 2022-11-04
Common scheduler

atrovato:
### Pull Request check-list

#### Gladys core and common scheduler

 - unify usage of the `node-schedule` library
 - add `scheduleJob` in Gladys scheduler (accessible from services)
 - split tests and improve coverage

#### Changes

Now `scheduler-jobs` are crons instead of interval.


server/lib/gateway/gateway.handleAlexaMessage.js
```diff
@@ -37,11 +37,11 @@ async function handleAlexaMessage(data, rawMessage, cb) {
     logger.debug(`gateway.handleAlexaMessage: New message : ${directiveNamespace}`);
     let response;
     if (directiveNamespace === 'Alexa.Discovery') {
-      response = service.alexaHandler.onDiscovery();
+      response = await service.alexaHandler.onDiscovery();
```
Pierre-Gilles: This function is synchronous (see: https://github.com/GladysAssistant/Gladys/blob/master/server/services/alexa/lib/alexa.onDiscovery.js#L10) so I don't think this is needed, I think there was a mistake in the code with the fake.resolve

server/lib/gateway/gateway.login.js
```diff
@@ -21,8 +16,6 @@ async function login(email, password) {
     if (this.gladysGatewayClient) {
       this.gladysGatewayClient.disconnect();
     }
-    // create a new instance of the client
-    this.gladysGatewayClient = new this.GladysGatewayClient({ cryptoLib, serverUrl, logger });
```
Pierre-Gilles: Why removing this from the login function? It's important to create a new client otherwise we risk mixing data from previous login session

I'm not sure I'm super confident on modification on the Gladys Plus code, I can't merge this without super serious tests on my side, and I don't think we want that on this PR 😅

server/lib/scheduler/scheduler.cancelJob.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @description Cancel job
+ * @param {string} jobName - Job name to cancel.
+ * @example
+ * scheduler.cancelJob('my-job');
+ */
+function cancelJob(jobName) {
+  this.nodeSchedule.cancelJob(jobName);
```
Pierre-Gilles, line 8: I can't see this method in the doc from node-schedule? https://github.com/node-schedule/node-schedule

server/test/lib/gateway/gateway.test.js
```diff
@@ -1,1022 +0,0 @@
-const { expect } = require('chai');
```
Pierre-Gilles, line 1: I see that you moved all the tests from the Gladys Gateway part into separate files

Did you just separate the tests without modifying them, or did you modify them ? 

If you modified them, it's going to be super hard to see the diff 😅

server/lib/gateway/gateway.init.js
```diff
@@ -44,13 +44,12 @@ async function init() {
...
-  const rule = new this.schedule.RecurrenceRule();
-  rule.tz = timezone;
-  rule.hour = 0;
-  rule.minute = 0;
-  rule.second = 0;
-
-  this.backupSchedule = this.schedule.scheduleJob(rule, this.checkIfBackupNeeded.bind(this));
+  const rule = { tz: timezone, hour: 0, minute: 0, second: 0 };
+  this.backupSchedule = this.scheduler.scheduleJob(
+    JOB_TYPES.GLADYS_GATEWAY_BACKUP,
```
Pierre-Gilles: Why do you use the JOB_TYPES here? I think we are expecting to emit an event here, so the proper object should be EVENTS, JOB_TYPES is only the key for job types

Pierre-Gilles (changes requested):
Thanks for this PR, great work! 🙂

I added a few comments. I'm not sure I'm confident with this PR modifying Gladys Plus code, it's changing the scope of this review to a whole different level 😅

Gladys Plus has paying customers, so I can't change things without serious real-life testing. I'm really trying to have Gladys Plus to be as stable as possible.

I hope you understand !

server/lib/gateway/gateway.login.js
atrovato: > It's important to create a new client otherwise we risk mixing data from previous login session

Wow I didn't imagine that.
I supposed that the client was stateless.

server/lib/scheduler/scheduler.cancelJob.js
atrovato: Not in the doc, but in the code ;)

https://github.com/node-schedule/node-schedule/blob/master/index.js

server/test/lib/gateway/gateway.test.js
atrovato: I'll take time to ensure that I only split them.

server/lib/gateway/gateway.init.js
atrovato: The first param of scheduleJob method is not an event, but the name of the job.
It can be used to retrieve or cancel the job by its name.
And it ensure the job is a singleton.

The scheduler here, core scheduler, is based on `node-schedule` the once used before my changes.

server/lib/gateway/gateway.login.js
Pierre-Gilles: The client is entirely stateful as it manages a long-live websocket connections with Gladys plus. 

The disconnect function should remove every state, but just to be safe I prefer to instantiate a new instance 🙂

server/test/lib/gateway/gateway.test.js
Pierre-Gilles: Sorry to ask this but this part is critical so I really want to review all the changes 😅

server/lib/gateway/gateway.login.js
atrovato: Rollback done

atrovato:
I totally understand, I rollback some un-necessary changes, but some are required to use the common scheduler.
I will generate a "easy-to-read" diff file to check that tests are not that impacted.

Pierre-Gilles:
@atrovato Thank you 🙏 Let me know when it's ready for review again

atrovato:
Hi @Pierre-Gilles, I'm ready for a new review.
There is some comments that I didn't mark as resolved to let you check easily if it's fine for you.

Please find a comparison between "ol" and "new" version of tests.
At left side, the old one.
At right side, the new one (before and after where cleaned in comparison to make it clearer).

[out2.zip](https://github.com/GladysAssistant/Gladys/files/10200012/out2.zip)

server/lib/gateway/gateway.init.js
Pierre-Gilles: But if we use different sources of data for this, it means we risk having jobs with same name (as it's not clear where all the jobs names are stored) ?

server/lib/scene/scene.addScene.js
```diff
@@ -68,7 +68,7 @@ function addScene(sceneRaw) {
           default:
             throw new BadParameters(`${trigger.scheduler_type} not supported`);
         }
-        trigger.nodeScheduleJob = this.schedule.scheduleJob(rule, () =>
+        trigger.nodeScheduleJob = this.scheduler.scheduleJob(scene.selector, rule, () =>
```
Pierre-Gilles: If my scene selector is the same as a native internal job, it'll create a job with the same name as a native job, making it hard to cancel the job

Maybe let's add a prefix?

server/lib/scene/scene.init.js
```diff
@@ -24,17 +24,14 @@ async function init() {
...
   // Recurrence rule (00:00 every day) to update sunrise/sunset time.
-  const rule = new this.schedule.RecurrenceRule();
-  rule.tz = this.timezone;
-  rule.hour = 0;
-  rule.minute = 0;
-  rule.second = 0;
-
-  this.schedule.scheduleJob(rule, this.dailyUpdate.bind(this));
+  const rule = { tz: this.timezone, hour: 0, minute: 0, second: 0 };
+  this.scheduler.scheduleJob('scene-daily-update', rule, this.dailyUpdate.bind(this));
```
Pierre-Gilles: Name constant could be stored somewhere

server/lib/scheduler/scheduler.scheduleJob.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Object|string|Date} rule - Rule to execute the job.
+ * @param {Function} method - The method to execute.
+ * @returns {Object} The scheduled job.
+ * @example
+ * scheduler.scheduleJob('my-job', { hour: 12 }, () => console.log('job is running'));
+ */
+function scheduleJob(name, rule, method) {
+  let wrappedRule = rule;
+
+  if (typeof rule === 'object') {
```
Pierre-Gilles: Not sure this condition is working :D 

Dates are also catched: 

Screenshot 2022-12-12 at 13 54 32

server/test/lib/gateway/gateway.downloadBackup.test.js
```diff
@@ -0,0 +1,85 @@
...
+      expect(e)
+        .instanceOf(NotFoundError)
+        .haveOwnProperty('message', 'GLADYS_GATEWAY_BACKUP_KEY_NOT_FOUND');
+    }
+
+    assert.notCalled(event.emit);
+  });
+
+  it('should download a backup', async () => {
+    const encryptedBackupFilePath = path.join(__dirname, 'encoded-gladys-db-backup.db.gz.enc');
```
Pierre-Gilles, line 75: Be careful because now this file is stored in the code and in this PR the file is there too! (don't think it's useful)

Before it was stored in a temp folder

server/test/lib/gateway/gateway.restoreBackup.test.js
```diff
@@ -0,0 +1,100 @@
...
+    gateway = new Gateway(variable, event, {}, sequelize, config, {}, {}, {}, job, scheduler);
+    gateway.config.storage = '/tmp/gladys-database-restore-test.db';
+  });
+
+  afterEach(() => {
+    sinon.reset();
+  });
+
+  it('should restore a backup', async () => {
+    const backupFilePath = path.join(__dirname, 'real-gladys-db-backup.db.gz.dbfile');
```
Pierre-Gilles, line 61: Same feedback here, the file was created in the code

server/test/lib/scheduler/scheduler.scheduleJob.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const rule = '*/5 * * * * *';
+    const jobMethod = fake.returns(true);
+
+    scheduler.scheduleJob(jobName, rule, jobMethod);
+
+    assert.notCalled(nodeScheduleMock.RecurrenceRule);
+    assert.calledOnceWithExactly(nodeScheduleMock.scheduleJob, jobName, rule, jobMethod);
+  });
+
+  it('should schedule object ruled job', async () => {
```
Pierre-Gilles, line 37: Maybe let's add a test with a fixed time (Date) job?

Pierre-Gilles (changes requested):
Thanks for your fixes! It's super clear to me, and I agree with the improvements in the Gateway tests, thanks for them 🙏

I have a few feedbacks but nothing really serious :)

server/lib/gateway/gateway.init.js
atrovato: According to `node-schedule` examples, I can ignore the name of jobs.
This means that we have to keep generated job in our process to cancel or reschedule it as a singleton if needed.

server/lib/scene/scene.init.js
atrovato: According to "name comment", I will remove this param.

server/lib/scene/scene.addScene.js
atrovato: According to last comment on "names", I will remove the name param.

server/lib/scheduler/scheduler.scheduleJob.js
atrovato: I checked that the lib already do this, I will remove this part.

server/test/lib/gateway/gateway.downloadBackup.test.js
atrovato: I was generated from a test, which was not a unit one, but handled two methods.
But I can generate it in `beforeEach` method instead of persist it in git.

server/test/lib/scheduler/scheduler.scheduleJob.test.js
atrovato: I will

server/test/lib/gateway/gateway.downloadBackup.test.js
Pierre-Gilles: No problem to persist it in git if it was on purpose, just wanted to check with you that it was not an unwanted side effect. If I run the tests locally, it won't update the file in git?

Pierre-Gilles:
@atrovato is this ready for a new review ? Let me know :)

atrovato:
I just rebased it, yes ready for new review :) @Pierre-Gilles

Pierre-Gilles:
@atrovato I reviewed the code and it looks good to me. But I need to test it in real-life because it's an important change! 

Do you have a Docker image to try it ?

atrovato:
`atrovato/gladys:common-scheduler` is building for now ;)

Pierre-Gilles:
@atrovato I started an instance with your image. Let's see in the coming day if all jobs are executing as expected

Which behavior have you tested in real life (just to know) ? 

**Needs to be tested :**

- [x] Hourly aggregate
- [x] Sunrise/Sunset scenes
- [x] Creating a scene
- [x] Daily backups
- [x] Gateway login
- [x] Amazon Alexa integration 
- [x] Google Home integration

Pierre-Gilles (approved):
Tested on my side in real life, thanks for this PR, everything works fine! 🙏

PR #1645, opened 2022-11-08
Scene action toggle

rpochet:
### Pull Request check-list

- [ x] If your changes affects code, did your write the tests?
- [ x] Are tests passing? (`npm test` on both front/server)
- [ x] Is the linter passing? (`npm run eslint` on both front/server)
- [ x] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- [ x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.

### Description of change

Add action toggle ON <> OFF to scene


Pierre-Gilles:
@rpochet Tell me when this is ready for review :)

rpochet:
C'est bon pour moi

front/src/config/i18n/fr.json
```diff
@@ -1076,12 +1076,18 @@
       "turnOffLights": {
         "label": "Sélectionnez les lumières que vous souhaitez éteindre"
       },
+      "toggleLights": {
+        "label": "Sélectionnez les lumière que vous souhaitez inverser"
```
Pierre-Gilles: *lumières

Pierre-Gilles (changes requested):
Thanks for this PR 🙏

it works fine on my side, good job! 

Screenshot 2022-12-18 at 16 13 45

Just a little UX feedback, maybe we should explain in the action what does this action do? 

Screenshot 2022-12-18 at 16 10 16

Like a little text saying "Si les lumières sont allumées, elle seront éteintes, et vice-versa".

What do you think?

front/src/config/i18n/fr.json
rpochet: it is true that in french it is not so pretty...

What about this?
image

But OK for a more readable little text

Pierre-Gilles: It's a bit dirty to have this long text in bold, what about having a shorter text in bold and an explanation in small ?

rpochet: I keep the same layout as in other actions. Should not be part of a new PR?

Pierre-Gilles: Yeah but other actions are not that long, so it's a problem for this one! 

Can be a conditional translations if you still want to re-use the same components

Pierre-Gilles (approved):
Good for me now with the changes! Thanks for taking the time 🙏

PR #1650, opened 2022-11-13
Add contact sensor on HomeKit & some fix

bertrandda:
### Pull Request check-list

### Description of change

Add contact sensors (door, window...)
Fix bulb brightness range
Notify HomeKit of new value


server/services/homekit/lib/notifyChange.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * @description Create HomeKit accessory service.
+ * @param {Object} accessories - HomeKit accessories.
+ * @param {Object} event - Gladys event to forward to HomeKit.
+ * @returns {undefined}
+ * @example
+ * notifyChange(accessories, event)
+ */
+function notifyChange(accessories, event) {
```
Pierre-Gilles, line 12: Just a small feedback: Sending every device state change to HomeKit is super heavy, some device send many state changes per second, and usually the only relevant information is the last value.

In the Alexa/Google Home integration, we send only 1 state per 5 seconds maximum per device. That's what Google recommended. Only the "last" value is relevant. 

See: https://github.com/GladysAssistant/Gladys/blob/master/server/lib/gateway/gateway.forwardDeviceStateToAlexa.js#L109

and: https://github.com/GladysAssistant/Gladys/blob/master/server/lib/gateway/index.js#L55

Pierre-Gilles (changes requested):
Nice PR! 🙏

server/services/homekit/lib/notifyChange.js
bertrandda: I improved that as asked (see https://community.gladysassistant.com/t/integration-homekit-et-siri/7493/48)

Pierre-Gilles: Looks way better now! Thanks for the changes

Pierre-Gilles (dismissed):
Thanks for your changes 🙏

Good for me ! 

Just need confirmation before merging, this can be deployed in PROD without any issues? People won't have to repair anything?

server/services/homekit/lib/createBridge.js
```diff
@@ -35,6 +37,8 @@ async function createBridge() {
     .map((device) => this.buildAccessory(device))
     .filter((accessory) => accessory !== null);
 
+  this.gladys.event.on(EVENTS.TRIGGERS.CHECK, eventFunctionWrapper(this.notifyChange.bind(this, accessories)));
```
Pierre-Gilles: I just re-read the code and found this, it seems to me that every time a home kit bridge is created, it creates a subscription on `this.gladys.event.on(EVENTS.TRIGGERS.CHECK`. It'll create multiple subscription, no ?

bertrandda: Indeed, it will create multiple subscriptions 🤔 for example when we reload the bridge after adding new accessory to Gladys. Thanks to delay before notify it should send only 1 message but need to improve that.
I quickly check `lib/event` file not sure how to delete previous one before create the new one. Any idea?

bertrandda: Maybe we need to implement `removeListener()` in `Event`?

Pierre-Gilles: Indeed, we miss a removeListener!

Pierre-Gilles: https://nodejs.org/api/events.html#emitterremovelistenereventname-listener

bertrandda: Done! I added `removeListener` on our Event implementation and use it for HomeKit reload

server/test/lib/event/event.test.js
```diff
@@ -15,4 +15,21 @@ describe('event', () => {
     // because the event emitter will be called before
     expect(alreadySeen).to.equal(false);
   });
+
+  it('should verify that event listener is removed', async () => {
```
Pierre-Gilles, line 19: I don't think this is testing the behavior, because our event emitting system is async, so even if you remove the 

> eventEmitter.removeListener('test', cb);

Line, you'll see that alreadySeen will still be false

bertrandda: Yes I was not sure about this test. I just found a better way by implementing and using `listenerCount`

Pierre-Gilles (dismissed):
Looks good to me now! 

Can it be merged to PROD? Was it tested in real life with latest changes? Will users needs to re-pair their devices?

bertrandda:
> Can it be merged to PROD? Was it tested in real life with latest changes? Will users needs to re-pair their devices?

I am going to create a  last image with last master changes and ask for a last test but yes I think everything fonctionnally as been tested.
Users will not have to repair their devices, just define new device's room in Home app

Pierre-Gilles:
> I am going to create a last image with last master changes and ask for a last test but yes I think everything fonctionnally as been tested.

Ok, let me know here when it's good for merge.

Pierre-Gilles:
@bertrandda  Same comment here, if we don't have more testers in one week, and you tested carefully on your side, I'm ok to merge on master :) Ping me if I forget!

PR #1656, opened 2022-11-28
Upgrade to Node.js 18 LTS

Pierre-Gilles:
### Pull Request check-list

### Description of change

Upgrading to Node.js 18 LTS


bertrandda:
We could take the opportunity to remove `get-value` dependency (at least on server code) we can replace almost everywhere get() by optional chaining operator `?.` Nodejs native

.github/workflows/docker-pr-build.yml
```diff
@@ -8,21 +8,16 @@ on:
 jobs:
```
NickDub, line 12: tips: Maybe you should add the following lines at the beginning of the YAML to avoid unnecessary parallel launches:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref }}
  cancel-in-progress: true
```

.github/workflows/docker-dev-build.yml
```diff
@@ -15,21 +15,16 @@ on:
 jobs:
   build-front:
     name: Front build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
NickDub, line 18: Why not use `ubuntu-latest` everywhere?

.github/workflows/docker-dev-build.yml
```diff
@@ -15,21 +15,16 @@ on:
...
         with:
-          node-version: '14'
-      - name: 📇 Use npm cache
-        uses: actions/cache@v2
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
+          node-version: '16'
```
NickDub:
```suggestion
          node-version-file: 'front/package.json'
```

NickDub: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

.github/workflows/docker-pr-build.yml
Pierre-Gilles: Good idea!

.github/workflows/docker-dev-build.yml
Pierre-Gilles: Using non-fixed version in CI is the best way to have PR broken when Ubuntu publish a big upgrade IMO ^^

I prefer to have stability, and upgrade manually when needed

.github/workflows/docker-dev-build.yml
Pierre-Gilles: Good idea!

Pierre-Gilles: Done

.github/workflows/docker-pr-build.yml
Pierre-Gilles: Done

Pierre-Gilles:
@bertrandda Good idea! But in a separate PR, as it will be quite a big change

Pierre-Gilles:
@atrovato I added a skip on the bluetooth cypress scan tests because it's really failing too often :/ I know it's not ideal but it's very frustrating

Pierre-Gilles:
I managed to upgrade to Node 18 everywhere so it's easier for everyone 😄

PR #1669, opened 2022-12-11
MQTT: Fake "promise timeout" in MQTT tests to increase MQTT tests speed

atrovato:
Fake "promise timeout" in mqtt tests.


Pierre-Gilles (approved):
Thanks for this PR!! It'll improve test speed, always good for the developer experience 😄

PR #1670, opened 2022-12-13
Duplicate scene

callemand:
### Pull Request check-list

### Description of change

The purpose of this PR is a proposal to add scene duplication on edit scene page.

Screenshot 2022-12-12 at 19 52 49

Screenshot 2022-04-15 at 14 40 15

### DIFF
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -1042,6 +1042,7 @@
     "startButton": "Start",
     "saveButton": "Save",
     "deleteButton": "Delete",
+    "duplicateButton": "Duplicate",
     "triggersTitle": "Triggers",
     "newTrigger": "New trigger",
     "addTriggerButton": "Add trigger",
@@ -1607,6 +1608,14 @@
     "createSceneButton": "Create scene",
     "invalidName": "Name is required and cannot be 'new'",
     "invalidIcon": "Icon is required",
+    "sceneAlreadyExist": "A scene with that name already exist."
+  },
+  "duplicateScene": {
+    "cardTitle": "Duplicate a scene",
+    "nameLabel": "Name",
+    "namePlaceholder": "Enter the name of your scene",
+    "duplicateSceneButton": "Duplicate scene",
+    "invalidName": "Name is required",
     "sceneAlreadyExist": "A scene with that name already exist."
   },
   "scene": {
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -1613,7 +1613,7 @@
   "duplicateScene": {
     "cardTitle": "Duplicate scene \"{{name}}\"",
     "nameLabel": "Name",
-    "namePlaceholder": "Enter a new name for the duplicate scene, this must be different from the previous scene name.",
+    "namePlaceholder": "Enter a new name",
     "iconLabel": "Select an icon for you scene",
     "duplicateSceneButton": "Duplicate",
     "invalidName": "Name is required",
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -1614,8 +1614,10 @@
     "cardTitle": "Duplicate scene \"{{name}}\"",
     "nameLabel": "Name",
     "namePlaceholder": "Enter a new name for the duplicate scene, this must be different from the previous scene name.",
+    "iconLabel": "Select an icon for you scene",
     "duplicateSceneButton": "Duplicate",
     "invalidName": "Name is required",
+    "invalidIcon": "Icon is required",
     "sceneAlreadyExist": "A scene with that name already exist."
   },
   "scene": {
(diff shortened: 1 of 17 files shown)


front/src/config/i18n/en.json
```diff
@@ -1609,6 +1610,16 @@
     "invalidIcon": "Icon is required",
     "sceneAlreadyExist": "A scene with that name already exist."
   },
+  "duplicateScene": {
+    "cardTitle": "Duplicate scene \"{{name}}\"",
+    "nameLabel": "Name",
+    "namePlaceholder": "Enter a new name for the duplicate scene, this must be different from the previous scene name.",
```
Pierre-Gilles: This translation is very long (we can't see the full text). "duplicate" should be "duplicated"

Why not simply: 

> Enter a new name

?

front/src/config/i18n/fr.json
```diff
@@ -1609,6 +1610,16 @@
     "invalidIcon": "L'icône est requise",
     "sceneAlreadyExist": "Une scène avec le même nom existe déjà."
   },
+  "duplicateScene": {
+    "cardTitle": "Dupliquer la scène \"{{name}}\"",
+    "nameLabel": "Nom",
+    "namePlaceholder": "Entrez un nouveau nom pour la scène dupliquée, celui-ci doit être différent du nom de scène précédent.",
```
Pierre-Gilles: Title is too long for a placeholder, we can't read it: 

Screenshot 2022-12-16 at 11 33 57

Why not:

> "Entrez un nouveau nom"

If you want to add an explanation, it should be added outside of the placeholder

server/test/lib/scene/scene.duplicate.test.js
```diff
@@ -0,0 +1,24 @@
...
+const SceneManager = require('../../../lib/scene');
+
+const event = new EventEmitter();
+
+describe('SceneManager', () => {
+  it('should duplicate a scene', async () => {
+    const sceneManager = new SceneManager({}, event);
+
+    const duplicatedScene = await sceneManager.duplicate('to-duplicate-scene', 'new-name', 'new-icon');
+    expect(sceneManager.scenes['new-name']).not.to.be.undefined; // eslint-disable-line
```
Pierre-Gilles: To avoid having to disable eslint for this, you can do `.not.to.equal(undefined)` for example

server/test/lib/scene/scene.duplicate.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+    expect(duplicatedScene).to.have.property('selector', 'new-name');
+    expect(duplicatedScene).to.have.property('name', 'new-name');
+    expect(duplicatedScene).to.have.property('icon', 'new-icon');
```
Pierre-Gilles, line 17: Can you add a bit more checks? What about the triggers? What about the actions?

It should be strictly verified with a .deep.equal() imo :)

server/lib/scene/scene.duplicate.js
```diff
@@ -0,0 +1,46 @@
...
+    actions: plainExistingScene.actions,
+    triggers: plainExistingScene.triggers,
+  };
+
+  const createdScene = await db.Scene.create(newScene);
+  const plainScene = createdScene.get({ plain: true });
+  // add scene to live store
+  this.addScene(plainScene);
+  // return created scene
+  return plainScene;
```
Pierre-Gilles: Just a small feedback, maybe it could have been wise here simply to call: 

```
return this.create(newScene);
```

? 

This way we avoid duplicating code, and if we add new things to the scene.create function, we avoid having to copy-paste code here :)

Pierre-Gilles (changes requested):
Thank you so much for this PR 🙏 

This is great work and it works very well on my machine 😄 

I just have a few small feedbacks, it's really nothing.

After that, I'm good to merge!

server/lib/scene/scene.duplicate.js
callemand: Yes, you are completely right 👍

front/src/routes/scene/duplicate-scene/index.js
```diff
@@ -0,0 +1,130 @@
...
+import get from 'get-value';
+
+@connect('httpClient', {})
+class DuplicateScene extends Component {
+  goBack = async () => {
+    route(`/dashboard/scene/${this.props.scene_selector}`);
+  };
+
+  getSourceScene = () => {
+    this.props.httpClient.get(`/api/v1/scene/${this.props.scene_selector}`).then(scene => {
```
Pierre-Gilles: Oups sorry I didn't see this during last review, can you use async/await ? 

We try to avoid using .then because it leads to long chain of .then

Pierre-Gilles (changes requested):
Thanks for your changes it's perfect now 🙏 

I just found a last tiny 1 line change that I didn't see last time, sorry 😂

I merge this right after the change! Thanks!

front/src/routes/scene/duplicate-scene/index.js
callemand: I prefer the await too :)

Pierre-Gilles (approved):
Thanks for the changes, it's good for me ! ✅

PR #1073, opened 2021-02-11
[WIP] Tasks Todoist

callemand:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [x] If your changes affects code, did your write the tests?
- [x] Are tests passing? (`npm test` on both front/server)
- [X] Is the linter passing? (`npm run eslint` on both front/server)
- [X] Did you run prettier? (`npm run prettier` on both front/server)
- [X] If you are adding a new features/services, did you run integration comparator? (`npm run compare-translations` on front)
- [ ] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).
- [X] Did you add fake requests data for the demo mode (`front/src/config/demo.json`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

Please provide a description of the change here. It's always best with screenshots, so don't hesitate to add some!

It's a fork on https://github.com/GladysAssistant/Gladys/pull/960 
It's just to add the tests and to do some graphic adjustment

Screenshot 2021-02-11 at 14 38 56
Screenshot 2021-02-11 at 14 41 15
Screenshot 2021-02-11 at 14 41 20

PR #1678, opened 2023-01-01
Tasks Todoist

callemand:
### Pull Request check-list

- [X] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.
- [X] If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)

### Description of change

I have made some adjustment from this PR : https://github.com/GladysAssistant/Gladys/pull/1073
Adding todolist box on the dashboard. for the moment it's connected to Todoist. But my goal is to connect Google Tasks and some other tools from google :)
107644064-3b7eb600-6c77-11eb-8d1e-65fcc4b48089
107644081-3faad380-6c77-11eb-8d7e-c72db66e48c4
107643782-e478e100-6c76-11eb-8f5a-a83e3cd26f66


Pierre-Gilles:
Very nice! :) Tell me when you need a review on this !

callemand:
@Pierre-Gilles : I think, it's ready for review, if you have some times :)

front/src/actions/dashboard/boxes/todolist.js
```diff
@@ -0,0 +1,48 @@
...
+        }
+      }
+    },
+
+    async closeTask(state, taskId, x, y) {
+      let { tasks } = boxActions.getBoxData(state, BOX_KEY, x, y);
+      tasks = tasks.map(t => (t.id === taskId ? { ...t, pending: true } : t));
+      boxActions.mergeBoxData(state, BOX_KEY, x, y, { tasks });
+
+      await Promise.all([state.httpClient.post(`/api/v1/todolist/tasks/${taskId}/close`), waitTime(1500)]);
```
atrovato, line 48: Can you tell why you wait here?

front/src/components/app.jsx
```diff
@@ -223,13 +225,15 @@ const AppRouter = connect(
         <EditXiaomiPage path="/dashboard/integration/device/xiaomi/edit/:deviceSelector" />
         <TasmotaPage path="/dashboard/integration/device/tasmota" />
         <TasmotaEditPage path="/dashboard/integration/device/tasmota/edit/:deviceSelector" />
+
```
atrovato: Oups, not needed change here :)

front/src/components/boxs/todolist/EditTodolistBox.jsx
```diff
@@ -0,0 +1,79 @@
...
+      <label>
+        <Text id="dashboard.boxes.todolist.editProjectLabel" />
+      </label>
+      <select onChange={props.updateTodolist} class="form-control">
+        <option>
+          <Text id="global.emptySelectOption" />
+        </option>
+        {props.todolist &&
+          props.todolist.map(todolist => (
+            <option selected={todolist.id === String(props.box.todolist_id)} value={todolist.id}>
```
atrovato: Do you have a reason to use String construction instead of String wrapper?
`${props.box.todolist_id}`?

front/src/components/boxs/todolist/EditTodolistBox.jsx
```diff
@@ -0,0 +1,79 @@
...
+      </select>
+    </div>
+  </BaseEditBox>
+);
+
+@connect('httpClient')
+class EditTodolistBoxComponent extends Component {
+  updateTodolist = e => {
+    this.props.updateBoxConfig(this.props.x, this.props.y, {
+      todolist_id: e.target.value ? String(e.target.value) : undefined
```
atrovato: Same question here.

> Do you have a reason to use String construction instead of String wrapper?
> ${props.box.todolist_id}?

The target value isn't already a String?

front/src/components/boxs/todolist/EditTodolistBox.jsx
```diff
@@ -0,0 +1,79 @@
...
+    });
+    try {
+      const todolist = await this.props.httpClient.get('/api/v1/todolist');
+      this.setState({
+        todolist,
+        loading: false
+      });
+    } catch (e) {
+      await this.setState({
+        loading: true
```
atrovato: In case of error while fetching data, do we really want to go back at a "loading" state?
Don't we want to show an error?

front/src/components/boxs/todolist/TodolistBox.jsx
```diff
@@ -0,0 +1,158 @@
...
+  const textStyle = {
+    textDecoration: task.pending ? 'line-through' : undefined
+  };
+
+  return (
+    <div class={style.todolistTask}>
+      <div class="container">
+        <div class="row">
+          <span>
+            <i className={`fe fe-${icon}`} style={{ marginRight: '10px', cursor: 'pointer' }} onClick={onClick} />
```
atrovato: We usually don't use `className` attribute.
And I'm not fan of explicit styles in HTML element, please prefer class usage (`mr-2 cursor-pointer`).

front/src/components/boxs/todolist/TodolistBox.jsx
```diff
@@ -0,0 +1,158 @@
...
+              </span>
+            </p>
+          </div>
+        )}
+        {props.boxStatus === RequestStatus.Getting && !props.tasks && (
+          <div>
+            <div class="card-body">
+              <div class="dimmer active">
+                <div class="loader" />
+                <div class="dimmer-content" style={padding} />
```
atrovato: Please use `p-1` class instead of explicit css style.

front/src/components/boxs/todolist/TodolistBox.jsx
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  componentDidMount() {
+    const { todolist_id: todolistId } = this.props.box;
+
+    // get tasks
+    this.props.getTasks(todolistId, this.props.x, this.props.y);
+
+    // refresh tasks every interval
+    this.refreshInterval = setInterval(
```
atrovato: Don't we want to use events over web socket instead of interval refresh time?

front/src/utils/consts.js
```diff
@@ -94,6 +94,10 @@ export const GetWeatherModes = {
   DailyForecast: 'dailyForecast'
 };
 
+export const GetTodolistStatus = {
```
atrovato: Can't we use the `RequestStatus` enum instead of adding a new one?

server/api/controllers/todolist.controller.js
```diff
@@ -0,0 +1,53 @@
...
+   * @apiName getTasks
+   * @apiGroup Todolist
+   */
+  async function getTasks(req, res) {
+    const tasks = await gladys.todolist.getTasks();
+    res.json(tasks);
+  }
+
+  /**
+   * @api {get} /api/v1/todolist Get Todolist
```
atrovato: A question, what does `Get Todolist` means?

server/api/controllers/todolist.controller.js
```diff
@@ -0,0 +1,53 @@
...
+   * @apiGroup Todolist
+   */
+  async function getTasksByTodolistId(req, res) {
+    const todolistId = req.params.todolist_id;
+    const tasks = await gladys.todolist.getTasks({ todolist_id: todolistId });
+    res.json(tasks);
+  }
+
+  /**
+   * @api {post} /api/v1/todolist/tasks/:task_id/close close Todolist task
```
atrovato: `close Todolist task` please uppercase first letter of the the first word :)

server/lib/todolist/todolist.closeTask.js
```diff
@@ -0,0 +1,19 @@
+const { ServiceNotConfiguredError } = require('../../utils/coreErrors');
+
+/**
+ * @description close task of the todolist.
+ * @param {string} taskId - The task id.
+ * @example
+ * gladys.todolist.closeTask('taskId');
+ */
+function closeTask(taskId) {
+  const todoistService = this.service.getService('todoist');
```
atrovato, line 10: I'm a bit confused...
It looks like you add the "task feature" in Gladys core, but it uses only the `todoist` specific service.

I think it  a good think to add a "todo list" abstraction in Gladys, but it is not an abstraction here.
I think you miss the "service" in your APIs...
I may be wrong of course...

atrovato (commented):
Hi, I review your code, good job 👏 
I added some remarks, open to discuss ;)

server/lib/todolist/todolist.closeTask.js
Pierre-Gilles: This is the same way of doing as we do in the "message" and "weather" integration, today it only uses one integration, but just because nobody developed another one ^^

I don't think it's necessarily wrong!

server/api/controllers/todolist.controller.js
callemand: I have change the comment to `Get list of todolist`

front/src/utils/consts.js
callemand: Yes, I can

front/src/components/boxs/todolist/TodolistBox.jsx
callemand: I cannot use websocket because the todoist api is on on RestApi. So I need to call the api to refresh the list of the task like the weather boxes (I don't like `setInterval` anymore but I don't find a best way to do this)

front/src/components/boxs/todolist/EditTodolistBox.jsx
callemand: Oops

front/src/components/boxs/todolist/EditTodolistBox.jsx
callemand: It's a mistake on my side

front/src/actions/dashboard/boxes/todolist.js
callemand: It's just to display task with strikethrough text on the interface and after that the tasks disappear.

front/src/components/boxs/todolist/EditTodolistBox.jsx
callemand: Yes yes :)

front/src/components/boxs/todolist/TodolistBox.jsx
atrovato: What I mean is that the refresh interval can be done on backend side, so the server fetch the API every N seconds and notify the front end with a websocket event.

callemand: Yes, It's a good idea. I try to do this :) Ty

callemand: @atrovato : I have made the change, it's ok for you ?

server/lib/todolist/todolist.closeTask.js
Pierre-Gilles: @atrovato @callemand I've read a bit the code of the integration, and I still see one downside, it seems this integration was entirely designed around the todoist API, meaning that the data model is exactly todoist data model, without any changes

What about when we want to develop another integration? We translate everything to todoist format?

front/src/components/boxs/todolist/TodolistBox.jsx
Pierre-Gilles: I don't necessarily agree on this change, because right now this integration doesn't store any data on Gladys side, so it means the backend will fetch Todoist API every minute even if the user is not on the frontend, for nothing.

**We have 2 types of integration on Gladys:**

- The weather API is not storing any data, and we fetch it every 30 minutes **from the frontend**, so the API is only queried when necessary
- The calendar API stores data in Gladys local DB (for scenes for example), and refresh this every 30 minutes **backend side** so the DB is always up to date

This integration is very nice but it lacks of general reflexion on what it's trying to achieve. Is it more like the weather API or the calendar API ?

Pierre-Gilles:
@callemand @atrovato I've wrote 2 comments on this topic. I like the integration functionally, but I think there needs to be a general reflexion on to-do lists in Gladys, outside of this Todoist integration.

What are we trying to achieve? Do we need to store to-do list in Gladys DB (or not) ? What about to-do lists in scenes ?

How is this going to work when we develop a second integration? Could we define a "Gladys data model" for to-do list in Gladys ?

All this work was done for other "generic integration" ( weather, messages, camera, devices, ...), I know it's a bit of work, but it's necessary otherwise Gladys would be a mess today ^^

PR #1683, opened 2023-01-09
Scenes - decimals in the "continue only if" block

### LINKED ISSUE #1585: Use of decimals in the "continue only if" block
This issue is related in this post :
https://community.gladysassistant.com/t/bloc-continuer-seulement-si-a-decimal/7471/3?u=spenceur

VonOx:
FIX #1585 

image

### DIFF
--- a/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
+++ b/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
@@ -24,12 +24,7 @@ class Condition extends Component {
   };
 
   handleValueChange = e => {
-    let newValue;
-    if (Number.isInteger(newValue)) {
-      newValue = Number.parseInt(e.target.value, 10);
-    } else {
-      newValue = Number.parseFloat(e.target.value);
-    }
+    const newValue = Number.parseFloat(e.target.value);
     const newCondition = update(this.props.condition, {
       value: {
         $set: newValue
--- a/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
+++ b/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
@@ -24,7 +24,12 @@ class Condition extends Component {
   };
 
   handleValueChange = e => {
-    const newValue = Number.isInteger(parseInt(e.target.value, 10)) ? parseInt(e.target.value, 10) : null;
+    const newValue = Number
+    if (newValue.isInteger) {
+      newValue = Number.isInteger(parseInt(e.target.value, 10)) ? parseInt(e.target.value, 10) : null;
+    } else {
+      newValue = e.target.value;
+    }
     const newCondition = update(this.props.condition, {
       value: {
         $set: newValue
--- a/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
+++ b/front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
@@ -24,11 +24,11 @@ class Condition extends Component {
   };
 
   handleValueChange = e => {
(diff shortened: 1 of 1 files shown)


front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
```diff
@@ -24,7 +24,12 @@ class Condition extends Component {
   };
 
   handleValueChange = e => {
-    const newValue = Number.isInteger(parseInt(e.target.value, 10)) ? parseInt(e.target.value, 10) : null;
+    let newValue;
+    if (Number.isInteger(newValue)) {
+      newValue = Number.parseInt(e.target.value, 10);
+    } else {
+      newValue = Number.parseFloat(e.target.value);
```
atrovato: I think you can use `parseFloat` for both integer and float.
I get the same result for integer with `Number.parseFloat(12)`

front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
```diff
@@ -24,7 +24,12 @@ class Condition extends Component {
   };
 
   handleValueChange = e => {
-    const newValue = Number.isInteger(parseInt(e.target.value, 10)) ? parseInt(e.target.value, 10) : null;
+    let newValue;
+    if (Number.isInteger(newValue)) {
```
atrovato: `newValue` is undefined here!

front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
atrovato: For more, 
`Number.parseFloat("ab") === NaN` but `Number.parseFloat("1ab") === 1`

I think we'll need a `/^\d+(\.\d+)?$/` regex to check that we only have numbers.

atrovato (changes requested):
Some comments here 😉

VonOx:
> Some comments here 😉

Thank you Alex, I will make the change.

I have also changed the event to onBlur ( lost focus) because on change with actual if statement prevent decimal input (dot). With only parsefloat I can revert to onChange, I don't know what is the best.

front/src/routes/scene/edit-scene/actions/only-continue-if/Condition.jsx
VonOx: Input type number do the job

atrovato:
onBlur forces the user to leave the field for updates to be taken into account.
So if the user uses "enter" key on keyboard without leaving the field, changes on the filed will be avoid.
@VonOx

VonOx:
@atrovato I've tested with "enter key" and focus is not lost and even if I hit save button on scene, new value is handled.
onBlur is working great here.

------------------------------------------------------

onChange

parseFloat(1.) = 1 => impossible to input decimal separator

onChange handle need to implement this kind of condition to prevent the behavior.
```js
if (e.target.value.charAt(str.length - 1) === '.') {
  return;
}
```
But In that condition we need need an input validator because if user set `1.` and nothing else , new value is not handled.

atrovato:
Do I understand everything is alright?

VonOx:
> Do I understand everything is alright?

With onBlur => Yes

atrovato (approved):
Ok :)

Pierre-Gilles (changes requested):
EDIT: Deleting my previous message, it works!! Issue on my side :)

Pierre-Gilles (approved):
Good for me, it works fine in the UI and it works fine with a real device / real scene! 

Thanks for this PR 🙏 

PS: forget my previous review, I messed up my test ^^

PR #1687, opened 2023-01-13
feat(ecowatt): Add day in box display

cicoub13:
CleanShot 2023-01-13 at 22 19 38@2x

DayJS proposes only 3 formats for day:
- d => day in number, not relevant
- dd => day on two letters
- ddd => day on three letters

With the available space, I thought that last format was the best

### DIFF
--- a/front/src/components/boxs/ecowatt/Ecowatt.jsx
+++ b/front/src/components/boxs/ecowatt/Ecowatt.jsx
@@ -117,7 +117,7 @@ class Ecowatt extends Component {
         days.push({
           day: dayjs(day.jour)
             .locale(this.props.user.language)
-            .format('LL'),
+            .format('ddd LL'),
           data: day.dvalue
         });
       });


Pierre-Gilles (approved):
I'm good to merge, but I'm ok with this approach only because this integration is used in France only! 

Creating custom dates format like that doesn't work with an international approach, because not all country put the day of the week first.

Just the english version of this is not strictly correct:

Screenshot 2023-01-16 at 10 08 28

Should be "Mon, January 16, 2023"

For future integration, what I recommend:

- Either using official formats in dayjs
- Or putting the format in the translation so it can be modified for each language

So ok for this one, but only because it's a FR-only integration!

cicoub13:
Totally agreeing, I took the fast path 😏 
There is no Localized format with Day + Month + Year without Hour https://day.js.org/docs/en/display/format#localized-formats.
I would rework in another PR to "putting the format in the translation so it can be modified for each language"

Pierre-Gilles:
Yes, no worries for this PR, it's good enough like that!

PR #1697, opened 2023-01-23
fix(front): Fix NaN dates display in Tasks list

cicoub13:
The idea is to transform created_at format to be in ISO8601 (compatible Firefox)

I took the `text` approach. To my opinion, better than using a Date library.
Seems not possible with Date object in Javascript 😞 

Still working on Chrome/Safari ✅ 

https://community.gladysassistant.com/t/affichage-en-nan-des-dates-dans-longlet-taches-en-arriere-plan/7784/12?u=cicoub13

CleanShot 2023-01-23 at 15 38 36

### DIFF
--- a/front/src/routes/settings/settings-background-jobs/index.js
+++ b/front/src/routes/settings/settings-background-jobs/index.js
@@ -8,10 +8,20 @@ import { WEBSOCKET_MESSAGE_TYPES } from '../../../../../server/utils/constants';
 const NUMBER_OF_JOBS_PER_PAGE = 15;
 
 class SettingsSystem extends Component {
+  /*
+   * Function to replace "2023-01-23 14:20:23.594 +00:00" by "2023-01-23T14:20:23.594+00:00"
+   * to be Firefox compatible (ISO8601)
+   */
+  convertGladysDateToISO8601 = gladysDate => {
+    return gladysDate.replace(' ', 'T').replace(' ', '');
+  };
+
   getJobs = async page => {
     try {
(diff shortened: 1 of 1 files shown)


front/src/routes/settings/settings-background-jobs/index.js
```diff
@@ -13,6 +13,9 @@ class SettingsSystem extends Component {
       const skip = page * NUMBER_OF_JOBS_PER_PAGE;
       const jobs = await this.props.httpClient.get(`/api/v1/job?take=${NUMBER_OF_JOBS_PER_PAGE}&skip=${skip}`);
 
+      // We replace "2023-01-23 14:20:23.594" +00:00 by "2023-01-23T14:20:23.594+00:00" to be Firefox compatible
+      jobs.forEach(job => (job.created_at = job.created_at.replace(' ', 'T').replace(' ', '')));
```
Pierre-Gilles: Good for me, but it's not enough! In this file, there are a few other functions where jobs are added/updated

Jobs are added/updated live in websockets

Pierre-Gilles (changes requested):
Some cases are not covered by this PR (Websockets insert/update)

front/src/routes/settings/settings-background-jobs/index.js
cicoub13: Updated ✅

Pierre-Gilles (approved):
Way better like this, thanks for this PR 🙏

PR #1703, opened 2023-02-01
Fix CalDAV update if sync is set

bertrandda:
### Pull Request check-list

### Description of change

If CalDAV calendar is updated while sync, do not updated sync attribute

### DIFF
--- a/server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
+++ b/server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
@@ -53,10 +53,8 @@ async function syncUserCalendars(userId) {
 
       // Else update it if sync is enable on calendar & events change
       if (gladysCalendar[0].sync && formatedCalendar.ctag !== gladysCalendar[0].ctag) {
-        await this.gladys.calendar.update(gladysCalendar[0].selector, {
-          ...formatedCalendar,
-          sync: gladysCalendar[0].sync,
-        });
+        delete formatedCalendar.sync;
+        await this.gladys.calendar.update(gladysCalendar[0].selector, formatedCalendar);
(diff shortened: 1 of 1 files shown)


server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
```diff
@@ -53,7 +53,10 @@ async function syncUserCalendars(userId) {
 
       // Else update it if sync is enable on calendar & events change
       if (gladysCalendar[0].sync && formatedCalendar.ctag !== gladysCalendar[0].ctag) {
-        await this.gladys.calendar.update(gladysCalendar[0].selector, formatedCalendar);
```
Pierre-Gilles, line 56: Why not simply:

```
delete formatedCalendar.sync; 
```
 ?

(just asking)

server/services/caldav/lib/calendar/calendar.syncUserCalendars.js
```diff
@@ -53,7 +53,10 @@ async function syncUserCalendars(userId) {
...
+        await this.gladys.calendar.update(gladysCalendar[0].selector, {
+          ...formatedCalendar,
+          sync: gladysCalendar[0].sync,
```
Pierre-Gilles: Don't you risk to still update it wrongly if it changed between the .get and the .update ?

Why not simply removing the sync attribute from the update if it's not required?

bertrandda: There is no action between get and update, so time is really, really short, I don't think change can happen. But indeed delete is a better option, more readable I changed.

Pierre-Gilles (approved):
Looks better now! Thanks for the change

PR #1750, opened 2023-04-23
Disabe large CIDR range scans

atrovato:
Default disable large range scans.

https://community.gladysassistant.com/t/detection-de-presence-via-le-reseau-local/6131/112?u=alextrovato


Pierre-Gilles (approved):
Good for me!

PR #1768, opened 2023-05-01
Use device feature for scene action of type turn on/off light/switch

rpochet:
### Pull Request check-list

- [x ] Is the linter passing? (`npm run eslint` on both front/server)
- [ x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to [the community](https://community.gladysassistant.com/) for testing before merging.

### Description of change

In scene, provide selection of device feature instance of device in 'Turn ON/OFF light/switch' actions.

image

Before:
image

New proposal:
image


front/src/routes/scene/edit-scene/actions/TurnOnOffLightParams.jsx
```diff
@@ -8,36 +8,40 @@ import { ACTIONS } from '../../../../../../server/utils/constants';
...
-      const deviceOptions = devices.map(device => ({
-        value: device.selector,
-        label: device.name
-      }));
-      await this.setState({ deviceOptions });
+      const deviceFeatureOptions = deviceFeatures
+        .flatMap(device => device.features)
+        .map(deviceFeature => ({
+          value: deviceFeature.selector,
+          label: deviceFeature.name
```
Pierre-Gilles: I agree with being able to select the feature in this scene action, but the deviceFeature.name is not necessarily good in all cases. 

Some integrations like Philips Hue never exposes the deviceFeature.name so the user doesn't even know it exists.

A mixed approach like we do on other scene actions with `getDeviceFeatureName` function would be better no ? 

Example: https://github.com/GladysAssistant/Gladys/blob/1222d3f9f1c687a1f7e30328abdef86b0c8bf1a4/front/src/routes/scene/edit-scene/actions/DeviceGetValueParams.jsx#L33

server/lib/scene/scene.actions.js
```diff
@@ -30,92 +30,56 @@ const actionsFunc = {
...
-          device,
-          DEVICE_FEATURE_CATEGORIES.LIGHT,
-          DEVICE_FEATURE_TYPES.LIGHT.BINARY,
-        );
+    await Promise.map(action.device_features, async (deviceFeatureSelector) => {
+      const deviceFeature = self.stateManager.get('deviceFeature', deviceFeatureSelector);
+      if (deviceFeature) {
+        const device = self.stateManager.get('deviceById', deviceFeature.device_id);
         await self.device.setValue(device, deviceFeature, 0);
-      } catch (e) {
```
Pierre-Gilles, line 91: Why removes all the catch ? the goal of the catch here was:

- If a light doesn't respond, we still want to shut down the other lamps, it's weird to stop the scene for this?

Pierre-Gilles (changes requested):
Thanks for the PR, it works fine on my side ! great work 👏 

You need to write a DB migration to update all existing scenes, otherwise all the scene already working will stop working for existing Gladys users.

I wrote other feedbacks in the code.

rpochet:
> You need to write a DB migration to update all existing scenes, otherwise all the scene already working will stop working for existing Gladys users.

How can I test DB migration?

Pierre-Gilles:
@rpochet Currently we don't have any ways to unit test data migration

I bet you could write a unit test importing a migration file and then calling the up function if you want to ? :)

Pierre-Gilles:
@rpochet Currently we don't have any ways to unit test data migration

I bet you could write a unit test importing a migration file and then calling the up function if you want to ? :)

Pierre-Gilles:
@rpochet Currently we don't have any ways to unit test data migration

I bet you could write a unit test importing a migration file and then calling the up function if you want to ? :)

PR #1798, opened 2023-05-21
Components as pure function

atrovato:
Removes 6 on 132 warnings on front side.

```
Component should be written as a pure function  react/prefer-stateless-function
```

Cloudflare link: https://components-as-pure-function.gladys.pages.dev/


front/src/routes/forgot-password/index.js
```diff
@@ -1,12 +1,68 @@
...
 class ForgotPassword extends Component {
-  render(props, {}) {
-    return <ForgotPasswordPage {...props} />;
+  forgotPassword = async e => {
+    const { forgotPasswordEmail } = this.state;
+    if (e) {
+      e.preventDefault();
+    }
+
+    if (!validateEmail(forgotPasswordEmail)) {
```
Pierre-Gilles, line 16: Forgot password crash here after changes

Pierre-Gilles (changes requested):
Nice! Tested the PR and the forgot password page is broken

front/src/routes/forgot-password/index.js
atrovato: Fixed, I changed
```javascript
import { validateEmail } from '../../utils/validateEmail';
```
to 
```javascript
import validateEmail from '../../utils/validateEmail';
```

Pierre-Gilles (approved):
Fixed, it works now! Thanks for the fixes 🙏

PR #1822, opened 2023-06-27
Fix update devices list into devices widget

callemand:
### Pull Request check-list

### Description of change

Fix update devices list into devices widget when drag and drop

### DIFF
--- a/front/src/components/boxs/device-in-room/EditDevices.jsx
+++ b/front/src/components/boxs/device-in-room/EditDevices.jsx
@@ -71,6 +71,10 @@ class EditDevices extends Component {
     this.getDeviceFeatures();
   }
 
+  componentWillReceiveProps() {
+    this.getDeviceFeatures();
+  }
+
   render(props, { selectedDeviceFeaturesOptions, deviceOptions, loading }) {
     return (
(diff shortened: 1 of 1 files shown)


front/src/components/boxs/device-in-room/EditDevices.jsx
```diff
@@ -83,7 +83,15 @@ class EditDevices extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.refreshSelectedOptions(nextProps);
+    if (nextProps.box && nextProps.box.device_features) {
+      if (
+        !this.props.box ||
+        !this.props.box.device_features ||
+        nextProps.box.device_features.length !== this.props.box.device_features.length
```
Pierre-Gilles: Adding a check on the length will not fix the issue, if I have 2 widgets with exactly the same number of deviceFeatures but different one, this will not be triggered and it should be

Pierre-Gilles (changes requested):

Pierre-Gilles (approved):
Good for me, I integrated the same changes in this PR: 

https://github.com/GladysAssistant/Gladys/pull/1838

PR #1831, opened 2023-07-06
fix(front): Increase chat height

cicoub13:
I'm bad at CSS so trying this

CleanShot 2023-07-06 at 20 16 45@2x


Pierre-Gilles:
@cicoub13 Ok, then let me know when it's ready for review :) Don't forget to try with all screen size with dev tools !

Pierre-Gilles (approved):
Tested, and it's indeed better!

PR #1894, opened 2023-09-22
Device new state event: better null check

rpochet:
### Pull Request check-list

- [ x] If your changes affects code, did your write the tests?
- [ x] Are tests passing? (`npm test` on both front/server)
- [ x] Is the linter passing? (`npm run eslint` on both front/server)
- [ x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([french forum](https://community.gladysassistant.com/)/[english forum](https://en-community.gladysassistant.com/)) for testing before merging.

### Description of change

Move check on device feature not available


server/test/lib/device/device.newStateEvent.test.js
```diff
@@ -121,4 +121,13 @@ describe('Device.newStateEvent', () => {
...
     expect(newDeviceFeature.last_value_changed).to.deep.equal(dateInTheFuture);
   });
+  it('should not save state missing device feature', async () => {
+    const stateManager = new StateManager(event);
+    const device = new Device(event, {}, stateManager, {}, {}, {}, job);
+    await device.newStateEvent({
+      device_feature_external_id: 'hue:binary:1',
+      state: 12,
+      created_at: '2019-02-12 07:49:07.556 +00:00',
+    });
```
Pierre-Gilles: Can you check that the state was indeed not saved?

Pierre-Gilles (changes requested):
Thanks for the PR ! 🙏 

Small feedback on the unit test, if you could add at least a check because right now the test doesn't test anything? :)

server/test/lib/device/device.newStateEvent.test.js
rpochet: Ok, I had a look, by the way the  catch block is hiding an issue due to DB not available during test (original: [Error: SQLITE_ERROR: no such table: t_device_feature]). I provide new check (I also change flow for newStringState)

server/lib/device/device.saveStringState.js
```diff
@@ -29,12 +36,6 @@ async function saveStringState(device, deviceFeature, newValue) {
     },
   );
 
-  deviceFeature.last_value_string = newValue;
-  deviceFeature.last_value_changed = new Date();
-
-  // save local state in RAM
-  this.stateManager.setState('deviceFeature', deviceFeature.selector, deviceFeature);
```
Pierre-Gilles, line 36: I don't think we want to change the order this is done, the goal is to update the state in RAM only after the DB is updated

Pierre-Gilles (changes requested):
Thanks for the PR! One feedback:

server/test/lib/device/device.newStateEvent.test.js
Pierre-Gilles: Ok weird, do you have this issue locally or on CI ? The table should be ready when the tests starts

server/lib/device/device.saveStringState.js
rpochet: Because this is how it is done in saveState and saveHistoricalState

server/test/lib/device/device.newStateEvent.test.js
rpochet: Probably only locally...

server/lib/device/device.saveStringState.js
Pierre-Gilles: Oh indeed, you're right! Good for me then

Pierre-Gilles (approved):
Good for me! Thanks for this PR 👏

PR #1897, opened 2023-09-24
fix typo mistake on openweather integration description (EN)

quentinlegay:
I have fixed a typo mistake on openweather integration description (EN language only)


Pierre-Gilles (approved):
Thanks! Good for me

PR #1899, opened 2023-09-25
Fix display mode widget weather

quentinlegay:
Since this change (https://community.gladysassistant.com/t/openweather-meteo-ne-fonctionne-plus/7501/38) the display mode is not correct, I have fixed this.


Pierre-Gilles (approved):
Thanks for the fixes!

PR #1900, opened 2023-09-26
Add tag on scene

callemand:
### Pull Request check-list

### Description of change
Screenshot 2023-09-26 at 08 32 48
Screenshot 2023-09-25 at 11 38 03


front/package.json
```diff
@@ -76,6 +76,7 @@
     "react-dnd-touch-backend": "^16.0.1",
     "react-select": "^4.3.1",
     "react-slider": "^2.0.6",
+    "react-tag-input-component": "^2.0.2",
```
Pierre-Gilles: Why use this library instead of react-select that we already use?

https://react-select.com/

The "multi" option + "creatable" seems to do the same?

Screenshot 2023-09-28 at 10 00 26

(Maybe the UX of the one you picked is better, but let's see if it's worth having a secondary lib?)

front/src/routes/scene/edit-scene/Settings.jsx
```diff
@@ -0,0 +1,93 @@
...
+          />
+          <h4 class="text-center card-title ">
+            <Text id="editScene.settings" />
+          </h4>
+          <div class="card-options" />
+        </div>
+        <div
+          class="card-body"
+          style={{
+            display: cardOpened ? '' : 'none'
```
Pierre-Gilles: Could we have a little CSS animation when opening the body?

Something like: 

```
/* Initial style for the hidden div */
.your-div {
  display: none;
  opacity: 0;
  height: 0;
  overflow: hidden;
  transition: opacity 0.3s ease, height 0.3s ease;
}

/* Style when the div is open */
.your-div.open {
  display: block;
  opacity: 1;
  height: auto;
}
```

front/src/routes/scene/edit-scene/Settings.jsx
```diff
@@ -0,0 +1,93 @@
...
+            })}
+          />
+          <h4 class="text-center card-title ">
+            <Text id="editScene.settings" />
+          </h4>
+          <div class="card-options" />
+        </div>
+        <div
+          class="card-body"
+          style={{
```
Pierre-Gilles: Use classes instead of direct style. I'm surprised ESLint is not saying something?

front/src/routes/scene/edit-scene/Settings.jsx
```diff
@@ -0,0 +1,93 @@
+import cx from 'classnames';
+import { Localizer, Text } from 'preact-i18n';
+import { useState } from 'preact/compat';
```
Pierre-Gilles: Except when a lib requires it, we don't use hooks in Gladys frontend for now. 

I'm open to discussing it (I'm not in favor of hooks in general), but it's probably best not to mix code style in the project for now?

server/migrations/20230627154012-add-tag-scene.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('t_scene', 'tags', {
```
Pierre-Gilles: I thought we said we were doing it the relational way?

With this option, it'll hard (if not impossible) to fetch existing tags in SQL

Pierre-Gilles (changes requested):
Thanks for the PR ! It works great 👏 

I have a few feedbacks, the major one is on the DB side. 

I really think we should stay relational in Gladys and store this as a relation table : 

```
t_tag_scene:
- scene_id uuid
- name text
```

It'll let you find existing tags pretty easily so you can display the existing tags in scene (otherwise, right now, the user can easily create duplicate tags)

On the UX, we should have a way to filter easily on the list of scene by tags (here again, you'll need a query to get all tags -> only possible in an efficient way with a relation table)

The general new UX of scene is really great and IMO users will love it ! 

Thanks for looking into this 😄

callemand:
> Thanks for the PR ! It works great 👏
> 
> I have a few feedbacks, the major one is on the DB side.
> 
> I really think we should stay relational in Gladys and store this as a relation table :
> 
> ```
> t_tag_scene:
> - scene_id uuid
> - name text
> ```
> 
> It'll let you find existing tags pretty easily so you can display the existing tags in scene (otherwise, right now, the user can easily create duplicate tags)
> 
> On the UX, we should have a way to filter easily on the list of scene by tags (here again, you'll need a query to get all tags -> only possible in an efficient way with a relation table)
> 
> The general new UX of scene is really great and IMO users will love it !
> 
> Thanks for looking into this 😄

I've try to implements the 't_tag_scene'. it's working but. Do I need to create a new api controller to get the tags ?
I need to get all the tags to display into the "react-select".

Pierre-Gilles:
>  Do I need to create a new api controller to get the tags ?

@callemand You can do it in the "scene" controller :)

server/migrations/20230627154012-add-tag-scene.js
callemand: Fixed

front/src/routes/scene/edit-scene/Settings.jsx
callemand: Fixed

front/src/routes/scene/edit-scene/Settings.jsx
callemand: Fixed

front/src/routes/scene/edit-scene/Settings.jsx
callemand: Added

server/lib/scene/scene.get.js
```diff
@@ -26,31 +27,57 @@ async function get(options) {
...
+        attributes: ['name'],
+      },
+    ],
   };
 
   if (optionsWithDefault.take !== undefined) {
     queryParams.limit = optionsWithDefault.take;
   }
 
+  // search by device feature selectors
```
Pierre-Gilles: wrong comment

server/lib/scene/scene.get.js
```diff
@@ -26,31 +27,57 @@ async function get(options) {
...
       [Op.like]: `%${optionsWithDefault.search}%`,
     });
   }
 
-  // search by device feature selectors
-  if (optionsWithDefault.selectors) {
+  if (optionsWithDefault.searchTags) {
+    const databaseTags = await Promise.all(
+      optionsWithDefault.searchTags.split(',').map((tag) => {
+        return db.TagScene.findAll({
```
Pierre-Gilles: You are doing one query per tag selected, it's not very efficient and can be done with a simple join in the scene.get query

server/lib/scene/scene.get.js
```diff
@@ -26,31 +27,57 @@ async function get(options) {
...
+        });
+      }),
+    );
+    const sceneIds = databaseTags.map((tags) => tags.map((tag) => tag.get({ plain: true }).scene_id));
+    const intersec = intersection(...sceneIds);
+
     queryParams.where = {
-      [Op.or]: optionsWithDefault.selectors.split(',').map((selector) => ({
-        selector,
+      [Op.or]: intersec.map((sceneId) => ({
```
Pierre-Gilles: This override all previous search params, so in the UI, I typed a search + added a tag, and it doesn't take into account the text search

server/lib/scene/scene.update.js
```diff
@@ -26,6 +34,21 @@ async function update(selector, scene) {
...
   await existingScene.update(scene);
 
+  await db.TagScene.destroy({
+    where: {
+      scene_id: existingScene.id,
+    },
+  });
+
+  if (scene.tags) {
+    await Promise.each(scene.tags, (tag) => {
```
Pierre-Gilles: I think you could use bulkCreate for that

https://sequelize.org/docs/v6/core-concepts/model-querying-basics/

server/api/controllers/scene.controller.js
```diff
@@ -93,6 +93,16 @@ module.exports = function SceneController(gladys) {
     res.json(scene);
   }
 
+  /**
+   * @api {get} /api/v1/tag_scene get
+   * @apiName get
+   * @apiGroup TagScene
+   */
+  async function getTag(req, res) {
+    const scenes = await gladys.scene.getTag();
```
Pierre-Gilles: const tags

Pierre-Gilles (changes requested):
Thanks for this PR ! The UI is really great like that 👏 

**A few feedbacks on the UI side:** 

- When I click outside of the tag select dropdown, it could be great to close the dropdown
- The current behavior when I select multiple tags is to do a "AND", and when comparing with other software, I have the feeling that usually it's more a "OR" done on that. What do you think of switching to OR ?

https://github.com/GladysAssistant/Gladys/assets/7365207/849fc214-abf1-49b8-a42f-525a435440f0

I was able to create a super long tag and then it break the UI :D 

Could we limit the tag length ?

Screenshot 2023-11-06 at 10 24 09

I have a few comments on the backend side (see comments in code)

Thanks for this PR 🙌

server/package.json
```diff
@@ -96,6 +96,7 @@
     "joi": "^17.7.0",
     "jsonwebtoken": "^8.5.1",
     "lodash.clonedeep": "^4.5.0",
+    "lodash.intersection": "^4.4.0",
```
Pierre-Gilles: I don't think this deps is needed anymore?

Pierre-Gilles (changes requested):
Thanks for the changes, it's way better! 👏

A few feedbacks:

When I click on a tag, it's a bit weird I can only click on the "title" but on the full width so when titles are a bit longer it's weird to have to change where to click:

https://github.com/GladysAssistant/Gladys/assets/7365207/22036bc0-8677-45f2-a1f0-85b0f38f05c8

It seems were are looking "scene density" on the scene list.

Before: 

Screenshot 2023-11-10 at 14 20 35

After:

Screenshot 2023-11-10 at 14 14 22

What about going back to 4 per width?

Pierre-Gilles (changes requested):
Thanks for the fixes!

For the click on the tags, it seems it's the same as before?

https://github.com/GladysAssistant/Gladys/assets/7365207/8d5b65a7-03c2-4286-9de6-725e856f9c35

For the responsive, maybe on mobile we should hide the "Nouveau" text and just keep the "+" (like we do on some pages) ?

Screenshot 2023-11-13 at 10 56 54

And I'm wondering if we don't need a button "unselect all" at the top ?

Screenshot 2023-11-13 at 11 01 28

front/cypress/e2e/routes/scene/Scene.cy.js
```diff
@@ -28,13 +29,14 @@ describe('Scene view', () => {
...
     });
+     */
   });
   beforeEach(() => {
     cy.login();
   });
   after(() => {
     // Delete all Bluetooth devices
-    cy.deleteDevices('mqtt');
+    //cy.deleteDevices('mqtt');
```
Pierre-Gilles: Is it normal that it's commented?

front/cypress/e2e/routes/scene/Scene.cy.js
```diff
@@ -28,13 +29,14 @@ describe('Scene view', () => {
       };
       cy.createDevice(device, 'mqtt');
     });
+     */
```
Pierre-Gilles: Is it normal that it's commented?

Pierre-Gilles (changes requested):
All good for me! I just have one question on the cypress tests :) 

And Eslint is not passing but it's just one line

After that I'm good to merge!

Edit: Little regression on the "new scene" view: 

Before:

Screenshot 2023-11-16 at 10 10 55


After: 

Screenshot 2023-11-16 at 10 10 49

front/cypress/e2e/routes/scene/Scene.cy.js
callemand: Not needed now. I have moved this device creation into the good test to avoid some issue

Pierre-Gilles (approved):
Good for me now! 👏

PR #1903, opened 2023-09-28
feat(zigbee2mqtt): Add z2m ui link in Setup page

### LINKED ISSUE #1893: Zigbee2mqtt: Add link to Zigbee2Mqtt instance in UI
[cce66](https://community.gladysassistant.com/u/cce66)
[10h](https://community.gladysassistant.com/t/zigbee2mqtt-capteur-desactive/7156/8?u=cce66)

oui c’est cela http://adresse IP:8080/#/
[@pierre-gilles](https://community.gladysassistant.com/u/pierre-gilles)
ce serait bien en terme d’ergonomie et de convivialité d’ajouter le lien http://adresse IP:8080/#/
comme l’a fait [@Lokkye](https://community.gladysassistant.com/u/lokkye) pour l’accès Node-Red (il y a des paramètres plus complets) car on se souvient pas forcement du comment y acceder surtout pendant l’été :partying_face: :wine_glass:

image

cicoub13:
Fix #1893

### DIFF
--- a/front/src/routes/integration/all/zigbee2mqtt/setup-page/actions.js
+++ b/front/src/routes/integration/all/zigbee2mqtt/setup-page/actions.js
@@ -3,6 +3,7 @@ import { RequestStatus } from '../../../../../utils/consts';
 import dayjs from 'dayjs';
 import relativeTime from 'dayjs/plugin/relativeTime';
 import get from 'get-value';
+import config from '../../../../../config';
 
 dayjs.extend(relativeTime);
 
@@ -38,6 +39,16 @@ const createActions = store => {
           networkModeValid: zigbee2mqttStatus.networkModeValid
         });
       }
+
+      // Compute Zigbee2Mqtt interface URL if connected locally (not through Gladys Plus Gateway)
+      let z2mUrl = null;
+      if (state.session.gatewayClient === undefined) {
+        const url = new URL(config.localApiUrl);
+        z2mUrl = `${url.protocol}//${url.hostname}:8080`;
+      }
+      store.setState({
+        z2mUrl
+      });
     },
 
     async startContainer(state) {
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -562,7 +562,7 @@
         "saving": "Saving USB Dongle configuration..."
       },
       "setup": {
-        "title": "Zigbe2mqtt configuration",
+        "title": "Zigbee2mqtt configuration",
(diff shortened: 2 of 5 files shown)


cicoub13:
CleanShot 2023-09-28 at 15 53 01

message is shown only when service is activated

atrovato:
What will be the URL if I'm not on localhost? Using a raspberry ou a NAS hosting Gladys? Or via Gladys Plus?

cicoub13:
> Using a raspberry ou a NAS hosting Gladys?

It will be something like http://192.168.1.11:8080 with device IP

> Or via Gladys Plus?

Link will not be showned (because not accessible from Internet)

atrovato:
I just check your code, it will use the "localApiUrl" configuration variable to define de base URL, which means that default "localhost" will be used for most of Gladys user.
I didn't set the LAN IP of my NAS as a docker variable (LOCAL_API_URL).

cicoub13:
I took code from @callemand about node-red https://github.com/GladysAssistant/Gladys/blob/a10436b10a62ba697dec05daf3b01ce0e884b53a/front/src/routes/integration/all/node-red/setup-page/SetupTab.jsx#L28

But let me test it in real world with a docker image 🤔

atrovato:
Maybe you should build the entire url from the front side, according to the `window.location.hostname`, and concat with configured the port.

cicoub13:
It's working (I don't have `LOCAL_API_URL` in Docker envs)
Port is fixed in Gladys to 8080. If you change it, you are a power user so you know where is the UI 😉 

CleanShot 2023-10-03 at 16 28 46

atrovato (approved):
Good for my tired eyes ;)

front/src/routes/integration/all/zigbee2mqtt/setup-page/actions.js
```diff
@@ -38,6 +39,16 @@ const createActions = store => {
           networkModeValid: zigbee2mqttStatus.networkModeValid
         });
       }
+
+      // Compute Zigbee2Mqtt interface URL if connected locally (not through Gladys Plus Gateway)
+      let z2mUrl = null;
+      if (this.session.gatewayClient === undefined) {
```
Pierre-Gilles: I don't think it's working! this.session doesn't contain the session object, does it?

Shouldn't it be `state.session`, like we do line 27 of this file to get the state.httpClient ?

Pierre-Gilles (changes requested):
I'm not 100% sure the Gladys Plus part is working!

front/src/routes/integration/all/zigbee2mqtt/setup-page/actions.js
cicoub13: Fixed

Pierre-Gilles (approved):
Thanks for the changes 🙏  Good for me !

PR #1908, opened 2023-10-04
fix some typo/trad mistakes on Gladys App

quentinlegay:
I've fixed some typos/trad mistakes on Gladys overall :
- Trad mistake "Watt heure" on z2m integration (English language)
- Lan manager title more clearer than others
- In French langague, when you click on "Paramètres", the title of the window is "Réglages": it's not very coherent: I've fixed it.
- A save button was misspelled if you take the other buttons as an example.
- A few typing/translation errors have also been fixed in the MQTT integration.

### DIFF
--- a/front/src/config/i18n/en.json
+++ b/front/src/config/i18n/en.json
@@ -1083,7 +1083,7 @@
         "discoveringError": "An error occurred, check scanner configuration, or check logs."
       },
       "setup": {
-        "title": "LANManager settings",
+        "title": "LAN Settings",
         "errorLabel": "An error occurred while loading configuration.",
         "noConfigLabel": "Configuration not loaded, please retry.",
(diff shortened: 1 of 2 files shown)


front/src/config/i18n/en.json
```diff
@@ -1083,7 +1083,7 @@
         "discoveringError": "An error occurred, check scanner configuration, or check logs."
       },
       "setup": {
-        "title": "LANManager settings",
```
Pierre-Gilles, line 1086: LANManager is the name of the integration, are we sure we don't keep the name of the integration here?

quentinlegay: Yes, compared with the other tabs in this integration and with what's translated into French, I found it more coherent. 
Or we could rename the tabs LAN Manager Discover, LAN Manager Devices, LAN Manager settings... (and the same in French language)
As you prefer

Pierre-Gilles: Ok 👍

Pierre-Gilles (approved):

PR #1911, opened 2023-10-13
Run scene when receive message on MQTT

callemand:
### Pull Request check-list

### Description of change

Launch scene when receive MQTT message on channel "gladys/scene/[sceneSelector]" 
The message must contain "START'


server/lib/scene/index.js
```diff
@@ -45,6 +47,7 @@ const SceneManager = function SceneManager(
   this.http = http;
   this.gateway = gateway;
   this.brain = brain;
+  this.service = service;
```
Pierre-Gilles: I'm not sure we want to inject services into the "scene" instance.

I think we should keep the same flow as the current MQTT integration:

- Only modify the MQTT integration
- Add a listener in the MQTT integration, and then emit an event (to start a scene)

Pierre-Gilles (changes requested):
As said on the forum, I think the topic should be: 

```
gladys/master/scene/[sceneSelector]/start
```

server/lib/scene/index.js
callemand: You are right, I have modified my code to use `handleGladysMessage()`

Pierre-Gilles (approved):
Way simpler like this, and it works perfectly on my side 👌

PR #1965, opened 2023-12-04
fix(docker): Node as first process

### LINKED ISSUE #1964: NodeJS doesn't take in account Docker SIGTERM
When we `docker stop` gladys container, the SIGTERM is not taken in account and we don't see these lines
```
index.js:25 (shutdown) SIGTERM received.
index.js:31 (shutdown) Closing database connection.
```

node is not the first process and so doesn't receive the SIGTERM

`ps aux`
```
PID   USER     TIME  COMMAND
    1 root      0:02 npm run start:prod
   19 root      0:00 node /src/server/node_modules/.bin/cross-env NODE_ENV=production node index.js
   26 root      0:26 node index.js
   67 root      0:00 /bin/sh
   78 root      0:00 ps aux
```

cicoub13:
Fixes #1964

New list of processes
```
PID   USER     TIME  COMMAND
    1 root      0:42 node index.js
   58 root      0:00 /bin/sh
   67 root      0:00 ps aux
```

When I `docker stop`, I've got these 2 lines 🎉 
```
2023-12-04T21:45:17+0100 <info> index.js:25 (shutdown) SIGTERM received.
2023-12-04T21:45:17+0100 <info> index.js:31 (shutdown) Closing database connection.
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,4 +35,4 @@ ENV SERVER_PORT 80
 # Export listening port
 EXPOSE 80
 
-CMD ["npm" ,"run", "start:prod"]
+CMD ["node", "index.js"]
--- a/docker/Dockerfile.buildx
+++ b/docker/Dockerfile.buildx
@@ -45,4 +45,4 @@ ENV SERVER_PORT 80
 # Export listening port
 EXPOSE 80
 
-CMD ["npm" ,"run", "start:prod"]
+CMD ["node", "index.js"]


Pierre-Gilles:
@cicoub13 are you sure it's the same as currently ?

```
cross-env NODE_ENV=production node index.js
```

Is the env production?

cicoub13:
Yes, the environment is set at this line https://github.com/GladysAssistant/Gladys/pull/1965/files#diff-80fa8164cacdc7ec0c0365e84f8154d16efdd8dd60b8b5d7d81bfb01824f6f31L42
I've tested on Raspberry Pi 4 (armv8) the new image and it's working.
I can ask people on the forum to be more sure.

And if @VonOx can have a look :)

Pierre-Gilles:
@cicoub13 Ok good for me if you tested it!

Pierre-Gilles (approved):

PR #1992, opened 2023-12-23
Fix eslint react/jsx-no-bind

atrovato:
Fixes 16 eslint warning: `react/jsx-no-bind`.


Pierre-Gilles (commented):
Are you sure it works? (tested in real life?)

If there is no bind, the function needs to be:

- A class attribute function
- Or the bind needs to be done in the constructor

atrovato:
You're right.
I tested first, on demo mode (as I can't test with real devices), it was failing.
Now I set method as class attributes, it works.
Thanks @Pierre-Gilles 

But I feel a bit sad that warnings are not checked on development phase...

Pierre-Gilles (approved):

PR #1993, opened 2023-12-28
Fix typos

quentinlegay:
Hi, I've fixed a few typos!


front/src/config/i18n/fr.json
```diff
@@ -30,17 +30,17 @@
...
+    "week": "Semaine",
+    "day": "Jour",
     "agenda": "Agenda",
-    "date": "date",
-    "time": "heure",
-    "event": "événement",
-    "noEventsInRange": "Aucuns évènements trouvés dans cette période.",
+    "date": "Date",
+    "time": "Heure",
+    "event": "Événement",
```
Pierre-Gilles, line 42: Are you sure it makes sense to change those buttons to capitalize?

Screenshot 2024-01-05 at 15 43 03

quentinlegay: They are capitalized in English and I think it looks cleaner, but that's just my opinion :)

Pierre-Gilles (approved):

PR #2025, opened 2024-02-08
fix(scene): Tag placeholder

cicoub13:
Following community feedbacks https://community.gladysassistant.com/t/changement-du-texte-pour-la-liste-des-tags-ambiguite/8716

Fixes done:
- correct display of the placeholder
- reword the text to express the fact that User "can creat or choose"

image

image


Pierre-Gilles (approved):
Tested, it works fine! Thanks for the PR 🙏

PR #2038, opened 2024-03-12
ZwaveJS-UI: Add support for multilevel switch

MathieuAndrade:
### Pull Request check-list

### Description of change

Add support for multilevel switch 

image

image

### DIFF
--- a/server/services/zwavejs-ui/lib/constants.js
+++ b/server/services/zwavejs-ui/lib/constants.js
@@ -98,8 +98,8 @@ const EXPOSES = {
   },
   multilevel_switch: {
     currentvalue: {
-      category: DEVICE_FEATURE_CATEGORIES.LIGHT,
-      type: DEVICE_FEATURE_TYPES.LIGHT.BRIGHTNESS,
+      category: DEVICE_FEATURE_CATEGORIES.SWITCH,
+      type: DEVICE_FEATURE_TYPES.SWITCH.DIMMER,
       min: 0,
       max: 99,
       keep_history: true,
--- a/server/services/zwavejs-ui/lib/constants.js
+++ b/server/services/zwavejs-ui/lib/constants.js
@@ -33,6 +33,9 @@ const STATES = {
       }
     },
   },
+  multilevel_switch: {
+    currentvalue: (val) => val,
+  },
   multilevel_sensor: {
     air_temperature: (val) => val,
     power: (val) => val,
@@ -73,6 +76,12 @@ const COMMANDS = {
       },
     },
   },
+  multilevel_switch: {
+    currentvalue: {
+      getName: (_nodeFeature) => 'set',
+      getArgs: (value, _nodeFeature) => [value],
+    },
+  },
 };
 
 const EXPOSES = {
@@ -82,6 +91,17 @@ const EXPOSES = {
       type: DEVICE_FEATURE_TYPES.SWITCH.BINARY,
       min: 0,
       max: 1,
+      keep_history: true,
+      read_only: false,
+      has_feedback: true,
+    },
+  },
+  multilevel_switch: {
+    currentvalue: {
+      category: DEVICE_FEATURE_CATEGORIES.LIGHT,
+      type: DEVICE_FEATURE_TYPES.LIGHT.BRIGHTNESS,
+      min: 0,
+      max: 99,
       keep_history: true,
       read_only: false,
       has_feedback: true,
--- a/server/test/services/zwavejs-ui/lib/zwaveJSUI.onNodeValueUpdated.test.js
+++ b/server/test/services/zwavejs-ui/lib/zwaveJSUI.onNodeValueUpdated.test.js
@@ -114,6 +114,7 @@ describe('zwaveJSUIHandler.onNodeValueUpdated', () => {
       state: 0,
     });
   });
+
   it('should save a new closed value', async () => {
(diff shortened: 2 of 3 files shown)


William-De71:
I have device with the same feature multilevel switch but in my case  it's not a lighting ! So are you sure this is the good category?
For example my devices are used to drive electrical heater and shutter.

MathieuAndrade:
Which category should we use instead? Gladys doesn't have a "generic" category for this device type that you can use in various cases, and the service doesn't seem designed to manually set the device type either. 🤔

William-De71:
I also started working on the integration and chose the switch.dimmer type.
I think it's the most 'generic' type you can use? What do you think? 
Because I don't want to see my shutters or heaters as lights

image

Pierre-Gilles:
@MathieuAndrade Thanks for the PR 🙌

Indeed, is there any way in ZwaveJS UI MQTT payloads to differentiate devices? How do you know if it's a light or a rolling shutter?

MathieuAndrade:
Hello !

@William-De71 If this category works as same as _LIGHT.BRIGHTNESS_ why not, if i have time this week a tried to check this.

@Pierre-Gilles There is no difference for ZwaveJSUI, a multilevel is a multilevel, you can plug in a light, a rolling shutter, a fan, a motor or anything that works with power variation. The must would be to let the user tell to gladys what kind of device is plugged.

PS: Sorry for the late response

William-De71:
@MathieuAndrade , Maybe my solution isn't the best solution, but in my opinion it's the most generic feature in Gladys that could work. 

@Pierre-Gilles , @MathieuAndrade is right, a multilevel is a multilevel, in my case the only way to differentiate devices is to read the "productDescription" 
Below an example of my shutter and one of my heater:
image

image

Pierre-Gilles:
@MathieuAndrade @William-De71 

> There is no difference for ZwaveJSUI

Shit, sure there is no way? It'll pose issues for basic lights, if they are detected as power plugs, it'll be really annoying in Gladys... Same for rolling shutters...

We need to find a way, even if it's a hacky one.

MathieuAndrade:
@William-De71 I have switched to _SWITCH.DIMMER_ and it works as expected, i can push it if is good for @Pierre-Gilles 

@Pierre-Gilles If you want, you can read product information (manufacturer, id, description and type) to know what type device is, but this adds a useless complexity in my opinion. And like I said above, the user can plugin anything that works with power management on this type of Zwave device so it won't work in many cases,

For exemple: [A multilevel roller shutter](https://opensmarthouse.org/zwavedatabase/116), [A multilevel RGBW controller](https://opensmarthouse.org/zwavedatabase/131), [A multilevel dimmer](https://opensmarthouse.org/zwavedatabase/133)

If you read the device's description:

- for rolling shutter -> `The device allows for controlling motors of roller blinds, awnings, venetian blinds, gates and others, which are single phase AC powered.`
- for RGBW controller -> `Fibaro RGBW Controller uses PWM output signal, which enables it to control LED, RGB, RGBW strips, halogen lights and fans.`
- for dimmer ->`Dimmer 2 remotely controlled light dimming module is designed to work with various types of light sources.`

So : 
- for rolling shutter: You can control anything that works with a motor.
- for RGBW controller: You can control any led strip but also a simple light or a fan (fun fact it can be used to read sensors data also 😂)
- for dimmer: This is the only Zwave device that you can be sure has a simple light plugged in.

These examples are only for Fibaro devices, but all Zwave devices are not same, and all manufacturer can implement functionalities differently in his device. So, there is no real method to know how a user use his Zwave device, otherwise, ZwaveJS would have already implemented it. :information_desk_person:

In my opinion, the generic type proposed by @William-De71 (_SWITCH.DIMMER_) is the best solution until the user can manualy tell to Gladys what kind of device is plugged on his Zwave device.

Pierre-Gilles:
@MathieuAndrade Ok let's do it then :)

sescandell:
Hello,

I also worked on the Multilevel Switch integration. Unfortunatelly, I saw your PR only when I first started to push my integration on Github. PR is [available here](https://github.com/GladysAssistant/Gladys/pull/2061).

> There is no difference for ZwaveJSUI, a multilevel is a multilevel, you can plug in a light, a rolling shutter, a fan, a motor or anything that works with power variation. The must would be to let the user tell to gladys what kind of device is plugged.

You are totally right for a multilevel switch device such as... "classical" switches. But, zWave provides some context about the device class. The property is called `deviceClass`. In the case of multilevel switch for curtains, usually, devices let us know they are for managing curtains.

That's one thing handled in my proposition. Thanks to that, we can nativelly map to COVER devices or DIMMER ones. There is no way for dimmers to know if there is a light or anything else at the end. But, for curtains, we can.

Another thing managed in my PR proposition is that one zWave property could be mapped to multiple "Gladys feature". This provide the capacity for dimmers to have two kind of actionners: dimmer (you can set any value from 0 to 99) or binary state ; and for multilevel curtains switch: position (you can set any value from 0 to 99) or managing OPEN|STOP|CLOSE. That way we can provide different ways to interact with the device. Example here:
image


Let me know what you think about this proposition

Pierre-Gilles:
Thanks @sescandell for the PR 🙏 

I agree with your PR, if there are no feedback from @William-De71 and @MathieuAndrade before Monday 6, I'll deploy your PR :)

Pierre-Gilles:
Merged in https://github.com/GladysAssistant/Gladys/pull/2061

PR #2127, opened 2024-10-07
fix: Update device feature label input placeholder

Terdious:
### Pull Request check-list

### Description of change
Fix: When you want to change the name of a feature in the charts and you completely delete the input field, it fills again with the original name of the feature. It's annoying!
https://community.gladysassistant.com/t/gladys-assistant-4-46-les-tableaux-de-bords/9112/3?u=terdious
https://community.gladysassistant.com/t/gladys-assistant-4-46-les-tableaux-de-bords/9112/5?u=terdious

and other Fix: when multiple devices present, deleting a device replaced the name of the next device


Pierre-Gilles (changes requested):
Hi ! Thanks for this PR 🙏

If I remove the label then save, I get an error : 

https://github.com/user-attachments/assets/4b647a84-8e2d-428f-9d20-518cd823726a

Terdious:
> If I remove the label then save, I get an error :

@Pierre-Gilles, tell me if this suits you, I took the following part:
- When you empty the field, it turns red with a shadow to warn.
image

- If you leave it like this and you no longer focus, it remains in a red frame
image

- If you save, it keeps the last name saved but authorizes the saving.
This avoids specific error management, and no longer breaks the saving. But there is less information. Does this suit you anyway?

Otherwise I see for a specific error management that I had started before, but we agree that in this case it must be managed in `front/src/routes/dashboard/edit-dashboard/index.js`? To be sure ^^

Pierre-Gilles (approved):
Very nice, works fine on my side! 👏

PR #2158, opened 2024-11-05
Fix Sensitive Data Exposure

bnematzadeh:
### Description
While reviewing the code, I encountered two vulnerabilities: one is the lack of access to call the /v1/user endpoint, and the other is the disclosure of sensitive user information, such as their encrypted passwords, etc. In `user.controller.js`, there is a function defined as `getUsers`, and according to the following code in `routes.js`, we see that it is possible to call this endpoint through a user who is simply logged into the system:
```
'get /api/v1/user': {
authenticated: true,
controller: userController.getUsers,
}
```
In `getUsers`, various parameters can be received through the query string. I came across an interesting parameter named `fields`, which allows us to set multiple values using commas in the query string and send them. The value of `fields `is stored in the `queryParams `variable within the get function and is included in the query. The problem here is that, unlike other application routes that remove the password and other sensitive fields before sending the response, there is no such code in this case to prevent this, and we can read sensitive user fields from the server. 
For the proof of concept, I wrote a security test. Before running it, ensure that the request is sent from a user who is not an admin. 
test

### DIFF
--- a/server/test/security/user.test.js
+++ b/server/test/security/user.test.js
@@ -1,7 +1,6 @@
 const { expect } = require('chai');
-const { request, authenticatedRequest } = require('../controllers/request.test');
+const { authenticatedRequest } = require('../controllers/request.test');
 
-// updated
 describe('/api/v1/user/', () => {
   it('should return all users without password', async () => {
     await authenticatedRequest
--- a/server/test/security/user.test.js
+++ b/server/test/security/user.test.js
@@ -1,6 +1,7 @@
 const { expect } = require('chai');
 const { request, authenticatedRequest } = require('../controllers/request.test');
 
+// updated
 describe('/api/v1/user/', () => {
   it('should return all users without password', async () => {
     await authenticatedRequest
--- a/server/test/security/user.test.js
+++ b/server/test/security/user.test.js
@@ -1,0 +1,15 @@
+const { expect } = require('chai');
+const { request, authenticatedRequest } = require('../controllers/request.test');
+
+describe('/api/v1/user/', () => {
+  it('should return all users with password - regular user', async () => {
+    await authenticatedRequest
+      .get('/api/v1/user?fields=password')
+      .expect('Content-Type', /json/)
+      .expect(200)
+      .then((res)=>{
+        console.log(res.body)
+      })
+   });
+})
+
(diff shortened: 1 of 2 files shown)


server/test/security/user.test.js
```diff
@@ -0,0 +1,15 @@
...
+const { request, authenticatedRequest } = require('../controllers/request.test');
+
+describe('/api/v1/user/', () => {
+  it('should return all users with password - regular user', async () => {
+    await authenticatedRequest
+      .get('/api/v1/user?fields=password')
+      .expect('Content-Type', /json/)
+      .expect(200)
+      .then((res)=>{
+        console.log(res.body)
```
Pierre-Gilles: Can you remove this log and add an "expect" statement to check that no password is returned?

server/test/security/user.test.js
```diff
@@ -0,0 +1,15 @@
+const { expect } = require('chai');
+const { request, authenticatedRequest } = require('../controllers/request.test');
+
+describe('/api/v1/user/', () => {
+  it('should return all users with password - regular user', async () => {
```
Pierre-Gilles: The title should be "should return all users without password"

Pierre-Gilles (changes requested):
Thanks for this PR 🙏
I wrote 2 small comments on the unit test 🙂

server/test/security/user.test.js
```diff
@@ -0,0 +1,15 @@
...
+const { request, authenticatedRequest } = require('../controllers/request.test');
+
+describe('/api/v1/user/', () => {
+  it('should return all users without password', async () => {
+    await authenticatedRequest
+      .get('/api/v1/user?fields=password')
+      .expect('Content-Type', /json/)
+      .expect(200)
+      .then((res)=>{
+        expect(res.body).to.not.have.key('password')
```
Pierre-Gilles: I don't think this work as res.body is a list.

You should probably do:

```
 res.body.forEach((user) => {
    expect(user).to.not.have.property('password');
  });
```

Pierre-Gilles (changes requested):
It seems you forgot to run prettier

Pierre-Gilles:
@bnematzadeh There is an issue with eslint in your file, let me know if you need help fixing it

bnematzadeh:
> @bnematzadeh There is an issue with eslint in your file, let me know if you need help fixing it

Prettier was run on this code before the commit, but I don’t know what the issue is. Can you help me?

Pierre-Gilles:
Screenshot 2024-11-08 at 14 52 06

Pierre-Gilles (approved):
Looks good to me! Thanks for this PR! :)

PR #2169, opened 2024-11-10
Add RangeValidator for take paramater

bnematzadeh:
### Description

In the `user.get.js ` file, no validator has been implemented for the take field. A client could send a request to the following endpoint and set takes to a large value, which could lead to reduced performance and a potential DoS attack on the application. I have added a new validator called RangeError.

In all controllers, it’s better to check that the value provided by the client does not exceed a certain range before passing this parameter to the database. For this purpose, I wrote two different tests: one for when the takes value is out of range, and another for when it is within the valid range.

```
if (optionsWithDefault.take) {
    if (optionsWithDefault.take > DEFAULT_OPTIONS.limit) {
      throw new RangeError('The value of the limit parameter is out of range');
    }
    queryParams.limit = optionsWithDefault.take;
  }
```

### DIFF
--- a/server/lib/user/user.get.js
+++ b/server/lib/user/user.get.js
@@ -1,11 +1,13 @@
 const { Op } = require('sequelize');
 const Sequelize = require('sequelize');
+const { RangeError } = require('../../utils/coreErrors');
 const db = require('../../models');
 
 const DEFAULT_OPTIONS = {
   fields: ['id', 'firstname', 'lastname', 'selector', 'email'],
   expand: [],
   skip: 0,
+  limit: 20,
   order_dir: 'ASC',
   order_by: 'firstname',
 };
@@ -40,6 +42,9 @@ async function get(options) {
   };
 
   if (optionsWithDefault.take) {
+    if (optionsWithDefault.take > DEFAULT_OPTIONS.limit) {
+      throw new RangeError('The value of the limit parameter is out of range');
+    }
     queryParams.limit = optionsWithDefault.take;
   }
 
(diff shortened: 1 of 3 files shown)


Pierre-Gilles:
@bnematzadeh Why not, but it's a bit of cherrypicking in the case of Gladys :) Don't forget that Gladys is a self-hosted/local software, and when you talk about the "client", it's the same person that started Gladys server, so if the "client" is asking for a big `?take`, the user could ddos... his own server! :D

bnematzadeh:
> @bnematzadeh Why not, but it's a bit of cherrypicking in the case of Gladys :) Don't forget that Gladys is a self-hosted/local software, and when you talk about the "client", it's the same person that started Gladys server, so if the "client" is asking for a big `?take`, the user could ddos... his own server! :D

You're right. In any case, this can also be added to the code to ensure the parameter is sent within a specific range. Overall, I agree with you.

Pierre-Gilles:
> In any case, this can also be added to the code to ensure the parameter is sent within a specific range. 

Yes, but I don't think it's worth it to put it everywhere. 

For example, the "GET user" should return only a few rows, as most people have from 1 to 5 users at home (depending on how many people are living in this home), so I don't think this route is a good candidate.

Routes with millions of rows are better candidates, but I still think that in some case it could be useful to query millions of rows.

For example, the other day, I wanted to know the minimum temperature recorded during this summer on my outside temperature sensor, so I asked for the last 200k rows, and I received them without any issues (it was 4.6Mb, retrieved in 500ms). 

If there was a low limit, I couldn't have done it

Pierre-Gilles:
@bnematzadeh So what so you think?

Pierre-Gilles:
I'm closing this, don't hesitate if you have any other feedback :)

PR #2171, opened 2024-11-19
Fix RCE in gateway.downloadBackup

bnematzadeh:
### Description

There is a Remote Code Execution vulnerability in the gateway controller that allows a logged-in user to execute arbitrary commands on the service. The vulnerable sink is located in the `gateway.downloadBackup `file. The `fileUrl `value corresponds to the file_url field in the `/api/v1/gateway/backup/restore` endpoint.

In the gateway.downloadBackup file, the `GLADYS_GATEWAY_BACKUP_KEY `value is first checked, and if it is not set, an error is thrown (this can be triggered by calling the `POST /api/v1/variable/GLADYS_GATEWAY_BACKUP_KEY` endpoint). To decrypt the backup, the `exec `function is used. However, the value of encryptedBackupFilePath is used in exec without being sanitized.

Before executing exec, the server sends a request to the `fileUrl`. If the link is problematic, an Axios error is returned. If a query string is included in the URL, its value is ignored. The issue here is that a malicious link containing an encrypted file and a hash value with an arbitrary command can be crafted. When this hash value is used in exec, the command gets executed.

Below, I have attached a sample POC and the command I executed. To fix this issue, the # character must also be stripped from the fileWithoutSignedParams.


Pierre-Gilles (commented):
Thanks for this PR! I'm not sure this could ever be exploited as the user is hosting the server himself, but it's still a good thing to fix this :) 

Why not use an URL validator to make sure the URL is really valid?

We use Joi in Gladys:

```javascript
const Joi = require('joi');

const schema = Joi.string().uri();

// Example validation
const url = 'https://example.com';
const { error, value } = schema.validate(url);
```

Don't forget to add the tests :)

bnematzadeh:
This scenario can be exploited by a new user created by an admin by calling the mentioned endpoint. Using the JoI library you referred to, I defined a regex with the pattern method to remove the specified characters if they appear at the end of a URL and return the cleaned URL. This is because simply using the uri method in this case is not sufficient and can be bypassed with other payloads. I made some changes to coreErrors and added a new custom error for URLs, which I have used both in the gateway file and in its corresponding test file.

bnematzadeh:
Oops, I think we have a problem. With the changes I've made to the downloadBackup file, two other tests are failing because when the downloadBackup method is called with a non-URL input, it reaches the validator and causes an error. I made a change in the file where, using a function, it checks whether the input is a URL or not. If it is a URL, the validateUrl function is called on it, and if the URL is valid, the value is returned. Otherwise, we're dealing with a local file.

Pierre-Gilles:
@bnematzadeh I tried with Joi with the native uri check and it works: 

Screenshot 2024-11-28 at 13 58 25

Why not use this one? Do you have example of security issues with this ?

bnematzadeh:
> @bnematzadeh I tried with Joi with the native uri check and it works:
> 
> Screenshot 2024-11-28 at 13 58 25
> 
> Why not use this one? Do you have example of security issues with this ?

The reason I didn’t use the native URI is that it recognizes the following URL as a valid URL and doesn’t handle the rest of the fragment. A custom command can be used inside `$()` after the fragment, and if the native URI is used, this technique can be bypassed. That’s why I stripped these two characters entirely.

server/lib/gateway/gateway.downloadBackup.js
```diff
@@ -20,15 +21,15 @@ async function downloadBackup(fileUrl) {
   if (encryptKey === null) {
     throw new NotFoundError('GLADYS_GATEWAY_BACKUP_KEY_NOT_FOUND');
   }
-  // Extract file name
-  const fileWithoutSignedParams = fileUrl.split('?')[0];
+
+  const value = isURL(fileUrl) ? validateUrl(fileUrl) : fileUrl;
```
Pierre-Gilles: I'm not sure it still works like that, for context, the fileUrl looks like that (it's a dummy URL) : 

```
https://example_storage.com/b6131425-8f90-4a05-ba00-12daf37f8279.enc?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=SDFDSFDSFDSFDSFFDF%2F20241202%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20241202T110928Z&X-Amz-Expires=21600&X-Amz-Signature=6768767678HHkjhjdhfjdshfjhdksfhjdshf&X-Amz-SignedHeaders=host
```

So the get parameters should be removed before doing path.basename()

Having the variable being name `value` isn't very clear

server/utils/url.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+/**
+ * @description Validate the url.
+ * @param {string} url - The url of the backup.
+ * @returns {string} Return a valid url.
+ * @example
+ * validateUrl();
+ */
+function validateUrl(url) {
```
Pierre-Gilles, line 27: You created a generic function "validateUrl" in `utils`, but isn't this behavior very gateway.downloadBackup specific? 

Will this be used elsewhere?

Pierre-Gilles (changes requested):
Ok thanks for you answer! I have a few feedbacks on the PR

server/utils/url.js
bnematzadeh: What change should I make so that the GET parameter is removed before it’s passed to path.basename()? The validator I wrote, in my opinion, can be used in other places, but it’s largely tied to gateway.downloadBackup.

Pierre-Gilles: Using `fileWithoutSignedParams` instead ? (as it was before)

```
const fileWithoutSignedParams = fileUrl.split('?')[0];
```

Pierre-Gilles:
@bnematzadeh I’m closing this PR for now due to inactivity. Feel free to reopen it anytime if you decide to continue working on it 🙂.

Thanks again for your contributions!

PR #2187, opened 2024-12-17
feat(scenes): Add does not contain for Calendar action

cicoub13:
### Pull Request check-list

- [X] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([french forum](https://community.gladysassistant.com/)/[english forum](https://en-community.gladysassistant.com/)) for testing before merging.

### Description of change

Provides a "does not contain" in "continue only if calendar event" action

image

⚠️ If multiple events are happening now and at least one contains the word, scene will stop.

### DIFF
--- a/front/src/routes/scene/edit-scene/actions/CalendarIsEventRunning.jsx
+++ b/front/src/routes/scene/edit-scene/actions/CalendarIsEventRunning.jsx
@@ -245,6 +245,13 @@ class CheckTime extends Component {
               </Localizer>
             </div>
           )}
+          {action.calendar_event_name_comparator === 'does-not-contain' && (
+            <div className="col-md-12">
+              <p>
+                <Text id="editScene.actionsCard.calendarEventIsRunning.doesNotContainExplanation" />
+              </p>
+            </div>
+          )}
         </div>
         <div class={cx('row', style.calendarEventIsComingGroupMargin)}>
           <div class="col-sm-6">
--- a/front/src/routes/scene/edit-scene/actions/CalendarIsEventRunning.jsx
+++ b/front/src/routes/scene/edit-scene/actions/CalendarIsEventRunning.jsx
@@ -217,6 +217,9 @@ class CheckTime extends Component {
                 <option value="contains">
                   <Text id="editScene.triggersCard.calendarEventIsComing.contains" />
                 </option>
+                <option value="does-not-contain">
+                  <Text id="editScene.triggersCard.calendarEventIsComing.doesNotContain" />
+                </option>
                 <option value="starts-with">
                   <Text id="editScene.triggersCard.calendarEventIsComing.startsWith" />
                 </option>
--- a/server/lib/calendar/calendar.findCurrentlyRunningEvent.js
+++ b/server/lib/calendar/calendar.findCurrentlyRunningEvent.js
@@ -54,6 +54,12 @@ async function findCurrentlyRunningEvent(calendars, calendarEventNameComparator,
         [Op.like]: `%${calendarEventName}%`,
       };
       break;
+    case 'does-not-contain':
+      // @ts-ignore
+      queryParams.where.name = {
+        [Op.notLike]: `%${calendarEventName}%`,
+      };
+      break;
     case 'starts-with':
       // @ts-ignore
       queryParams.where.name = {
--- a/server/lib/scene/scene.actions.js
+++ b/server/lib/scene/scene.actions.js
@@ -440,12 +440,33 @@ const actionsFunc = {
     }
   },
   [ACTIONS.CALENDAR.IS_EVENT_RUNNING]: async (self, action, scope, columnIndex, rowIndex) => {
-    // find if one event match the condition
-    const events = await self.calendar.findCurrentlyRunningEvent(
-      action.calendars,
-      action.calendar_event_name_comparator,
-      action.calendar_event_name,
-    );
+    let events = [];
+    // manage special does-not-contain condition
+    if (action.calendar_event_name_comparator === 'does-not-contain') {
+      // we want events containing the word from the condition
+      const eventThatContainsEventName = await self.calendar.findCurrentlyRunningEvent(
+        action.calendars,
+        'contains',
+        action.calendar_event_name,
+      );
+      // if at least one event is found containing the word, we want to stop the scene (so events stays empty)
+      // otherwise we need to check if there is at least another event not containing the word from condition
+      if (eventThatContainsEventName.length === 0) {
+        const eventThatDoesNotContainsEventName = await self.calendar.findCurrentlyRunningEvent(
+          action.calendars,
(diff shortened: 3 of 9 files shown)


Pierre-Gilles:
@cicoub13 FYI: https://community.gladysassistant.com/t/liste-deroulante-dune-condition-incomplete/9238/7?u=pierre-gilles

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
```diff
@@ -43,6 +43,7 @@ async function findCurrentlyRunningEvent(calendars, calendarEventNameComparator,
   switch (calendarEventNameComparator) {
     // do nothing in that case
     case 'has-any-name':
+    case 'does-not-contains':
```
Pierre-Gilles: Why didn't you do the opposite of contains?

```
queryParams.where.name = {
        [Op.notLike]: `%${calendarEventName}%`,
      };
```

Pierre-Gilles (changes requested):

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
cicoub13: It was my first implementation, but the case where we have two events with:
- event 1 with calendarEventName in name
- event 2 without calendarEventName in name
will continue the scene (because at least one event 2 found) and it's not what we want.

Pierre-Gilles: Oh yes indeed, you're right 🫣

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
```diff
@@ -70,6 +71,18 @@ async function findCurrentlyRunningEvent(calendars, calendarEventNameComparator,
   // we search in the database if we find events that match our request
   const eventsMatching = await db.CalendarEvent.findAll(queryParams);
 
+  // manage special does not contain condition
+  if (calendarEventNameComparator === 'does-not-contain') {
+    const eventMatchingDoesNotContain = eventsMatching.filter((event) => !event.name.includes(calendarEventName));
+    const eventNotMatchingDoesNotContain = eventsMatching.filter((event) => event.name.includes(calendarEventName));
+    // If at least an event is found not containing the word, we want to stop the scene
```
Pierre-Gilles: The comment is not related, we are in the function to calendar.findCurrentlyRunningEvent

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
```diff
@@ -70,6 +71,18 @@ async function findCurrentlyRunningEvent(calendars, calendarEventNameComparator,
...
+    if (eventMatchingDoesNotContain.length > 0) {
+      return [];
```
Pierre-Gilles: I'm not sure of the implementation

this function is supposed to return event matching a condition

Here you are doing the job of the scene it seems, or am I wrong?

Pierre-Gilles (changes requested):

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
cicoub13: You are right. But the calendar events returned by this function are not sufficient to implement the does not contain in the scene code. I will move this part in the scene.actions.js
Do I need to create a new function or tweak this one?

Pierre-Gilles: No you can just tweak this function, but this function should do what's advertised: find events matching a condition. If this function was used at other places, it should work as well :) 

> I will move this part in the scene.actions.js

Yes!

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
cicoub13: Moved to `server/lib/scene/scene.actions.js`

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
cicoub13: Moved to `server/lib/scene/scene.actions.js`

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
```diff
@@ -41,8 +41,9 @@ async function findCurrentlyRunningEvent(calendars, calendarEventNameComparator,
   };
   // eslint-disable-next-line default-case
   switch (calendarEventNameComparator) {
-    // do nothing in that case
+    // do nothing in that cases
     case 'has-any-name':
+    case 'does-not-contains':
```
Pierre-Gilles: Still not sure why this function does the same thing for 'has-any-name' and 'does-not-contains' 

On peut parler en français si c'est plus simple pour toi pour expliquer :)

Pierre-Gilles (changes requested):

server/lib/calendar/calendar.findCurrentlyRunningEvent.js
Pierre-Gilles: Pour moi, la fonction findCurrentlyRunningEvent est indépendante des scènes, et doit donc avoir un comportement logique même si on oublie toute la partie scène

Hors là ce n'est pas le cas

server/lib/scene/scene.actions.js
```diff
@@ -440,12 +440,33 @@ const actionsFunc = {
...
+    let events = [];
+    // manage special does-not-contain condition
+    if (action.calendar_event_name_comparator === 'does-not-contain') {
+      // we want events containing the word from the condition
+      const eventThatContainsEventName = await self.calendar.findCurrentlyRunningEvent(
+        action.calendars,
+        'contains',
+        action.calendar_event_name,
+      );
+      // if at least one event is found containing the word, we want to stop the scene (so events stays empty)
```
Pierre-Gilles: > we want to stop the scene (so events stays empty)

Depends on what the user has chosen no?

front/src/routes/scene/edit-scene/actions/CalendarIsEventRunning.jsx
```diff
@@ -217,6 +217,9 @@ class CheckTime extends Component {
                 <option value="contains">
                   <Text id="editScene.triggersCard.calendarEventIsComing.contains" />
                 </option>
+                <option value="does-not-contain">
+                  <Text id="editScene.triggersCard.calendarEventIsComing.doesNotContain" />
+                </option>
```
Pierre-Gilles, line 222: As this new mode is quite tricky to understand, I would add a small paragraph below it that's displayed only when the option is selected to explain exactly what's this mode is doing

Pierre-Gilles (changes requested):

PR #2323, opened 2025-05-25
PR3 - Vehicle electrical - Add convert units imperial US / SI

Terdious:
### Pull Request check-list

### Description of change


server/utils/constants.js
```diff
@@ -1037,6 +1055,84 @@ const DEVICE_FEATURE_UNITS_BY_CATEGORY = {
...
+    DEVICE_FEATURE_UNITS.WATT_HOUR_KM,
+    DEVICE_FEATURE_UNITS.WATT_HOUR_MILE,
+    DEVICE_FEATURE_UNITS.KILOWATT_HOUR_PER_100_MILE,
+    DEVICE_FEATURE_UNITS.MILE_PER_KILOWATT_HOUR,
+    DEVICE_FEATURE_UNITS.FEET_PER_SECOND,
+    DEVICE_FEATURE_UNITS.MILE_PER_HOUR,
+  ],
+};
+
+const DISTANCE_UNIT_CONVERSIONS = {
```
Pierre-Gilles: I'm not sure the `constants` file is designed for conversion functions :D 

Can you migrate this to the units file (or to a separate file, but not in the constants file?)

Pierre-Gilles (changes requested):
Thanks for this PR :) A few feedbacks!

server/utils/constants.js
Terdious: Okay for me, I proposed a new file. The conversion constant is already quite large (and likely to grow in the future).

front/src/components/boxs/device-in-room/device-features/sensor-value/BadgeNumberDeviceValue.jsx
```diff
@@ -79,15 +80,28 @@ const BADGE_VALUE_CONVERTERS = {
...
+  let displayUnit = unit;
+
+  const isDistanceUnit = DISTANCE_UNITS.DEVICE_FEATURE_UNITS.includes(displayUnit);
+  // Conversion if necessary (and if user is present)
+  if (value !== null && isDistanceUnit) {
+    const conversion = convertUnitDistance(lastValue, unit, userUnit);
+    value = conversion.value;
+    displayUnit = conversion.unit;
+  }
   const valued = value !== -1;
```
Pierre-Gilles: `valued` should be `value === null`, otherwise here any value that's equal to -1 will display "No value"

front/src/components/boxs/device-in-room/device-features/sensor-value/BadgeNumberDeviceValue.jsx
```diff
@@ -79,15 +80,28 @@ const BADGE_VALUE_CONVERTERS = {
...
     return <RawDeviceValue {...props} />;
   }
 
-  let value = lastValue === null ? -1 : lastValue;
   let valueIsEnum = false;
+
+  let value = lastValue;
+  let displayUnit = unit;
+
+  const isDistanceUnit = DISTANCE_UNITS.DEVICE_FEATURE_UNITS.includes(displayUnit);
```
Pierre-Gilles: Just raising a thought: what if we split this into a dedicated component?

That’s what we did for the Temperature features, for example.

I’m a bit concerned that we’re adding distance-specific logic into a generic file that handles many types of devices. As we introduce more conversions or device types, it might become harder to maintain and reason about.

What’s your take on this?

front/src/components/boxs/device-in-room/device-features/sensor-value/RawDeviceValue.jsx
```diff
@@ -1,15 +1,32 @@
...
-  </div>
-);
+import { convertUnitDistance } from '../../../../../../../server/utils/units';
+import { DISTANCE_UNITS } from '../../../../../../../server/utils/constants';
+
+const RawDeviceValue = ({ deviceFeature, user }) => {
+  let value = deviceFeature.last_value;
+  let displayUnit = deviceFeature.unit;
+
+  const isDistanceUnit = DISTANCE_UNITS.DEVICE_FEATURE_UNITS.includes(displayUnit);
```
Pierre-Gilles: Same here

Pierre-Gilles (commented):
A few feedbacks :)

front/src/components/boxs/device-in-room/device-features/sensor-value/BadgeNumberDeviceValue.jsx
Terdious: Indeed !!

front/src/components/boxs/device-in-room/device-features/sensor-value/BadgeNumberDeviceValue.jsx
Terdious: Hello @Pierre-Gilles ,
I agree with your suggestion. I just thought it would be more cumbersome, especially if I added the volume, area, elevation, distance, and speed conversions later, but your solution actually seems much more readable.

The change is complete.

However, is there still the modification made to the `Gauge` component? As well as the one in the `Chart` component, which I'm currently adding...? What do you think?

server/utils/units.js
```diff
@@ -66,8 +68,61 @@ function hslToRgb(h, s, l) {
...
+ * // result = { value: 6.21, unit: 'mile' }
+ */
+function convertUnitDistance(value, fromUnit, userPreference) {
+  const conversions = DISTANCE_UNIT_CONVERSIONS[userPreference];
+  if (conversions && conversions[fromUnit]) {
+    const conversion = conversions[fromUnit];
+    // Convert the value and apply smart rounding
+    const convertedValue = smartRound(conversion.convert(value));
+    // Choose the target unit (function or string)
+    const unit = typeof conversion.unit === 'function' ? conversion.unit(value) : conversion.unit;
```
S-Axel: Why not having unit of type `function` only?
`unit-conversions.js` may look like this:
```
unit: (_value) => DEVICE_FEATURE_UNITS.WATT_HOUR_PER_MILE,
```

but the type would remain the same and would be easier to use:

```
const unit = conversion.unit(value);
```

front/src/components/boxs/chart/Chart.jsx
```diff
@@ -248,7 +249,17 @@ class Chartbox extends Component {
       } else {
         series = data.map((oneFeature, index) => {
           const oneUnit = this.props.box.units ? this.props.box.units[index] : this.props.box.unit;
-          const oneUnitTranslated = oneUnit ? this.props.intl.dictionary.deviceFeatureUnitShort[oneUnit] : null;
+          // Convert distance units if necessary
+          let displayUnit = oneUnit;
```
S-Axel: Could we do this instead?

```suggestion
          const userUnit = this.props.user.distance_unit_preference;
          // Convert unit display format (using 0 as a dummy value since we only need the unit)
          const conversion = convertUnitDistance(0, oneUnit, userUnit);
          const displayUnit = conversion.unit;

```

It is much simpler as convertUnitDistance makes all the necessary checks before converting.

_Seen with @Terdious_

front/src/components/boxs/chart/Chart.jsx
```diff
@@ -248,7 +249,17 @@ class Chartbox extends Component {
...
+          const userUnit = this.props.user.distance_unit_preference;
```
S-Axel: I find `userUnit` variable name not explicit enough. Why not `userUnitPreference`?
Because I had to read the full line to understand the real meaning of the variable. I thought it would contain a unit like "milimeter"

front/src/components/boxs/chart/Chart.jsx
S-Axel: Can be repeated below with other uses of convertUnitDistance

front/src/components/boxs/chart/Chart.jsx
S-Axel: Or maybe it is the original name that triggers a confusion : `distance_unit_preference`
Should it be `distance_unit_system_preference` instead?

S-Axel: Or even `unit_system_preference` to take into account other units than distance, according to _@Terdious_

front/src/components/boxs/gauge/GaugeBox.jsx
```diff
@@ -97,20 +98,34 @@ class GaugeBox extends Component {
...
 
   formatValueWithUnit = (value, unit) => {
+    const { user } = this.props;
+    let displayUnit = unit;
+
+    const isDistanceUnit = DISTANCE_UNITS.DEVICE_FEATURE_UNITS.includes(displayUnit);
+    // Conversion if necessary (and if user is present)
+    if (user && value !== null && isDistanceUnit) {
+      const userUnit = user.distance_unit_preference;
+      const conversion = convertUnitDistance(value, displayUnit, userUnit);
```
S-Axel: same as charts

front/src/components/boxs/device-in-room/device-features/sensor-value/DistanceSensorDeviceValue.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+const DistanceSensorDeviceValue = ({ deviceFeature, user }) => {
+  const { last_value: lastValue = null, unit } = deviceFeature;
+  const userUnit = user.distance_unit_preference;
+  let value = lastValue;
+  let displayUnit = unit;
+
+  // Convert the value to the user unit if needed
+  if (displayUnit !== userUnit) {
+    const conversion = convertUnitDistance(value, displayUnit, userUnit);
```
S-Axel: same as charts

server/utils/unit-conversions.js
S-Axel, line 1: As discussed with _@Terdious_,
`convertUnitDistance` from `units.js` could be. renamed as `checkAndConvertUnit` and handle any type of unit. 
For instance, this would handle pressure tire units that are not a distance.

Here, we could export a global constant `UNIT_CONVERSIONS`, but keep a private constant `DISTANCE_UNIT_CONVERSIONS`, and have future private constants like `PRESSURE_UNIT_CONVERSIONS`.

`UNIT_CONVERSIONS` would be built from private constants while private constants would allow us to keep to source code organized. 






For instance
```
const { DEVICE_FEATURE_UNITS, DISTANCE_UNITS } = require('./constants');

const _DISTANCE_UNIT_CONVERSIONS = {
  [SYSTEM_UNITS.US]: {
    [DEVICE_FEATURE_UNITS.MM]: { unit: DEVICE_FEATURE_UNITS.INCH, convert: (mm) => mm / 25.4 },
    },
};

// Expected in a future pr
const _PRESSURE_UNIT_CONVERSIONS = {
  [SYSTEM_UNITS.US]: {
    [DEVICE_FEATURE_UNITS.BAR]: { unit: DEVICE_FEATURE_UNITS.PSI, convert: (bar) => bar * 14.504 },
  },
}

const UNIT_CONVERSIONS = {
  [SYSTEM_UNITS.US]: {
    ..._DISTANCE_UNIT_CONVERSIONS[SYSTEM_UNITS.US],
    ..._PRESSURE_UNIT_CONVERSIONS[SYSTEM_UNITS.US],
    
  }
}

module.exports = {
  UNIT_CONVERSIONS,
};
```

S-Axel (commented):
I added some feedbacks! I hope I won't be too annoying 😌

server/utils/units.js
Pierre-Gilles: I’m not sure I agree with this approach, why make it more complex than it needs to be? As far as I can tell, we always use a simple string for the unit, not a function. So unless there’s a clear use case for using a function here, wouldn’t it make more sense to just stick with a string?

front/src/components/boxs/device-in-room/device-features/sensor-value/DistanceSensorDeviceValue.jsx
```diff
@@ -0,0 +1,26 @@
+import { Text } from 'preact-i18n';
+import { checkAndConvertUnit } from '../../../../../../../server/utils/units';
+
+const DistanceSensorDeviceValue = ({ deviceFeature, user }) => {
+  const { last_value: lastValue = null, unit } = deviceFeature;
+  const userUnitPreference = user.distance_unit_preference;
+
+  // Convert the value to the user unit if needed
+  const conversion = checkAndConvertUnit(lastValue, unit, userUnitPreference);
```
Pierre-Gilles: Little JS tricks :D 

```javascript
const { value: displayValue, unit: displayUnit } = checkAndConvertUnit(lastValue, unit, userUnitPreference);
```

front/src/components/boxs/device-in-room/device-features/sensor-value/IconBinaryDeviceValue.jsx
```diff
@@ -8,8 +8,8 @@ const ICON_MAP = {
     1: 'lock'
   },
   [DEVICE_FEATURE_CATEGORIES.ELECTRICAL_VEHICLE_STATE]: {
-    0: 'unlock',
-    1: 'lock'
+    0: 'lock',
```
Pierre-Gilles: I think we should stay consistent with how other door-type devices are handled in Gladys. If we use 0 for “unlocked” and 1 for “locked”, let’s stick to that convention 🙂
If a manufacturer’s API uses the opposite, the integration should handle the conversion accordingly.

server/utils/constants.js
```diff
@@ -1160,6 +1166,30 @@ const DEVICE_FEATURE_UNITS_BY_CATEGORY = {
   ],
 };
 
+const SYSTEM_UNITS = {
```
Pierre-Gilles: Why is this variable called SYSTEM_UNITS? The name feels unclear to me, it seems to mix several unrelated concepts, and it appears to refer specifically to distance units without actually saying so. Maybe we could choose a more descriptive name?

server/utils/units.js
```diff
@@ -66,8 +68,71 @@ function hslToRgb(h, s, l) {
...
+ * @description Converts a value from one distance unit to another according to the user's preference.
+ * @param {number} value - Value to convert.
+ * @param {string} fromUnit - Original unit (e.g. 'km', 'mile', ...).
+ * @param {string} userPreference - User preference ('us' or 'metric').
+ * @returns {{ value: number, unit: string }} Object containing the converted value and the target unit.
+ * @example
+ * const result = convertUnit(10, DEVICE_FEATURE_UNITS.KM, SYSTEM_UNITS.US);
+ * // result = { value: 6.21, unit: 'mile' }
+ */
+function checkAndConvertUnit(value, fromUnit, userPreference) {
```
Pierre-Gilles, line 108: If this function is specifically related to distance units, as the description suggests, that should be clearly reflected in its name.

If not, the description should be updated to more accurately reflect its actual purpose.

Pierre-Gilles (changes requested):

server/utils/constants.js
Terdious: Because, as specified in @s-axel's review, it will also contain the psi-to-bar (tire pressure) conversion functions/units. In the future, it will also be able to contain liters and gallons.
This seemed pretty clear to us, so Google it:
image

server/utils/units.js
S-Axel: image
 In a previous version, functions were used in some cases.... but not anymore, you're right! So indeed, we can rollback to single string as values.

server/utils/constants.js
Pierre-Gilles: In the context of Gladys, “SYSTEM_UNITS” sounds like it refers to units used by the Gladys system itself.

S-Axel: What do you think about `MEASUREMENT_UNITS`?

front/src/components/boxs/device-in-room/device-features/sensor-value/DistanceSensorDeviceValue.jsx
Terdious: Oh yes sorry, it's me ... !!

front/src/components/boxs/device-in-room/device-features/sensor-value/IconBinaryDeviceValue.jsx
Terdious: I'm making the change, but it still doesn't seem logical to me in the naming of constants, to have to modify in node-red (today) the data received by Tesla or Tessie, expressly named "door-open" and "window-open" to transform "0" into "1" and vice versa.

Besides, other manufacturers may have data named "door-close" and "window-close."
Perhaps it would be better to name these constants "door-state" and "window-state" at this point? No?

Pierre-Gilles (approved):
Looks good to me, tested locally, it works great! Great work 👏

PR #2407, opened 2025-12-27
Fix typos

quentinlegay:
Hi, 
I spotted a few typos in Gladys and took the opportunity to scan the project with GitHub Copilot free directly via GitHub and correct several additional typos.


front/src/actions/map.js
```diff
@@ -4,19 +4,19 @@ function createActions(store) {
...
         );
         store.setState({
           usersWithLocation,
-          GetUsersWithlocationStatus: RequestStatus.Success
+          getUsersWithlocationStatus: RequestStatus.Success
         });
       } catch (e) {
         store.setState({
-          GetUsersWithlocationStatus: RequestStatus.Error
+          getUsersWithlocationStatus: RequestStatus.Error
```
Pierre-Gilles: You're modifying the code here, is it the goal?

Pierre-Gilles (commented):
Thanks a lot for this PR, it's great! Just one feedback

front/src/actions/map.js
quentinlegay: Not a typo it's true, but this is to comply with the codebase’s camelCase naming convention, but I’ve realized it was still not correct, so I’ve just updated it

Pierre-Gilles: Ok, I searched the codebase to make sure this is not used elsewhere, and it's not used anywhere so we're good :) Thanks!

Pierre-Gilles (approved):

PR #2424, opened 2026-01-29
fix(chat): preserve spaces in messages by using pre-wrap

quentinlegay:
Hi,

This PR fixes the issue where multiple spaces in chat messages were being collapsed into a single space.
Previously, .msg_txt_wrap used white-space: pre-line, which removed extra spaces while keeping line breaks.

With this change, we replace it with white-space: pre-wrap to ensure that multiple consecutive spaces typed by the user are preserved in the chat.

This behavior is now consistent with Telegram, where users can use multiple spaces for formatting or notifications.

**_Before:_**
_Gladys Chat :_
image
_Telegram notification :_ 
image

**_After:_**
_Gladys Chat :_
image
_Telegram notification :_ 
image

This improves the chat UX, especially for notifications or formatting that rely on spacing.

### DIFF
--- a/front/src/routes/chat/style.css
+++ b/front/src/routes/chat/style.css
@@ -177,5 +177,5 @@
 }
 
 .msg_txt_wrap {
-  white-space: pre-line;
+  white-space: pre-wrap;
 }


Pierre-Gilles (approved):
Amazing !! Thanks for the fix !

PR #2500, opened 2026-03-22
fix: use numeric bitmap for ColorControl optionsMask

david-digitis:
## Description of Change

`ColorControl.moveToHueAndSaturation` expects `optionsMask` and `optionsOverride` as **uint8 numbers** (raw bitmap), unlike `LevelControl.moveToLevel` which accepts typed bitmap objects.

Passing `{ executeIfOff: true }` (object) causes a TLV validation crash:

```
ValidationDatatypeMismatchError/128: Expected number, got object.
  at TlvNumberSchema.validate
  at Object.ClusterClient.commands [as moveToHueAndSaturation]
  at MatterHandler.setValue (matter.setValue.js:160)
```

This breaks **all color changes** on Matter devices. On/off and brightness work fine since they use different clusters.

### Fix

| Parameter | Before (broken) | After (fixed) |
|---|---|---|
| `optionsMask` | `{ executeIfOff: true }` | `1` (bit 0 = executeIfOff) |
| `optionsOverride` | `{}` | `1` (executeIfOff = true) |
| `transitionTime` | `null` | `0` (instant) |

Test assertions updated accordingly.

### Tested with

- Matterbridge devices (Magic Home LED controllers exposed as Matter lights via [matterbridge-magic-home](https://github.com/david-digitis/matterbridge-magic-home))
- Gladys 4.58 on Docker (Synology NAS)

## Summary by CodeRabbit

* **Chores**
  * Improved light color control parameter handling for internal optimization.

### DIFF
--- a/server/services/matter/lib/matter.setValue.js
+++ b/server/services/matter/lib/matter.setValue.js
@@ -160,11 +160,9 @@ async function setValue(gladysDevice, gladysFeature, value) {
     await colorControl.moveToHueAndSaturation({
       hue: matterHue,
       saturation: matterSaturation,
-      transitionTime: null,
-      optionsMask: {
-        executeIfOff: true,
-      },
-      optionsOverride: {},
+      transitionTime: 0,
+      optionsMask: 1, // bitmap: bit 0 = executeIfOff
+      optionsOverride: 1, // bitmap: bit 0 = executeIfOff
     });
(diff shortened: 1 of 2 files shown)


Pierre-Gilles (approved):
Great PR, thanks for the fix!

PR #2528, opened 2026-05-18
core: Energy monitoring - extend destroyStatesFrom (device) and wrapper (job) for partial recalc support - PR1

Terdious:
### Pull Request check-list

- [ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging.

## Summary

> This PR is the first independent slice extracted from #2413 as discussed
> on https://community.gladysassistant.com/t/amelioration-re-calcul-suivi-de-lenergie/10069 ;
> the rest of #2413 will follow in subsequent PRs only if this one is merged.

Extends two existing core primitives with an optional parameter, in preparation for upcoming energy-monitoring features (selective recalculation, date-range recalculation). The change is intentionally **incremental** rather than introducing sibling functions: every pre-existing caller keeps the same effective behavior, no new file is added, no new public surface area is exposed.

- `device.destroyStatesFrom(selector, from, to = new Date())` — adds an optional `to` upper bound (inclusive) defaulting to `now`.
- `job.wrapper(type, func, { detached = false } = {})` — adds an optional `{ detached }` option flag. When false (default), the wrapper body is byte-for-byte identical to master; when true, the wrapper starts the job, runs `func` in the background, returns the started job immediately and never re-throws to the caller.

### `device.destroyStatesFrom(selector, from, to = new Date())`

- The existing function previously had the signature `(selector, from)`.
- The new optional `to` parameter defaults to `new Date()`, so every pre-existing caller keeps the same effective behavior: the SQL goes from `WHERE created_at >= from` to `WHERE created_at >= from AND created_at <= now`, which is identical in practice.
- When `to` is explicitly provided, the function deletes only states inside `[from, to]` (both inclusive).
- When `from > to`, the SQL is naturally empty and the function silently no-ops — documented in the JSDoc and covered by an explicit test.
- Continues to throw `NotFoundError('DeviceFeature not found')` for unknown selectors. Continues to use parameterized DuckDB DELETE (no string interpolation in SQL).

### `job.wrapper(type, func, { detached = false } = {})`

- The existing function previously had the signature `(type, func)`.
- When `detached === false` (default), the wrapper performs the exact same SQL sequence as on master (`start` → `await func` → `finish(SUCCESS)` / on error `finish(FAILED)` then re-throw) and returns the same value (`res`). The shared body is now extracted into a local IIFE (`runAndFinish`) so the legacy and detached modes share it instead of duplicating it.
- When `detached === true`, the wrapper calls `this.start(type)`, fires `runAndFinish()` without awaiting it, and returns the started `job` immediately. Errors thrown by `func` are reported on the job (status=FAILED) and never re-thrown to the caller; a secondary `try/catch` falls back to `logger.error('job.wrapper: failed to finish job ...')` if `finish(FAILED)` itself rejects.
- This is the fire-and-forget mode needed by long-running "from beginning" recalculations that would otherwise hold the HTTP request open until completion.

### Tests

- `server/test/lib/device/device.destroyStatesFrom.test.js`:
  - the 5 pre-existing master tests are unchanged,
  - 2 new tests cover the bounded form: a happy path with a range cut-out (states inside `[from, to]` deleted, states outside kept) and an inverted-range silent no-op.
- `server/test/lib/job/job.test.js`:
  - the existing `describe('job.wrapper')` legacy tests are unchanged,
  - the new `describe('job.wrapper with { detached: true }')` block carries 3 tests covering success path, failure path, and the secondary log path when `job.finish` itself rejects,
  - a small additional test on `updateProgress` asserts it rejects when the job id is unknown — easy win for coverage of a pre-existing branch.

### Coverage

- `device.destroyStatesFrom.js`: **100%** statements / lines / functions / branches.
- `job.wrapper.js`: **100%** statements / lines / functions; branch coverage at 75% (one defensive guard `if (error && error.toString)` is not exercised — identical to the pre-existing situation on master, no new uncovered branch introduced).

### Tested manually

- A long-running "from beginning" recalculation on my own production instance now returns the HTTP response immediately and runs as a background job (progress visible in the jobs page) — this is the actual functional motivation for the `{ detached }` mode and the reason this PR is the prerequisite for the follow-ups.

## Scope

Diff vs base branch (`master`):

| Area | Insertions | Deletions |
|---|---|---|
| server (production) | 67 | 13 |
| server (tests) | 148 | 6 |
| **Total** | **215** | **19** |

Files touched:

- `server/lib/device/device.destroyStatesFrom.js` — `+23 / -4` (add optional `to` parameter)
- `server/lib/job/job.wrapper.js` — `+25 / -13` (factor out shared body, add `{ detached }` option)
- `server/test/lib/device/device.destroyStatesFrom.test.js` — `+87 / 0` (2 new tests for the bounded form)
- `server/test/lib/job/job.test.js` — `+80 / -6` (3 new tests for the detached mode + 1 on `updateProgress`)

No new file, no new public function, no new prototype binding on `DeviceManager` or `Job`.

## Summary by CodeRabbit

* **New Features**
  * Jobs can now run in detached/background mode, enabling non-blocking job execution
  * Device state deletion now supports specifying an optional date range (from and to) for more precise cleanup

* **Tests**
  * Added comprehensive test coverage for detached job execution and error handling
  * Added test cases for bounded date range state deletion, including edge cases

[Review Change Stack](https://app.coderabbit.ai/change-stack/GladysAssistant/Gladys/pull/2528?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/lib/device/device.destroyStatesBetween.js
```diff
@@ -0,0 +1,32 @@
...
+ * @description Destroy states between two dates.
+ * If from is after to, no states are deleted (silent no-op).
+ * @param {string} selector - Device feature selector.
+ * @param {Date} from - The start date (inclusive).
+ * @param {Date} to - The end date (inclusive).
+ * @returns {Promise<void>}
+ * @example
+ * await gladys.device.destroyStatesBetween('kitchen-washer-consumption', new Date('2025-01-01'), new Date());
+ */
+async function destroyStatesBetween(selector, from, to) {
```
Pierre-Gilles: Why create a new function when destroyStatesFrom() already exists and could simply be extended?

The `to` parameter could default to now, making the change fully backward-compatible with the current usage of the function.

server/lib/job/job.wrapper.js
```diff
@@ -32,6 +33,46 @@ function wrapper(type, func) {
...
+/**
+ * @public
+ * @description Start a job and run it in background without awaiting completion.
+ * @param {string} type - Type of the job.
+ * @param {Function} func - The function to wrap.
+ * @returns {Function} Return function that resolves when job is started.
+ * @example
+ * gladys.job.wrapperDetached('daily-aggregation', func);
+ */
+function wrapperDetached(type, func) {
```
Pierre-Gilles: This code duplicates the logic from the wrapper function, with the only difference being that it does not resolve once finished.

What are you trying to achieve here?

I’d avoid duplicating the code, as it makes maintenance harder over time. I’m pretty sure this could be solved with a small change.

Pierre-Gilles (changes requested):
Thanks for separating the PR! I've added my feedback.

I'm not sure whether this was AI-invented or based on your own ideas, but this is typically the kind of change we should try to avoid with AI-assisted PRs: copying large blocks of code that were already working well and could have been extended more incrementally.

In this case, I think a smaller and more targeted change would have made the code easier to review and maintain, probably closer to what an experienced developer would naturally do.

Terdious:
Thanks for the review @Pierre-Gilles.

> I'm not sure whether this was AI-invented or based on your own ideas, but this is typically the kind of change we should try to avoid with AI-assisted PRs: copying large blocks of code that were already working well and could have been extended more incrementally.

On the architectural choice (new functions vs. extending existing ones), this was my own design call, not an AI suggestion. The AI executed within constraints I had set; the strategy itself was deliberate.

When I wrote the original energy PR (#2413) in January, I was operating from years of conversations with you about Gladys stability — your recurring theme that the calculation engine is sensitive and that modifications there carry validation cost. 

> In this case, I think a smaller and more targeted change would have made the code easier to review and maintain, probably closer to what an experienced developer would naturally do.
I totally agree with that and it would have been my initial choice actually. I was just afraid of getting negative feedback from you for fear of breaking the core (especially with the wrapper used in many places in the code).
Adding *new* isolated functions (`destroyStatesBetween`, 
`wrapperDetached`) rather than modifying `destroyStatesFrom` and `wrapper` 
felt like the safest path: zero impact on existing code, full backward 
compatibility by construction, and a smaller blast radius if anything 
went wrong. That was my reading of the room.

With hindsight, your suggestion is cleaner here and I agree on the merits.
I will refactor both functions along these lines and deploy the updates quickly.

Thanks again for the time on the review.

server/lib/device/device.destroyStatesBetween.js
Terdious: Agreed. I'll merge this into `destroyStatesFrom` with an optional `to` 
parameter defaulting to `now`, keeping full backward compat.

server/lib/job/job.wrapper.js
Terdious: The goal of `wrapperDetached` was to allow the HTTP response to return 
immediately while the long-running job continues in the background — 
this fixes the from-beginning timeout on large installs (the response 
currently blocks until the full recalculation finishes, which can be 
hours).

I went for a separate function to avoid touching the existing `wrapper`, 
but I agree your suggestion is cleaner. I'll refactor it as an optional 
`{ detached: true }` parameter on `wrapper` and push the update.

server/lib/device/device.destroyStatesBetween.js
Terdious: `destroyStatesBetween` is removed. `destroyStatesFrom` now accepts an optional `to` parameter that defaults to `new Date()`, so all pre-existing callers keep their effective behavior (the SQL becomes `>= from AND <= now`, identical in practice to `>= from`). 2 new tests cover the bounded form (range cut-out + inverted `from > to` no-op); the 5 existing master tests are unchanged.

server/lib/job/job.wrapper.js
Terdious: Fixed in 600f4ce0 by extracting the shared "run func + finish + handle error" body into a local IIFE (`runAndFinish`); the legacy and detached paths now only differ on the three points where they actually have to:

1. legacy awaits the inner promise, detached fires it and returns the started `job` immediately;
2. legacy re-throws `func` errors to the caller, detached reports them on the job and swallows;
3. legacy lets a `finish(FAILED)` rejection propagate (same as master), detached catches it and falls back to `logger.error`.

Legacy SQL sequence and return value are unchanged from master.

Terdious:
@Pierre-Gilles update on testing:

I've pushed a Docker test image `terdious/gladys:energy-monitoring-pr1` 
running since last night on an isolated copy of my prod DB (I dropped 
data older than 2025 to keep the run manageable).

I launched a full cost recalculation (costs only, not consumption) — 
completed in 1h30, logs are clean. I compared the recalculated values 
against my actual prod month by month and year by year: results match. 
The only delta is on yesterday's data, which is expected — it 
corresponds to the gap between the DB snapshot and the test instance 
startup.

I can let it run another day or two if you'd like more confidence on 
the long-run incremental mode, or I can move on to preparing a similar 
test image for PR2. Let me know what you prefer.

server/lib/job/job.wrapper.js
```diff
@@ -1,34 +1,54 @@
...
+ * @param {boolean} [options.detached=false] - When true, the wrapper starts the job, returns it
+ *   immediately and runs `func` in the background; errors are reported on the job (status=FAILED)
+ *   and never re-thrown. When false (default), the wrapper awaits `func` and re-throws on error.
+ * @returns {Function} Wrapped function.
  * @example
  * gladys.job.wrapper('daily-aggregation', func);
+ * gladys.job.wrapper('long-running-recalc', func, { detached: true });
  */
-function wrapper(type, func) {
+function wrapper(type, func, { detached = false } = {}) {
```
Pierre-Gilles, line 18: Je ne suis pas sûr de bien comprendre le besoin ni l’utilité de cette modification, j’avoue.

Tu mentionnes des requêtes HTTP qui bloquent, mais de mon côté il me semblait que ce problème avait déjà été corrigé. Le correctif est disponible ici : https://github.com/GladysAssistant/Gladys/blob/master/server/services/energy-monitoring/api/energy-monitoring.controller.js#L23

L’idée était justement de ne plus attendre la fin du traitement avant de répondre à l’appel API (suppression du await), afin que la requête soit résolue immédiatement.

Du coup, quel est exactement le bug que cette modification est censée résoudre ?

Pierre-Gilles: Au passage, s’il reste effectivement un bug côté interface, je préférerais avoir une PR dédiée qui corrige le problème de bout en bout, plutôt qu’une PR qui mélange plusieurs sujets différents :)

PR #2620, opened 2026-07-04
Netatmo: Discover and save modules that are unreachable in homestatus

Terdious:
### Pull Request check-list

- [x] If your changes affect the code, did you write the tests?
- [x] Are server tests passing with coverage? (`cd server && npm run coverage`) — Codecov requires **100% coverage on lines changed in this PR**
- [x] Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [x] If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)
- [x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging. *(validated on a real installation with 5 powered-off modules — see screenshot below)*
- [ ] ~~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~ *(no API change)*
- [ ] ~~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~ *(no new feature)*
- [ ] ~~Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~~ *(not needed)*

## Summary

Netatmo reports powered-off modules in the `/homestatus` **`errors`** array (`{ code: 6, id }`) instead of the `modules` array. The discovery flow only iterated `homestatus.modules`, so these devices were later re-added from the legacy `getthermostatsdata` response as raw objects **without `name` and without `room`**. Two user-facing consequences:

1. A device literally named **"undefined"** in the discovery page.
2. A hard crash of the whole discovery page: `TypeError: Cannot read properties of undefined (reading 'reachable')` in `NetatmoDeviceBox.isDeviceReachable` — the page freezes and requires a full reload. It only takes one powered-off module in the home (e.g. a heat-pump relay switched off for the season).
image

## Details

**Server**
[...]

PR #2618, opened 2026-07-04
Netatmo: Harden connection lifecycle and surface setValue failures

Terdious:
### Pull Request check-list

- [x] If your changes affect the code, did you write the tests?
- [x] Are server tests passing with coverage? (`cd server && npm run coverage`) — Codecov requires **100% coverage on lines changed in this PR**
- [ ] ~~Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)~~ *(no UI change in this PR)*
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [ ] ~~If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)~~ *(no i18n change in this PR)*
- [x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging. *(validated on a real installation with real Netatmo devices)*
- [ ] ~~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~ *(no API change)*
- [ ] ~~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~ *(no new feature)*
- [ ] ~~Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~~ *(not needed)*

> **Merge order:** please merge #2620 first — its discovery-page fix is required to test this PR on a real installation that has powered-off modules. This branch already includes #2620.

## Summary

Five small robustness fixes on the Netatmo service, following up on the resilience work from #2561/#2562. No new feature, no schema change.

## Details

1. **`setValue` no longer fails silently (or crashes).** The `catch` block accessed `e.response.status`, a leftover from the axios era: with undici, a network timeout has no `response`, so the handler crashed with a `TypeError`. And in every failure case the error was swallowed, so the user believed their setpoint was applied. The error is now logged with context and re-thrown so callers get proper feedback. An HTTP error response also throws (after the existing status/websocket reporting) instead of returning silently.
[...]

PR #2617, opened 2026-07-04
Netatmo: Fix zero values being dropped and stop emitting absent states

Terdious:
### Pull Request check-list

- [x] If your changes affect the code, did you write the tests?
- [x] Are server tests passing with coverage? (`cd server && npm run coverage`) — Codecov requires **100% coverage on lines changed in this PR**
- [ ] ~~Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)~~ *(no UI change in this PR)*
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [ ] ~~If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)~~ *(no i18n change in this PR)*
- [x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging. *(validated on a real installation with real Netatmo devices)*
- [ ] ~~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~ *(no API change)*
- [ ] ~~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~~ *(no new feature)*
- [ ] ~~Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~~ *(not needed)*

> **Merge order:** please merge #2620 first — its discovery-page fix is required to test this PR on a real installation that has powered-off modules. This branch already includes #2620.

## Summary

This PR fixes a long-standing data-accuracy bug in the Netatmo weather/energy module updates: legitimate zero values were treated as "missing" because value selection used the `||` operator.

Concretely, with the previous code:
- a rain gauge reporting **0 mm** (the nominal state most of the time),
- a thermometer at **0°C**,
- wind or gust coming from the **north (0°)**,
- a wind sensor reporting **0 km/h**

would either fall back to the (potentially stale/different) `dashboard_data` value, or emit a `NEW_STATE` event with an `undefined` state when no fallback was available.

## Details

[...]

PR #2627, opened 2026-07-05
[WIP] Netatmo: Receive webhook events via Gladys Plus for Energy devices

Terdious:
### Pull Request check-list

- [ ] ~~Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)~~ *(no E2E scenario covers the Netatmo setup page)*
- [x] If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)
- [ ] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging. *(runtime validation in progress on a real installation with a real Gladys Plus account — the full chain is proven end-to-end: `addwebhook` accepted, real Netatmo events relayed by the gateway and received by the service)*
- [ ] ~~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~~ *(no REST API change; the existing `/configuration` endpoint simply carries one more key)*
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com). *(a short "Netatmo instant events via Gladys Plus" section will be proposed once this PR is validated)*
- [ ] ~~Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~~ *(not needed)*

> **Stacked PR:** based on #2619 and #2618 (which already include #2617 and #2620). Please merge those first — this PR then reduces to its last commits. The branch also embeds the small front fix #2628 (found while runtime-testing this PR), so real-life testing includes it; its 10 lines are counted in its own PR, not in the Scope below.

## Summary

Netatmo can push events over webhooks, but a local Gladys instance is not reachable from the internet. @pierre-gilles wired a cloud route for exactly this back in 2021 ([forum topic](https://community.gladysassistant.com/t/webhook-gladys-plus/6071/5)): `POST https://api.gladysgateway.com/v1/api/netatmo/:open_api_key` relays the raw Netatmo webhook body to the primary instance over the gateway websocket as a `gladys-open-api` message with the `netatmo-webhook` action. The route is still live in [gladys-gateway](https://github.com/GladysAssistant/gladys-gateway) — but nothing consumed that action on the instance side until now.

With this PR, setpoint and heating-mode changes made **outside** Gladys (Netatmo app, physical knob on a valve/thermostat, schedule transitions) are reflected in Gladys within seconds of Netatmo pushing them, instead of waiting for the next 2-minute poll. Scope is **Energy devices only** for now: the weather station emits no webhook events at all (polling remains its only source), and Security/camera events are intentionally ignored — they will be handled in a follow-up PR of the camera series.

## How it works

1. **Core:** `gateway.handleNewMessage.js` gets a `netatmo-webhook` branch — the exact Owntracks precedent — emitting the new `EVENTS.GATEWAY.NEW_MESSAGE_NETATMO_WEBHOOK` event and answering `200` immediately, before any processing (Netatmo bans webhook URLs that fail to answer quickly). The open-api dispatch is also grouped under a single `gladys-open-api` if/else chain (one type check, mutually exclusive actions made explicit) — no behavior change. ⚠️ Worth knowing for the cloud side: the gateway answers the third party's HTTP request only after the instance acks the relayed message (socket.io ack without timeout), so a webhook event reaching an instance that does not handle the action (any Gladys without this PR) or no instance at all leaves/fails the cloud request — with Netatmo, five failures in a row ban the app's webhook (observed in real life). The durable fix belongs to gladys-gateway's `handleNetatmoWebhook`: answer `200` in every case, with a timeout on the instance round-trip.
2. **Configuration:** the user creates an Open API key in Gladys Plus ("Settings" → "Open API") and pastes **just the key** in the Netatmo setup tab — the fixed beginning of the webhook URL is displayed as an input-group prefix (masked key input with an eye toggle; pasting a full URL also works and is stored as-is). Stored as the `NETATMO_WEBHOOK_URL` service variable. **An empty field disables the feature entirely — zero behavior change for users without Gladys Plus.**
3. **Registration:** `registerWebhook` calls Netatmo's `addwebhook` fire-and-forget on each successful connection (init with stored tokens, OAuth callback) and on configuration change. It is best-effort: a failure is logged and never affects the connection status, and re-registering on every connection recovers from a Netatmo-side drop or ban. `dropWebhook` is called when the key is cleared and on disconnect (while the access token is still valid). No need to fill the "webhook URI" field of the Netatmo developer portal: the integration manages the registration itself.
4. **Event handling:** the service subscribes to the gateway event in `init()` (remove-then-on so repeated starts never double-subscribe) and unsubscribes in `disconnect()`. `handleWebhookEvent` dispatches on `event_type`:
   - Energy events trigger a devices values refresh through the **existing** single-flight `refreshNetatmoValues` (#2618), debounced by 2 s to coalesce bursts. The whitelist contains `setpoint_event` (push_type `display_change`) — **the name observed in real life** for a setpoint change — plus the names documented by the Home Assistant integration (`set_point`, `cancel_set_point`, `therm_mode`, `schedule`).
   - Events are also processed while a poll or a discovery is running (those flows temporarily overwrite the connected status; caught in real-life testing).
   - `webhook_activation` confirms the registration; anything else (camera events like `movement`/`human`, unknown future types) is ignored at debug level with its `push_type` — the handler never throws (`eventFunctionWrapper` as the safety belt).

Design choice: the refresh-based approach reuses the whole homesdata/homestatus pipeline, so all states stay consistent. Events missed while the instance is offline are caught up by the poll, which remains the source of truth.

## Real-life behavior of Netatmo Energy webhook events (observed, worth knowing)

Documented from real payloads captured during testing — they show that the event **must only be used as a trigger, never as data**, which is why this PR deliberately refreshes through `homestatus` instead of applying the payload directly (the Home-Assistant-style direct apply would display wrong values in every case below):

- `setpoint_event` carries `temperature: 0` when `mode` is `home` (= "follow the schedule, no manual override"): the field only holds the *override* setpoint, and the effective scheduled setpoint only exists in `homestatus`. Same for `therm_setpoint_end_time: 0` (= no override window).
- `device_id` is the **relay** (NAPlug), not the valve/thermostat, and the event reflects the **commanded** state: the relay acknowledges a setpoint order even when the actual thermostat is offline behind it.
- Latency is erratic and entirely Netatmo-side: ~5 s observed for a return-to-schedule, ~30 s for a home `therm_mode` change, ~1 min for a manual setpoint change — occasionally slower than the 2-minute poll. Mode transitions are chained (setting "away" emitted `hg → schedule` then `schedule → away`) and events can arrive **out of order**: a stale `hg → schedule` was delivered a full minute *after* the `schedule → away` event. A direct payload apply would have reverted the away mode to a stale state — the trigger-a-refresh design is immune to duplicates, delays and reordering (debounce + single-flight guard + idempotent refresh). Camera events, in contrast, arrive within seconds (relevant for the follow-up camera-events PR, which reuses this exact plumbing).

## Scope

Diff computed against the parent state of the stack (#2619 merged with #2618), excluding the embedded #2628:

| Area | Insertions | Deletions |
| --- | ---: | ---: |
| Core production (`server/lib/gateway`, `server/utils/constants.js`) | 18 | 24 |
| Core tests (`server/test/lib/gateway`) | 74 | 0 |
| Netatmo service (`server/services/netatmo`) | 200 | 4 |
| Netatmo tests (`server/test/services/netatmo`) | 510 | 2 |
| Front (`front/src/routes/integration/all/netatmo`) | 106 | 2 |
| i18n (en/fr/de) | 15 | 0 |
| **Total** | **923** | **32** |

256 tests passing (netatmo suite + gateway `handleNewMessage`), 100% coverage on all touched files, prettier/eslint clean on server and front, `compare-translations` OK.

## Summary by CodeRabbit

* **New Features**
  * Added Netatmo webhook support with optional webhook key setup, instant-event explanation, and improved validation for a `https://` webhook URL.
  * Enables faster Netatmo updates via webhook-triggered refresh.
* **Bug Fixes**
  * Improved Netatmo discovery warnings by showing device-specific error messages (including a dedicated message for error code 6) and more accurate reachability handling.
  * Improved Netatmo device conversion for more reliable names/labels and handling of unreachable/offline devices.
  * Preserved existing webhook settings when reopening the setup screen.


Terdious:
Backport external integration [#Terdious/gladys-netatmo](https://github.com/Terdious/gladys-netatmo)

Terdious:
Closing — work moves to the external integration.

The core Netatmo service is being retired in favor of [`Terdious/gladys-netatmo`](https://github.com/Terdious/gladys-netatmo), so webhook support is being rebuilt there rather than in core, tracked in Terdious/gladys-netatmo#5.

⚠️ Note this is the one item **not yet shipped**: the gateway → external-integration **container** relay leg does not exist yet (the core only forwards `netatmo-webhook` to the internal service bus). Candidate design to work out with @Pierre-Gilles: a generic `external-integration.webhook` WS message exposed by the SDK as `onWebhook`. The anti-replay front fix (#2628) is already covered container-side by the single-use OAuth state.

Disposition & parity map: Terdious/gladys-netatmo#15.

---

PR #2679, opened 2026-07-17
Free Mobile: add multi-user support so each user can send/receive SMS

William-De71:
### Pull Request check-list

- [ ] Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)
- [ ] If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)
- [x] Did you test this pull request in real life? With real devices?
- [ ] If your changes modify the API (REST or Node.js), did you modify the API documentation?
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation?
- [x] Did you add fake requests data for the demo mode? (N/A — no new widget)

### Description of change

This PR adds **multi-user support to the Free Mobile integration**.

#### Problem

Until now, only the **admin** account could configure a Free Mobile API key. The customer ID and access token were stored as **global** service variables, so Gladys could only send SMS to that single account. There was no way to send an SMS to another user, nor to build scenes targeting a specific user.

#### Solution

Free Mobile is now aligned with the **Telegram / CallMeBot** pattern:

- **Per-user configuration**: the `FREE_MOBILE_USERNAME` and `FREE_MOBILE_ACCESS_TOKEN` variables are now stored per-user (`userRelated: true`). Each user can enter their own credentials from the integration page to receive SMS.
- **Automatic migration**: on service start, the legacy global configuration (if any) is migrated to the **admin** user's per-user configuration, then the old global variables are removed. Existing installations keep working with no manual action.
- **New message handler**: a `message.send(userId, message)` handler reads the user's own credentials on the fly and sends the SMS.
- **Scene action**: the "Send an SMS" action now has a **recipient selector** — you can send to a specific user, or to **all users** who configured Free Mobile.

## Summary by CodeRabbit

- **New Features**
  - Free Mobile credentials are now configured per user.
  - Scene SMS actions let you select a recipient (including “all users”).
  - The Free Mobile settings page now shows guidance for per-user SMS setup.
  - SMS sending now uses the scene text with variables expanded.

- **Bug Fixes**
  - The integration loads even when legacy or incomplete Free Mobile credentials are missing.
  - Legacy Free Mobile configuration is migrated automatically to per-user settings.
  - SMS sending skips recipients missing required credentials and continues for others.

- **Tests**
  - Expanded coverage for Free Mobile SMS sending and migration behavior.

### DIFF
--- a/front/src/config/i18n/de.json
+++ b/front/src/config/i18n/de.json
@@ -2580,7 +2580,7 @@
       },
       "smsSend": {
         "userLabel": "Empfänger der SMS",
-        "allUsersOption": "Alle Benutzer",
+        "allUsersOption": "Alle Free-Mobile-Benutzer",
         "textLabel": "Nachricht",
         "explanationText": "Um eine Variable in den Text einzufügen, gib \"{{\" ein. Um einen Variablenwert festzulegen, musst du zuerst das Feld \"Gerätewert abrufen\" verwenden.",
         "messagePlaceholder": "Meine Message"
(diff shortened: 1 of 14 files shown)


Pierre-Gilles:
I think we can close this in favor of the external integration, no? It seems like the better long-term approach.

PR #2714, opened 2026-07-27
External integration store: add a catalog refresh control and lower the cache TTL to 30 minutes

Pierre-Gilles:
### Pull Request check-list

- [x] Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)
- [x] If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)
- [ ] Did you test this pull request in real life? With real devices?
- [ ] If you are adding a new features/services which needs explanation, did you modify the user documentation?
- [ ] Did you add fake requests data for the demo mode (`front/src/config/demo.js`)?

Notes on the unchecked items:

- **Real devices**: not applicable, nothing here touches a device. The UI was exercised in a real browser against the demo dev server (desktop 1400px and mobile 390px, all three outcome states).
- **User documentation**: `docs/specs/external-integrations.md` is updated for the new cadence. Nothing added to the v4-website user docs — say the word if the refresh control deserves a mention there.
- **Demo mode**: `front/src/config/demo.js` has no fixtures for the external integration routes (already the case on `master`, so the catalog is empty on the demo site). The refresh control is now visible to the demo admin and clicking it shows the error state. Happy to add the three fixtures if you want the demo to behave.

### Description of change

Two related changes to the external integration store.

**1. The catalog could lag half a day behind the store.** The public indexer rebuilds `index.json` hourly, but Gladys kept its in-memory copy for 12 hours and only refreshed on that same period. `STORE_INDEX_TTL_MS` goes to 30 minutes; it drives both the freshness check in `getIndex()` and the periodic refresh interval set up in `init()`, so the catalog now stays within one indexer cycle of upstream. The persistent `t_variable` cache and the offline fallback are unchanged.

**2. `POST /api/v1/external_integration/store/refresh` existed but was never called from the front.** An admin had no way to pick up an integration published since the last refresh short of restarting Gladys. The catalog now ends with a discreet muted line — *"An integration missing from the list? ↻ Refresh catalog"*. It sits at the very end of the list on purpose: it is a rare gesture that does not belong in the options row, and that position also puts it directly under the "no result" message when a search for a freshly published integration comes back empty — the moment someone actually wants it. Admin-only, like the catalog itself and like the route.

**API change:** `POST /store/refresh` now returns an extra `refreshed: true|false` alongside `refreshed_at` and `integrations`.

This exists because an unreachable store is not an error: `getIndex()` swallows the failure and serves the cached index, so the endpoint answers `200` with a stale catalog and the UI would otherwise announce a refresh that never happened. A new `refreshCatalog()` on the supervisor captures `storeIndexFetchedAt` around the call — which only moves on a successful download — and reports the outcome. The front shows three distinct states: catalog up to date, store unreachable so this is still the cache, or the request itself failed.

An earlier revision inferred this client-side by comparing `refreshed_at` against the previously held value. That was wrong when the initial catalog `GET` had failed: the baseline was `null`, and any non-null cached timestamp then looked like a successful download. The server-side comparison happens within a single call, so there is no baseline to get wrong.

### Test Plan

Server: the TTL fresh/stale boundary, and the `refreshed` flag for a successful download, an unreachable store and no index at all — at both the supervisor and controller level. They stub `refreshIndex` rather than the HTTP layer, so they assert the refresh decision itself rather than the transport.

Front: verified in Chromium against the demo dev server — the end-of-list placement, the empty-search placement, and the three outcome states, at 1400px and 390px.

All three locales (en/fr/de) are at key parity and `npm run compare-translations` reports no errors.

https://claude.ai/code/session_01L6jC1TXBM8cBBjG2CqfzvM


front/src/config/i18n/fr.json
```diff
@@ -583,6 +583,15 @@
       "noIntegrations": "Aucune intégration trouvée",
       "noSearchResults": "Aucune intégration correspondant à « {{searchKeyword}} » n'a été trouvée.",
       "noSearchResultsSuggestion": "Essayez <a href=\"/dashboard/integration/device/matter\">Matter</a> pour les appareils compatibles Matter/Thread, ou <a href=\"/dashboard/integration/device/matterbridge\">Matterbridge</a> pour connecter des appareils Somfy, Shelly, Daikin et bien d'autres.",
+      "refreshStore": {
+        "hint": "Une int\u00e9gration manque \u00e0 l'appel ?",
```
Pierre-Gilles: C'est quoi ça, met des vrais caractères

Pierre-Gilles: Corrigé dans 156738d. C'était mon script d'insertion qui avait écrit des séquences d'échappement `é` au lieu des caractères réels — le reste du fichier est en UTF-8 brut, donc c'était incohérent en plus d'être illisible en diff.

Les quatre occurrences sont remplacées (`hint` et `refreshing` en fr, `refreshing` en en et de), il ne reste plus aucun `\u` dans les trois fichiers de langue.

---

front/src/routes/integration/index.js
```diff
@@ -71,11 +72,45 @@ class Integration extends Component {
...
+      const refreshedAt = externalStoreResponse ? externalStoreResponse.refreshed_at : null;
+      // the server never fails on an unreachable store: it answers with the
+      // cached catalog, and only bumps refreshed_at on a successful download.
+      // A timestamp that did not move means we are still looking at the cache
+      await this.setState({
+        externalInstalled,
+        externalStore: externalStoreResponse ? externalStoreResponse.integrations : [],
+        externalStoreRefreshedAt: refreshedAt,
+        refreshStoreStale: !refreshedAt || refreshedAt === previousRefreshedAt,
+        refreshStoreStatus: RequestStatus.Success
```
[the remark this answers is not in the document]
Pierre-Gilles: Valid finding, fixed in 156738d — flagging it here since it changes the API response shape.

The scenario is real: if the initial `GET /store` fails, the front stored a `null` baseline, and any non-null cached `refreshed_at` coming back from a later cache-fallback refresh then looked like a successful download.

`POST /api/v1/external_integration/store/refresh` now returns an extra `refreshed: true|false` alongside `refreshed_at` and `integrations`. It comes from a new `refreshCatalog()` on the supervisor, which captures `storeIndexFetchedAt` around the `getCatalog({ refresh: true })` call — the comparison happens server-side within one call, so there is no client baseline to get wrong. The front reads the flag and no longer tracks `refreshed_at` at all.

Covered by three supervisor tests (downloaded / store unreachable / no index ever fetched) and one controller test. They stub `refreshIndex` rather than the HTTP layer, so they assert the refresh decision itself.

---

PR #2717, opened 2026-07-27
Preserve integration catalog filters in URL for back navigation

Pierre-Gilles:
### Description of change

This PR adds URL-based state management for the integration catalog, allowing users to preserve their search filters and sort order when navigating back to the catalog from an integration detail page.

**Key changes:**

1. **New module `catalog-url.js`**: Centralized URL management for the integration catalog with functions to:
   - Build catalog URLs with filter parameters (search keyword, sort order)
   - Extract filters from query strings
   - Generate integration URLs that track their source catalog
   - Reconstruct the original catalog view when navigating back

2. **Integration component updates**:
   - Initialize filters from URL query parameters on mount
   - Update URL whenever filters change (search/sort)
   - Pass catalog context to external integration install pages
   - Maintain filter state across category navigation

3. **Install page updates**:
   - Replace hardcoded "back to catalog" link with dynamic URL that restores the user's previous view

**Benefits:**
- Users can use browser back button to return to their filtered catalog view
- Bookmarking or sharing integration URLs preserves the catalog context
- Filter state persists across navigation without relying on component state alone

### Pull Request check-list

- [ ] If your changes affect the code, did you write the tests?
- [ ] Are server tests passing with coverage?
- [ ] Did Cypress E2E tests pass?
- [ ] Is the linter passing?
- [ ] Did you run prettier?
- [ ] Did you test this pull request in real life?

https://claude.ai/code/session_01S8Xa7QqkDWwrxHY7R7uEas

## Summary by CodeRabbit

* **New Features**
  * Catalog URLs now preserve selected categories, search terms, and sort order.
  * Install flows now remember the originating catalog view for smoother “back” navigation.
  * Catalog navigation now updates browser URLs to improve sharing and back/forward behavior.
* **Bug Fixes**
  * “Back to catalog” now returns to the correct destination instead of a fixed route.


Pierre-Gilles:
Skipping the `await setState()` finding, with the reasoning:

- `index.js#L110-L116` (`refreshStore`) comes from #2714 and is outside this PR's diff, so it is not touched here.
- `L354-L365`: `updateURL()` already receives the new filter value explicitly (532b5d9), so the URL never depends on state timing. The remaining `getIntegrations()` calls do read `this.state`, and that is safe here: `setState()` calls `enqueueRender()`, which schedules preact's `process` through `Promise.prototype.then.bind(Promise.resolve())` (`preact/src/component.js`), and Gladys does not override `options.debounceRendering`. That microtask is queued before the `await` continuation, so the state is committed by the time the handler resumes:

```js
const defer = Promise.prototype.then.bind(Promise.resolve());
let applied = false;
(async () => {
  defer(() => {
    applied = true;
  });
  await undefined;
  console.log(applied); // true
})();
```

On top of that, the search path goes through a 300 ms debounce, so it reads the state long after the render.

`await this.setState(...)` is the established idiom across the Gladys front-end; changing it in these two handlers alone would be inconsistent, and reworking it globally is out of scope for this pull request.

---

PR #2718, opened 2026-07-28
Render the Markdown of the AI answers in the chat

Pierre-Gilles:
## Description

The AI answers with Markdown — `**27 °C**`, headings, bullet lists, `code` — and the chat displayed the raw syntax:

> La température actuelle dans le salon est de \*\*27 °C\*\*.

### Front: Markdown rendering in the chat

Messages written by Gladys are now rendered as Markdown (`front/src/utils/markdownRenderer.js`, `ChatMessageText.jsx`), with `marked` + `DOMPurify` — already dependencies, used for the store documentation — downloaded on the first message rather than in the main bundle. Auto-linked URLs keep the short label of the plain text renderer, and the plain rendering stays as fallback for what the user typed, for the tool call payloads, and while the renderer loads. A chunk that fails to download does not disable the formatting for the rest of the session.

One thing that had to move: the chat bubble was styled through its `<p>` tag (`.received_withd_msg p`). Markdown emits its own `<p>`, which would have inherited the bubble background, so the wrapper became a `<div>` and the style now targets `.msg_txt_wrap`.

### Spec: the text relayed to external integrations is Markdown

The core cannot render the Markdown for a communication channel — it knows neither Telegram's HTML subset, nor Discord's native Markdown, nor the fact that an SMS channel displays nothing. That belongs to each integration. What was missing is that the contract never said what format `text` is in, so an integration could only guess: `docs/specs/external-integrations.md` now documents it, in B.15 and in the C.4 protocol table, which did not even carry the `external-integration.message.send` row although the code implements it.

The Telegram side of the fix lives in the external integration that replaces the in-core service: [gladysassistant/gladys-telegram#3](https://github.com/GladysAssistant/gladys-telegram/pull/3). An earlier revision of this PR also patched `server/services/telegram`; it was dropped rather than maintaining the same converter twice in a service on its way out.

### Testing

The front has no unit-test script; the rendering was checked against the real markup and CSS (headings, bold, italic, lists, inline and fenced code, blockquote, table, shortened link), and `prettier-check`, `eslint`, `compare-translations` and `npm run build` are clean.

https://claude.ai/code/session_01K4UZQCgNF79FPsSKBPG7yF


Pierre-Gilles:
On the three CodeRabbit nitpicks:

**`markdownRenderer.js` — no retry after a failed dynamic import.** Valid, fixed in 587473d. A chunk that fails to download (offline, deploy in progress) no longer disables the formatting for the whole session: `loadingPromise` is cleared on rejection so the next message retries.

**`style.css` — `word-break: break-word` is deprecated.** Skipped. This file already uses `word-break: break-word` in two other places (`.msg_txt_wrap`, `.toolCallText`); switching only the new rule to `overflow-wrap: anywhere` would leave the file inconsistent for no behavioral gain. Worth doing as its own sweep over the three occurrences.

**`message.send.js` — no try/catch around `sendPhoto`.** Skipped on purpose. `sendPhoto` used not to be awaited at all, so a failure was an unhandled rejection; now that it is awaited, the rejection reaches the callers, which all handle it — `message.reply` catches and logs, `forwardToChannels` catches per channel and keeps going. Swallowing it here would hide a failed image send from those logs without saving anything: the text message has already been delivered at that point.

---

PR #2719, opened 2026-07-29
feat(external-integration): disconnect an oauth2 account

callemand:
### Pull Request check-list

- [x] Are server tests passing with coverage? — **100% of lines, branches and functions** on the new `externalIntegration.disconnectOAuth.js` (9 new tests)
- [ ] Did Cypress E2E tests pass? — no Cypress spec covers the external-integration pages, so none exercises the changed UI
- [x] Did you run the integration comparator? (`npm run compare-translations`) — the 3 locales are complete
- [x] Did you test this pull request in real life? With real devices? — see below
- [x] If your changes modify the API, did you modify the API documentation? — apidoc block on the new route
- [ ] User documentation — the disconnect button is self-explanatory; happy to document it if you'd rather
- [x] Fake requests data for the demo mode? — not needed: `demo.js` has no `external_integration` route at all

### Description of change

The `oauth2` config field could link an account but nothing could **unlink** it:
a user with a stale, wrong or revoked account had no way back other than
deleting the integration. This adds the missing counterpart.

**Disconnect (server).** `POST /external_integration/:selector/oauth/disconnect`
forgets the credentials the integration stored **off-schema** — that is where
the contract puts the tokens, see `relayOAuthCallback` — then pushes the emptied
config so the integration drops its session through its existing
`onConfigUpdated` handler.

Two consequences worth highlighting:

- **No SDK change is required**, so integrations already published get the
  feature for free.
- The values the user filled in the generated form, and the reserved `GLADYS_*`
  preferences, are **kept**: they are settings, not credentials.

**`noreferrer` on the authorize URL.** The Gladys URL is often a private LAN
address that has no business reaching the provider. It also fixes a real
breakage: some providers reject an authorize URL opened with a cross-site
`Referer`. Measured against the Xiaomi login endpoint:

| Request | Result |
| --- | --- |
| `Referer` present | rejected → `lpLogin/result?code=10012` ("Invalid request") |
| no `Referer`, `Sec-Fetch-Site: cross-site` | login page served |

`noopener` alone does not suppress the `Referer`; `noreferrer` does.

**UI polish**, all of it coming out of using the screen for real:

- the connection badge sits next to the account it qualifies, instead of next to
  the button — a button is an action, not a state;
- a field description explains what its button does, so it belongs above it;
- when connected, the button becomes a red **Disconnect**;
- an action button carries the action label ("Link the Xiaomi account") instead
  of a generic "Run";
- the per-transport counter on the Devices tab is gone: with a handful of devices
  it only repeated the badge that each device already shows.

### Tested in real life

Built a Roborock integration against these changes and drove a real
**Roborock S6** with it: account linked through the OAuth2 button, session
persisted, silent reconnection after a restart, state polled and commands
(return to dock, suction level) verified on the device. The connect → disconnect
→ reconnect cycle was exercised end to end.

### Notes

- Two pre-existing test failures on this machine are unrelated to the change and
  reproduce on a pristine `master`: `install_service_dependencies` (leaks
  `INSTALL_SERVICES_SILENT_FAIL` between tests) and `HomeKitService`
  (`EADDRINUSE` when a dev server is already running).

## Summary by CodeRabbit

* **New Features**
  * Added OAuth disconnect support for external integrations.
  * OAuth settings now show Connect or Disconnect actions with localized labels in English, French, and German.
  * Action buttons now display the specific action name.
  * Improved OAuth authorization privacy by limiting referrer sharing.

* **Improvements**
  * Removed transport status summaries from the device view.


server/lib/external-integration/externalIntegration.disconnectOAuth.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // config keys are stored uppercase in t_variable (see getIntegrationConfig)
+  const schemaKeys = configSchema.map((schemaField) => schemaField.key.toUpperCase());
+  const variables = await db.Variable.findAll({
+    where: { service_id: service.id, user_id: null },
+  });
+  const credentialVariables = variables.filter(
+    (variable) => !variable.name.startsWith(RESERVED_PARAM_PREFIX) && !schemaKeys.includes(variable.name),
+  );
+  await Promise.all(credentialVariables.map((variable) => this.variable.destroy(variable.name, service.id)));
```
Pierre-Gilles, line 42: The `key` parameter is validated but never used to scope the deletion: every off-schema variable of the service is destroyed. Two consequences worth weighing:

1. The manifest schema allows several `oauth2` fields in one `config_schema` — disconnecting account A would also wipe account B's tokens.
2. Off-schema storage isn't credentials-only by contract: `setIntegrationConfig` documents it as "a free internal storage of the integration (pairing state, third party tokens...)". Pairing state, device caches or webhook ids an integration stores there would be destroyed too, even though they have nothing to do with the OAuth session.

To be fair, without an SDK change there is no way to know which off-schema keys belong to which oauth2 field, so this may well be the right trade-off at constant contract. But it should at least be stated in the SDK/manifest documentation ("off-schema storage may be cleared when the user disconnects an oauth2 field"), or a per-field key-prefix convention considered for the future — otherwise a future integration can break silently.

---

front/src/routes/integration/all/external-integration/config-page/ActionsCard.jsx
```diff
@@ -54,7 +54,7 @@ const ActionsCard = ({
               onClick={() => runAction(action)}
             >
               <i class="fe fe-play mr-1" />
-              <Text id="integration.externalIntegration.actions.runButton" />
+              {label}
```
Pierre-Gilles, line 57: Two small things on this change:

- `integration.externalIntegration.actions.runButton` is no longer referenced anywhere — the key can be removed from the three locales.
- With the `<h4>` gone, an action that declares `fields` now renders its inputs with no title above them (the label only lives on the button, below the fields). Worth a visual check with a multi-field action.

---

front/src/routes/integration/all/external-integration/config-page/index.js
```diff
@@ -434,7 +434,24 @@ class ExternalIntegrationConfigPage extends Component {
...
+      // reject an authorize URL opened with a cross-site Referer.
+      window.open(authorizeUrl, '_blank', 'noopener,noreferrer');
+      this.setState({ oauthStatus: RequestStatus.Success });
+    } catch (e) {
+      console.error(e);
+      this.setState({ oauthStatus: RequestStatus.Error });
+    }
+  };
+
+  disconnectOAuth = async field => {
```
Pierre-Gilles, line 448: On failure this surfaces the `oauthConnectError` message ("Check that the integration is running"), which is misleading for a disconnect: the endpoint works even when the integration is down (your own test covers that case) — it only clears variables server-side. A dedicated `oauthDisconnectError` key would be clearer.

Related UX nit: the button drops the tokens with no confirmation. Defensible since the account can be re-linked, but a confirm step would match the gravity of the action.

---

Pierre-Gilles (commented):
Reviewed the full diff plus the surrounding code (`relayOAuthCallback`, `setIntegrationConfig`, `getIntegrationConfig`, `sendMessage`, `setConnectionStatus`, the config-page websocket handlers). Overall this is a well-built, well-tested PR — real-life tested, and the description is unusually thorough. One design point deserves discussion before merge; the rest is minor.

### What works well

- **No SDK change needed, and it holds up**: `sendMessage` is fire-and-forget and tolerates a disconnected integration, and the contract already requires integrations to resync `GET /config` on every (re)connection — so an offline integration correctly picks up the emptied config when it comes back.
- **Live UI update verified**: `setConnectionStatus` emits `CONNECTION_STATUS_UPDATED`, which the config page already listens to — badge and button flip without a reload.
- **Security is right**: route is `authenticated + admin`, the key is validated against the manifest (oauth2 fields only), `variable.destroy` is called with the correct signature, and the uppercase key handling matches `getIntegrationConfig`.
- **Tests are solid**: 9 tests against a real DB covering the nominal path, the offline-integration path, and all error cases, in the style of the existing suite.
- **The `noreferrer` change is a real fix**, backed by an actual measurement (Xiaomi rejecting a cross-site `Referer`), and `noopener` alone indeed does not suppress the `Referer`.

### Main discussion point

The deletion in `disconnectOAuth` is scoped to the *service*, not to the *field*: `key` is validated but never used to limit what gets wiped. Since off-schema storage is, per the existing contract, a free internal storage (pairing state, tokens, caches...), and since the manifest schema allows several oauth2 fields, this can destroy state unrelated to the account being disconnected. Details in the inline comment — probably acceptable at constant contract, but it should be documented or a namespacing convention considered.

### Minor

- Dead `runButton` i18n key in the three locales (inline comment).
- Disconnect failures reuse the connect error message; no confirmation before dropping tokens (inline comment).
- Theoretical race: an integration refreshing its tokens at the exact moment of the disconnect can re-write credentials via `POST /config` right after the wipe. Self-correcting via the `CONFIG_UPDATED` push that follows, so just worth knowing.

---

PR #2737, opened 2026-08-01
Add device migration feature for deprecated integrations

Pierre-Gilles:
### Description

This PR implements a device migration feature that allows users to migrate devices from deprecated integrations (Philips Hue, TP-Link) to replacement integrations in a single operation. The migration:

- **Moves DuckDB state history** from source to destination device features, with overlap handling (states older than destination&#39;s oldest state are moved; overlap period is deleted)
- **Rewrites scene and dashboard references** to point to the destination device and features
- **Handles feature mapping** with automatic matching and manual override capability
- **Cleans up SQLite leftovers** from pre-DuckDB-migration installs
- **Re-points energy children and energy price contracts** to the destination
- **Inherits room assignment** from source to destination device
- **Deletes the source device** after successful migration

The implementation includes:
- Backend migration logic (`device.migrate.js`) with comprehensive state/reference rewriting, an admin-only endpoint and an in-flight guard against concurrent migrations of the same source
- Frontend modal UI for destination selection and feature mapping with auto-matching (warnings on type/unit mismatch and unmapped features)
- Migrate buttons on deprecated device pages (Philips Hue, TP-Link, Netatmo, MELCloud, Tuya)
- Full test coverage for migration scenarios including overlap handling and reference rewriting
- Living specification document (`device-migration.md`) defining the behavior contract
- Job progress tracking with multiple steps (moving states, cleaning SQLite, updating references, rewriting scenes/dashboards)

### Screenshots

The whole flow below was executed for real against a dev instance (the success screen reports an actual migration of 8,640 DuckDB states, 1 scene and 1 dashboard).

**Deprecated badges in the integration catalog (Philips Hue, Netatmo):**

Catalog with deprecated badges``

**Deprecation banner and Migrate button on the device page:**

Device page with Migrate button``

**Migration modal — destination picked, features auto-matched by category/type:**

Migration modal with feature mapping

**Success screen with the migration report:**

Migration success

*(The images are pinned to an intermediate commit of this branch and were removed from the tree in the following commit, so they never land in `master`.)*

### Checklist

- [x] Tests pass: Added comprehensive unit tests in `device.migrate.test.js` and controller tests covering migration scenarios, overlap handling, and reference rewriting
- [x] No breaking changes: Feature is additive; existing device deletion behavior unchanged

https://claude.ai/code/session_01QPKCtDanRPEDg94pe1Hzhz

## Summary by CodeRabbit

* **New Features**
  * Added device migration to move compatible devices to replacement integrations with automatic or manual feature mapping.
  * Preserves applicable history, states, rooms, energy links, scenes, and dashboards while removing the original device.
  * Added migration controls, progress tracking, validation, success messages, and error handling across supported device pages.
* **Deprecations**
  * Marked Philips Hue and TP-Link integrations as deprecated and added notices to their pages.
* **Localization**
  * Added migration translations in English, French, and German.


Pierre-Gilles:
All review feedback from Cursor and CodeRabbit has been addressed in 28dfddc:

**Fixed**
- `t_energy_price.electric_meter_device_id` is now re-pointed to the destination device before the source is destroyed (+ test + spec §A row). Good catch — destroying a migrated meter would have silently detached the energy contracts.
- The migrate route is now `admin: true` (instance-wide operation: it rewrites the dashboards of every user; the integration pages carrying the button are admin-only anyway). Documented in the spec.
- Concurrent migrations of the same source device are now rejected with a `409 Conflict` via an in-memory in-flight guard (+ tests) — covers the client-timeout-retry scenario.
- The modal is hardened: confirm button disabled and `migrate()` re-entrant-safe while a migration is in flight, and `close()` (X / Escape / overlay / Cancel) is ignored mid-migration.
- Unit mismatches in the feature mapping now show a non-blocking warning like type mismatches (values are moved without conversion).
- The failure copy no longer claims retrying is fully "safe": it now says some history may already have been moved and points to the jobs page (en/fr/de).
- Added the `jobsSettings.jobTypes.device-migrate` labels (en/fr/de), added `key={device.selector}` to the Hue/TP-Link device lists, fixed the FR plural in `successText` and the pre-existing FR wording in `tooMuchStatesToDelete`, added the prettier commands to the spec's verification section, and added a test for the null `last_value_changed` branch.

**Declined, with rationale recorded in the spec ("Atomicity — a deliberate non-goal")**
- Wrapping the SQLite writes in one transaction: DuckDB cannot join it, so the irreversible part (the history move) stays outside anyway, and rolling back scene rewrites outside the scene manager path would desynchronize the RAM trigger cache from the DB. The design relies on ordering + convergent re-runs instead: the only data irreversibly deleted before the final destroy is history the user explicitly chose to drop (unmapped features) or overlap duplicates.
- Scoping the whole-table DuckDB cleanup in the new test suites: kept as-is for consistency with the existing pattern in `device.destroy.test.js` (suites run sequentially and each seeds its own states).

Screenshots of the full flow (executed for real against a dev instance) have been added to the PR description.

---

Pierre-Gilles:
Second review round addressed in 6a96517:

- **Jobs page renderer** (Cursor's remaining blocker): `JobData.jsx` now registers a `DEVICE_MIGRATE` renderer showing "Migration: source › destination" plus the translated current step, and the six step labels are in `jobsSettings.jobData.steps` (en/fr/de) — the failure/timeout copy pointing users at the jobs page now has something meaningful to show.
- **Dual-endpoint lock** (CodeRabbit): the in-flight guard now reserves both the source and the destination selectors for the whole run, so two migrations sharing a destination (or chained through one) are rejected with a 409 (+ test).
- **Synchronous double-submit latch** (CodeRabbit): the modal now uses an instance-level latch checked before `setState`, cleared in `finally`.
- **Stricter auto-match** (Cursor, optional hardening — taken): the automatic match now requires `(category, type, unit)` equality (null units compare equal); a sole °F candidate is no longer silently pre-selected for a °C source. Manual mapping still allows it, with the existing warning. Spec §B.1 updated.
- **DuckDB write-queue stall on huge histories**: acknowledged as a known trade-off of the single `UPDATE` per feature; recorded in the spec (§B.4) with the purge-style `created_at` slicing named as the hardening lever if field feedback asks for it. Left as-is for this one-off admin gesture, as suggested.

---

PR #2739, opened 2026-08-01
Show external communication integrations to non-admin users

Pierre-Gilles:
### Description

This PR implements role-based access control for external integrations, allowing non-admin users to link their own accounts on communication integrations (like Telegram) while restricting administrative screens to admins only.

**Key changes:**

- **Backend API**: Non-admin users now receive a reduced view of external integrations (manifest and status only, excluding operational fields like webhook URLs and docker images). The `/config` endpoint returns 403 for non-admins.
- **Frontend**: 
  - Non-admin users only see installed communication integrations in the catalog
  - Admin-only tabs (Supervision, Logs, Device screens) are hidden from non-admins
  - The configuration card and related admin cards (Webhooks, Actions, Hardware) are hidden from non-admins
  - Non-admins can still access the "Link Account" functionality on communication integrations
- **Tests**: Added comprehensive test coverage for non-admin user behavior, including list filtering, reduced detail view, and permission checks.

This follows the same pattern as native Telegram/Nextcloud Talk services, where non-admin users can only manage their personal account links, not the shared integration configuration.

### Checklist

- [x] Tests pass: Added new test suite for non-admin user access patterns
- [x] Linter and prettier pass
- [x] No breaking changes: Non-admin users gain new access to communication integrations (previously completely hidden); admin behavior unchanged

https://claude.ai/code/session_01WMPJfyQ6u53QRoANcFxXtt

## Summary by CodeRabbit

* **New Features**
  * Non-admin users can view installed communication integrations with limited details.
  * Users can access personal account-linking options for supported communication integrations.
  * Administrators retain full access to integration management, configuration, monitoring, and hardware controls.

* **Access & Security**
  * Sensitive integration data and administrative configuration are restricted to administrators.
  * Non-communication and incomplete integrations remain hidden from non-admin users.
  * External integration store and discovery features are limited to administrators.


Pierre-Gilles:
Both blockers addressed in 9d60f91 — the review was right that the RBAC story was half server-side, half "the frontend does not call it".

**1. `GET .../:selector` served device integrations to non-admins**

Fixed, with one deliberate difference from the suggestion: a non-admin asking for a non-communication selector now gets the exact same `404 EXTERNAL_INTEGRATION_NOT_FOUND` as an unknown selector, not a `403`. A `403` would still confirm the install exists, which is precisely the enumeration signal the comment was about. The test asserts both responses are byte-identical.

**2. Admin surfaces still open to any authenticated user**

Marked `admin: true` rather than loosening the prose, since all five are genuinely administration surfaces and the frontend only reaches them from admin screens:

| Route | Frontend caller |
|---|---|
| `GET .../store` | catalog, admin-only branch |
| `GET .../store/docs` | `DocsLink`, in the install screen and the (now admin-only) configuration card |
| `GET .../hardware` | install screen + configuration screen, admin-only fetch |
| `GET .../:selector/discovered_device` | Discovery tab (device screen, hidden from non-admins) |
| `POST .../:selector/scan` | same |

C.5 prose, the `*(admin)*` markers and `routes.js` now agree, and the reduced-view wording is precise about the manifest: it stays whole on purpose — it is the public description published by the store, `manifest.docker_image` included. What never leaves are the resolved `docker_image` of the install, the containers state and the webhook URLs. The comment above `toNonAdminView` says so too, so the next reader does not assume the image reference is hidden.

**Also applied:** the configuration card is now gated on `isAdmin` alone; a load error shows a non-admin a plain inline alert instead of the admin "Configuration" chrome.

New tests: device detail refused as an unknown selector, and the five newly admin-only routes refused for a non-admin. Full server suite run locally — no regression from these changes.

---

PR #2730, opened 2026-07-31
[zWaveJS] Thermostat support

sescandell:
### Pull Request check-list

To ensure your Pull Request can be accepted as fast as possible, make sure to review and check all of these items:

- [x] If your changes affect the code, did you write the tests?
- [x] Are server tests passing with coverage? (`cd server && npm run coverage`) — Codecov requires **100% coverage on lines changed in this PR**
- [ ] ~Did Cypress E2E tests pass? (`npm run cypress:run` from repo root, if UI changed)~
- [x] Is the linter passing? (`npm run eslint` on both front/server)
- [x] Did you run prettier? (`npm run prettier` on both front/server)
- [x] If you are adding a new feature/service, did you run the integration comparator? (`npm run compare-translations` on front)
- [x] Did you test this pull request in real life? With real devices? If this development is a big feature or a new service, we recommend that you provide a Docker image to the community ([forum](https://community.gladysassistant.com/)) for testing before merging. An AMD64 preview image is built automatically for non-draft PRs; for ARM64, comment `/build-arm64` on the PR.
- [ ] ~If your changes modify the API (REST or Node.js), did you modify the API documentation? (Documentation is based on comments in code)~
- [ ] ~If you are adding a new features/services which needs explanation, did you modify the user documentation? See [the GitHub repo](https://github.com/GladysAssistant/v4-website) and the [website](https://gladysassistant.com).~
- [ ] ~Did you add fake requests data for the demo mode (`front/src/config/demo.js`) so that the demo website is working without a backend? (if needed) See [https://demo.gladysassistant.com](https://demo.gladysassistant.com).~

NOTE: these things are not required to open a PR and can be done afterwards / while the PR is open.

### Description of change

Add Z-Wave thermostat support to the `zwavejs-ui` service. Until now, Gladys barely handled thermostat devices — this adds full read/write support for the core thermostat Command Classes, validated against a real Heatit Z-TRM6 floor thermostat.

- **Thermostat Mode (CC 64)** — Off / Heat / Cool / Energy Heat, exposed as an `AIR_CONDITIONING.MODE` feature (extends the existing `AC_MODE` enum with `OFF` and `ENERGY_HEAT`).
- **Thermostat Setpoint (CC 67)** — heating, cooling and energy-save-heating setpoints, each mapped to a dedicated read/write temperature feature (`THERMOSTAT.TARGET_TEMPERATURE` / `AIR_CONDITIONING.TARGET_TEMPERATURE`).
[...]

PR #2752, opened 2026-08-02
[Core] Enhancing Thermostat feature

sescandell:
### Description

Improving the Thermostat category (as requested in PR https://github.com/GladysAssistant/Gladys/pull/2730) introducing modes.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added thermostat mode controls for off, heating, cooling, auto, and energy-saving heating.
  * Added operating-state indicators for idle, heating, and cooling.
  * Thermostat options now appear in a consistent order, with the active mode highlighted.
  * Added thermostat icons and localized labels in English, German, and French.
  * Added clear fallback messaging for unavailable or unrecognized operating states.
  * Thermostat settings update smoothly when a mode is selected.


Pierre-Gilles (changes requested):
Thanks a lot for this PR @sescandell, the thermostat category really needs this! 🙏

The direction is great: generic `mode` / `operating-state` types, and `operating-state` with its unknown-value fallback is well designed — it maps cleanly to Matter `ThermostatRunningState` and Zigbee `runningState`. Two changes to discuss before merge — the Cursor review comments are all valid, and I'd like to expand on the main one.

## 1. Keeping `THERMOSTAT_MODE` to the cross-ecosystem common denominator

"Energy Save Heat" is specific to the Z-Wave Thermostat Mode CC: the Matter Thermostat cluster `SystemModeEnum` (see `server/services/matter/utils/thermostatMatterMapping.js`), the Zigbee ZCL equivalent and HomeKit don't have a matching mode. For core enums we try to stick to the values that exist across all ecosystems, so that every integration maps cleanly onto the same shared vocabulary — and since enum values are hard to walk back once integrations start publishing them, I'd rather be conservative here.

Could you:
- Reduce `THERMOSTAT_MODE` to `OFF / HEATING / COOLING / AUTO` (common to Matter, Zigbee, HomeKit and Z-Wave).
- In the zwavejs-ui integration, map "Energy Save Heat" onto `HEATING` — the same approach the Matter mapping takes for `EMERGENCY_HEAT` → `HEATING` and `SLEEP` → `AUTO`. The device stays fully usable, we just don't surface a niche distinction in the core model.
- Keep `supported_options` as designed: declaring which subset of the core catalog a given device supports (with `sort_order`). It technically tolerates out-of-catalog values via the label fallback, but that path gives an untranslated UI, so it's best kept as a robustness net rather than an extension mechanism.

## 2. Front component: reuse the existing adaptive control

As Cursor pointed out, `ThermostatModeDeviceFeature.jsx` should follow `AirConditioningModeDeviceFeature.jsx`: `resolveFeatureOptions(...)` + `AdaptiveOptionControl` instead of a hand-rolled fixed button group. That refactor also fixes two real issues for free:

- **Legacy fallback always shows Cool**: `mode <= THERMOSTAT_MODE.COOLING || mode <= deviceFeature.max` always passes for `COOLING`, so a heat-only thermostat (`max = 1`) still gets a Cool button — and heat-only devices (floor heating, radiators) are the main consumers of this category. The baseline here should be Off + Heat, with Cool/Auto gated on `max`.
- **Strict equality on `last_value`**: `lastValue === mode.value` without `Number()` coercion means the active button won't highlight when the value arrives as a string (which is why the AC component coerces defensively).

## Minor

- `fr.json`: `"Clim."` for the cooling mode reads as air-conditioning shorthand — `"Refroidissement"` (already used for the value key) keeps action/value labels consistent.
- `deviceFeatures.thermostat.mode: "Mode"` breaks the naming convention of its siblings ("Thermostat Temperature", "Thermostat Operating State") and is ambiguous out of context.
- `({ children, ...props })` — `children` is destructured but never used.

Everything here is a small change — happy to re-review quickly once updated, and thanks again for pushing the thermostat support forward! 🙂

---

sescandell:
@Pierre-Gilles If that is ok for you, this branch is ready to merge. So I can rebase and remap the zWave integration

Thanks.

Pierre-Gilles (dismissed):
Thanks for this PR, good for me!

Pierre-Gilles:
@sescandell Can you fix conflicts and I'll merge it after that :)

sescandell:
It's done @Pierre-Gilles 

Thanks,

PR #2756, opened 2026-08-02
Add a generic Charge Station status device feature

sescandell:
## Summary
- External integrations relaying EV chargers currently have no first-class way to report a charging station connector's status — only a raw `TEXT` feature, which loses the enum-aware dashboard treatment (icon + translated per-value label) other integer-state features get, the same treatment `VACUUM_CLEANER.STATE` already has.
- Adds a new `CHARGING_STATION` device feature category (`server/utils/constants.js`), kept separate from the vehicle-oriented `ELECTRICAL_VEHICLE_CHARGE`: connector/session status describes the charge point/EVSE's own lifecycle, not the vehicle's.
- Exposes **two** independent enum-backed features under it, mirroring OCPP 2.x's own split status model instead of OCPP 1.6's single flat enum — done this way to keep the contract forward-compatible with a future OCPP 2.x integration:
  - `CONNECTOR_STATUS` (`EV_CHARGE_CONNECTOR_STATUS`, 5 values: `AVAILABLE` / `OCCUPIED` / `RESERVED` / `UNAVAILABLE` / `FAULTED`) — aligned with OCPP 2.x's `ConnectorStatusEnumType`, reported independently of any session.
  - `CHARGING_STATE` (`EV_CHARGE_CHARGING_STATE`, 5 values: `CHARGING` / `EV_CONNECTED` / `SUSPENDED_EV` / `SUSPENDED_EVSE` / `IDLE`, new) — aligned with OCPP 2.x's `ChargingStateEnumType`, only meaningful while a charging session/transaction is active; no value otherwise.
- An OCPP 1.6 integration (the immediate target) maps its single, more granular `ChargePointStatus` deterministically and losslessly onto both enums — the full mapping table is documented as a comment directly above `EV_CHARGE_CHARGING_STATE` in `constants.js`.
- Frontend follows the existing `VACUUM_CLEANER.STATE` pattern for both features: `EvChargeConnectorStatusDeviceValue.jsx` and a new `EvChargeChargingStateDeviceValue.jsx`, both registered in `SensorDeviceFeature.jsx`, icon entries in `utils/consts.js`, and en/fr/de translations for all values plus the feature/category display names.
- MQTT catalog (`front/src/routes/integration/all/mqtt/device-page/utils.js`) wired for both types: correct sensor detection (`isSensorCategory`), read-only 0-4 default range instead of the generic 0-100 actuator default, representative preview values, and catalog preview labels.

## Context
Motivated by building [gladys-ocpp-integration](https://github.com/sescandell/gladys-ocpp-integration), an external integration relaying OCPP 1.6 EV chargers into Gladys. Without this, connector status can only be exposed as a generic `TEXT` feature (a plain "Text" badge on the dashboard) instead of a proper enum-aware status like other device categories get.

This PR iterated significantly from its initial draft, based on review feedback:
1. **Started** with a single 9-value status (`EV_CHARGE_CONNECTOR_STATUS`, mirroring OCPP 1.6's `ChargePointStatus` verbatim) living under the existing `ELECTRICAL_VEHICLE_CHARGE` category.
2. **Split off `CHARGING_STATION`** as its own category (prompted by cursor[bot]'s taxonomy comment): connector/EVSE lifecycle is a different device class than vehicle-side charge metrics (`charge-on`, `plugged`, `target-charge-limit`, ...).
3. **Redesigned as two features instead of one**, to natively match OCPP 2.x's split status model (`connectorStatus` + `chargingState`, two independent fields/messages) instead of OCPP 1.6's single flat enum. OCPP 2.0.1/2.1 was released after 1.6 and deliberately separates "is this connector available" from "what's happening in the current session" — this repo has no OCPP service yet, but anticipating that split now avoids a breaking change to the contract once a 2.x integration exists. A 1.6 integration still maps cleanly onto the two features (deterministic, lossless mapping — see the comment above `EV_CHARGE_CHARGING_STATE` in `server/utils/constants.js`); a future 2.x integration becomes a near-passthrough.

## Test plan
- [x] `npm run compare-translations` (front) — all three locale files complete
- [x] `npx eslint` on all changed frontend files — clean
- [x] `npx prettier --check` on all changed files — clean
- [x] `npx eslint` / `npx prettier --check` on `server/utils/constants.js` — clean (one pre-existing, unrelated warning)

## Summary by CodeRabbit

- **New Features**
  - Added support for displaying electric-vehicle charging-station connector status and charging state.
  - Added localized badges, labels, icons, and fallback values for charging-station information.
  - Added MQTT device previews for charging-station status and charging states.
  - Added English, German, and French translations, including discharging and unknown states.


Pierre-Gilles (commented):
Thanks for the PR and for opening it as a draft to discuss scope first — the execution is clean (the `VACUUM_CLEANER.STATE` pattern is followed faithfully, all three locales are complete, and the history view resolves the labels generically), but the question you raise yourself in the description needs to be settled before merge, because the type slug, enum values and i18n keys become a public contract that's hard to change later.

## Main point: this is an EVSE status, not a vehicle status

All `ELECTRICAL_VEHICLE_*` categories describe **the vehicle** (the comments in `constants.js` say so explicitly). The proposed enum describes the lifecycle of a **charge point connector**: `AVAILABLE`, `RESERVED`, `UNAVAILABLE`, `PREPARING`, `SUSPENDED_EVSE` make no sense reported by a car — a Tesla will never be "Reserved" or "Unavailable" in the OCPP sense.

**Recommendation: move this to a dedicated charging-station category** (e.g. `charging-station` / `ev-charger`) with a simple type name like `status`. A charger is a standalone device in Gladys (an OCPP integration creates a "charger" device, not a "car" device), and that category will naturally grow other charger-side types later (session power/energy, max allowed current…), which carry a different semantic than their vehicle-side counterparts.

## Enum coverage across brands

The OCPP 1.6 superset is a good base for **chargers** (OCPP 2.x is a subset; Wallbox/Easee/go-e map fine), but as-is it does not cover **vehicle** integrations:

- **`COMPLETE` is missing** — the most common state in car APIs (Tesla `Complete`, Renault, Kia…). Mapping it to `FINISHING` is misleading: OCPP's `Finishing` is a transient session wrap-up, and the label would read "Finishing" for hours on a full, parked car.
- **`DISCHARGING` is missing** — Matter Energy EVSE has `PluggedInDischarging`, and V2G/V2L is coming fast. Worth adding now: `min`/`max` (0–8) are frozen on features once created in DB.
- `SCHEDULED` (off-peak charging wait) only maps lossily to `SUSPENDED_EV` — worth discussing.

So I'd suggest not trying to make this one enum serve car integrations too. If a vehicle-side "charge status" need is confirmed (Tesla etc.), it should be a **separate type** under `ELECTRICAL_VEHICLE_CHARGE` with brand-neutral values including `COMPLETE`. Merging both produces absurd labels on one side or the other ("Available" on a car, "Charge complete" on a free connector).

## Minor nits

- `EV_CHARGE_CONNECTOR_STATUS` is inserted between `VACUUM_CLEANER_STATE` and `VACUUM_CLEANER_MODE`, splitting the vacuum enum block — please move it out of that group.
- Alphabetical ordering: in `DEVICE_FEATURE_TYPES.ELECTRICAL_VEHICLE_CHARGE` and in the three i18n files the new key is inserted after `plugged`/`last-charge-energy-consumption` while the lists are sorted (it belongs after `charge-voltage`). The icon file got it right — placement is inconsistent across files.
- `SUSPENDED_EVSE`/`SUSPENDED_EV` is OCPP jargon for a generic enum; `PAUSED_BY_CHARGER`/`PAUSED_BY_VEHICLE` would be more readable, and the English label "Suspended (EVSE)" exposes that jargon to end users ("charger" would match the French "(borne)").
- `front/src/routes/integration/all/mqtt/device-page/utils.js` isn't updated: the MQTT page's example value falls back to `42`, which will render as "42 (unknown)". Non-blocking, but a dedicated entry (and the 0–8 min/max hint) would be consistent with the rest of the category.

A scene-trigger value dropdown would be a nice follow-up regardless of the category decision (vacuum state doesn't have one either, so parity is respected here).

---

Pierre-Gilles (commented):
Thanks for the quick follow-up — the structural rework in 43a5666 addresses the main points of my previous review:

✅ **Resolved**
- Dedicated `CHARGING_STATION` category created, with `CONNECTOR_STATUS` moved out of `ELECTRICAL_VEHICLE_CHARGE` (correct alphabetical placement in categories, icons and i18n).
- MQTT page updated: sensor category, `min: 0 / max: 8 / read_only: true` defaults, preview value and translated catalog label.
- German label and OCPP 2.x comment fixed.

⚠️ **Still open before merge**

1. **Enum value set** — still the raw OCPP 1.6 nine values. Main remaining decision: please add `DISCHARGING: 9` now (Matter Energy EVSE has `PluggedInDischarging`, V2G/V2L is coming fast), and consider `SCHEDULED`. This matters more now that `max: 8` is hardcoded in the MQTT defaults: extending the enum later means touching existing features in DB.
2. **Leftover naming from the old scope** — the enum is still `EV_CHARGE_CONNECTOR_STATUS` and the front component `EvChargeConnectorStatusDeviceValue`. Now that the category is `charging-station`, please rename to `CHARGING_STATION_CONNECTOR_STATUS` / `ChargingStationConnectorStatusDeviceValue` while nothing is published against them.
3. **Enum placement** — `EV_CHARGE_CONNECTOR_STATUS` is still inserted between `VACUUM_CLEANER_STATE` and `VACUUM_CLEANER_MODE`, splitting the vacuum enum block; please move it out of that group.
4. **`SUSPENDED_EVSE` / "Suspended (EVSE)"** — OCPP jargon in a generic enum; `PAUSED_BY_CHARGER`/`PAUSED_BY_VEHICLE` (and "(charger)" in the English label, matching the French "(borne)") would read better for end users.
5. Forward-looking nit: the MQTT `if (category === CHARGING_STATION)` defaults block applies `0–8` to the whole category — fine with a single type, but it will need to be conditioned per type as soon as the category grows (session power/energy…).

Once point 1 is decided and the renames are in, this looks good to me.

---

Pierre-Gilles (approved):
Great PR 🙌  Good for me!

PR #2769, opened 2026-08-03
Route external integration OAuth2 through an HTTPS redirect page

Pierre-Gilles:
### Description

OAuth2 providers now require an HTTPS redirect URI. Spotify [enforces it since April 2025](https://developer.spotify.com/documentation/web-api/concepts/redirect_uri): only `https://` is accepted, plus the loopback literals `http://127.0.0.1:PORT` and `http://[::1]:PORT` — `localhost` is refused, and so is any LAN address.

The frontend built `redirect_uri` from `window.location.origin`, so anyone reaching Gladys at `http://192.168.1.50:1443` — most self-hosted users — got `redirect_uri: Not matching configuration` and simply could not connect a Spotify account.

`redirect_uri` is now the fixed HTTPS page `https://my.gladysassistant.com/redirect/oauth` (static site, source in [GladysAssistant/my.gladysassistant.com](https://github.com/GladysAssistant/my.gladysassistant.com)). The address to come back to travels in the OAuth2 `state`, the only parameter a provider hands back untouched:

```js
state = base64url(JSON.stringify({ v: 1, origin, path, state: '<the integration state>' }));
```

The redirect page unwraps it, shows the visitor which instance they are about to return to, and bounces them there with every provider parameter forwarded. Nothing runs server-side there: no authorization code, token or credential ever reaches a Gladys server. It only redirects to an HTTPS origin or to a plain-HTTP origin that can exist solely on the visitor's own network, always behind an explicit click.

This is the same approach Home Assistant takes with `my.home-assistant.io`.

**Integrations need no change.** They already receive the `redirect_uri` and use it as-is (`onOAuthAuthorizeUrl(key, redirectUri)`), and they get their own `state` back to verify. The server is untouched too — the whole change is in the frontend.

Also in this PR:

- the configuration screen now **shows the redirect URI to copy** into the provider's developer application, with a copy button. Guessing it wrong is what people were stuck on in the forum thread.
- users who already serve Gladys over HTTPS can **opt out** and use their own address instead (offered only when the frontend is served over HTTPS, remembered in `localStorage`).
- the `redirect_uri` used is stored alongside the oauth2 key so the callback sends back exactly the same one — the provider compares it byte for byte at token exchange.

**One behaviour change for integration authors**: the `state` becomes mandatory in the authorize URL. Without it there is no way home, so an integration that omits it now gets an explicit error message instead of failing silently. `docs/specs/external-integrations.md` is updated accordingly.

A side benefit worth knowing: a single URL now has to be declared at the provider, whether the user reaches Gladys locally or through Gladys Plus. Before, the redirect URI depended on how you were browsing.

## Forum

Forum: https://community.gladysassistant.com/t/impossible-de-creer-le-plugin-spotify/10439

### Checklist

- [ ] Tests pass: `cd server && npm run coverage` (Codecov requires 100% coverage on changed lines) and Cypress (`npm run cypress:run`) if the UI changed

On the first box, being explicit about what was and was not run:

- No server code changed, so the server suite is untouched by this PR.
- Cypress was not run — it needs a live Gladys stack with an installed external integration.
- `npm run build`, `npm run eslint` (0 errors), `npm run prettier-check` and `npm run compare-translations` all pass on the front.
- The wrapping was verified end to end against the redirect page: a realistic Spotify authorize URL goes through `wrapAuthorizeUrl()`, then through the page's `decodeWrappedState()` / `buildRedirectUrl()`, and comes back as `http://192.168.1.50:1443/dashboard/.../oauth-callback?code=…&state=<the integration state>` with `client_id`, `scope` and `redirect_uri` untouched. The redirect page itself has 15 unit tests covering the origin and path checks.
- The front has no unit test runner today, so `front/src/utils/oauth.js` ships without tests. Happy to add one if you want a runner introduced.

---
_Generated by [Claude Code](https://claude.ai/code/session_0197FhTmFSecAUY5CqCYG7vN)_

## Summary by CodeRabbit

- **New Features**
  - Added copyable HTTPS redirect URIs for OAuth setup.
  - Added support for instance-specific redirect addresses when available, with a user-selectable option.
  - Improved OAuth callback handling with stronger security-state validation.
  - Added clearer connection guidance and invalid-state error messages.
  - Added English, German, and French localization updates.

- **Documentation**
  - Expanded OAuth guidance covering redirect setup, local HTTP instances, callback routing, and token handling.


Pierre-Gilles:
All six findings were valid and are fixed in `1f086a7`. Two of them were real bugs, so thanks for the catch.

**Legacy redirect URI fallback** (`cursor`, `coderabbitai`) — the fallback contradicted the comment right above it. A flow started before this release stored no redirect URI and authorized with the address of the callback page itself, so falling back to the fixed page would have been refused at the token exchange. It now falls back to `${window.location.origin}${window.location.pathname}`, which is exactly the pre-PR value.

**Missing state validation in the instance-redirect branch** (`coderabbitai`) — correct, that branch skipped `wrapAuthorizeUrl()` and with it the only check that the integration generated a state, so the mandatory-state rule only applied to one of the two modes. The check is now a separate `getAuthorizeUrlState()` called by both branches.

**Bounds not mirrored before `window.open`** (`cursor`) — worth doing: an integration emitting a state longer than the page accepts (a JWT, say) would only fail *after* the user consented at the provider, at a dead end. The frontend now mirrors the same limits and fails on "Connect". Since the error covers both cases now, `oauthMissingStateError` became `oauthInvalidStateError` ("missing or too long").

**Description shown in the wrong mode** (`cursor`) — real trap: users who opted into their own HTTPS callback saw the URL of their instance in the input while the text below explained the Gladys-hosted page, inviting them to register the wrong URL at the provider. Added `oauthRedirectUriInstanceDescription` for that case.

**Clipboard rejection** (`coderabbitai`) — wrapped in try/catch; the input stays selectable for manual copy.

**"No data goes through our servers" overstates the trust model** (`cursor`) — agreed, and this was the most useful comment. The authorization code does appear in the query string of a request to a Gladys-operated host, so that claim was wrong regardless of the page being static. Reworded en/fr/de to what is actually true: nothing is stored, Gladys neither stores nor uses the code, tokens are exchanged directly between the instance and the provider. The living spec now spells out what the page does and does not see, with an explicit "do not claim that no data ever touches Gladys-operated infrastructure". The same wording was corrected on the redirect page itself and its README in [GladysAssistant/my.gladysassistant.com@aa1df5e](https://github.com/GladysAssistant/my.gladysassistant.com/commit/aa1df5e).

Re-verified after the changes: front build, eslint (0 errors), prettier, `compare-translations`, the 15 unit tests of the redirect page, the browser checks on that page, and the cross-repo round trip — including the three new cases (missing state, over-long state, instance-redirect mode validating the state).

One note on the `cursor` remark about the 308 redirect: `my.gladysassistant.com` is not deployed yet. It needs to be live on Cloudflare Pages before this merges, otherwise the "Connect" button points at a page that does not exist.

---

Pierre-Gilles (commented):
## Review

The approach is the right one: routing the flow through a fixed HTTPS redirect page is the proven `my.home-assistant.io` pattern, and the `state` parameter is indeed the only reliable channel for the round trip. The tricky OAuth2 details are handled well: storing `externalIntegrationOAuthRedirectUri:${selector}` so the token exchange sends back the exact same `redirect_uri` byte for byte, the backward-compatible fallback in the callback page for flows started before this release, and the mandatory `state` failing fast at "Connect" time (with a dedicated error message) rather than dead-ending after the user has already consented at the provider. The spec update is honest about what the Gladys-operated hosting does and does not see.

### Requested before merge

**1. The copy button is invisible for the very users this PR targets** — `ConfigSchemaForm.jsx` gates the button on `window.isSecureContext`, but the audience this PR unblocks is precisely on plain-HTTP LAN addresses, i.e. a non-secure context: they get the URI field but never the button. `navigator.clipboard` is indeed unavailable there, but the legacy `document.execCommand('copy')` works in HTTP contexts (user gesture + DOM selection required), and the repo already ships exactly this fallback in `TuyaGithubIssueSection.jsx` (lines 189–194). Please reuse that pattern so HTTP users get the one-click copy too.

**2. A Cypress pass on a real flow before merge** — given the `risk:high` label and that Cypress was not run, a manual pass against a live stack with an installed external integration (both the redirect-page path and the instance-redirect opt-out) would derisk this significantly. The offer to introduce a front unit runner for `front/src/utils/oauth.js` is worth taking too: those functions are pure and ideal test candidates.

### Minor

- `onCopyRedirectUri` reaches the value via `e.target.closest('.input-group').querySelector('input')` — fragile coupling to the Bootstrap markup. The `redirectUri` value is already computed in `render()`; recomputing it in the handler (or passing it as a parameter) would be more robust than a DOM query.
- `wrapAuthorizeUrl()` mutates `url.searchParams`, which re-serializes the whole query string and can change the encoding of provider parameters (`%20` → `+` in `scope`, for instance). Verified end to end against Spotify per the PR description, so fine as is — just worth keeping in mind if an exotic provider ever chokes; a string-level replacement of only the `state` parameter would be the conservative alternative.
- `MAX_INNER_STATE_LENGTH` / `MAX_WRAPPED_STATE_LENGTH` are mirrored from the redirect page living in another repo — silent-drift risk. Referencing the exact file in `my.gladysassistant.com` in the comment would make it actionable.
- In `config-page/index.js`, the comment says the preference is "remembered for the whole instance", but `localStorage` is per-origin per-browser: a user switching between LAN and Gladys Plus gets two independent preferences (arguably desirable — the Plus origin is HTTPS — but the comment overstates it).
- The `de.json` hunk replacing `„` escapes with literal quotes in `contextTypes.device_feature` is unrelated to this PR (likely a prettier pass).

### Security notes

The wrapped state is unsigned, so anyone can craft one pointing at an arbitrary origin — mitigated by the explicit click + origin display on the redirect page (out of scope of this repo), and crucially the inner state is still verified by the integration, so the anti-CSRF protection is intact. The instance's local origin (private IP) now transits in the `state` seen by the provider — a minor, inherent disclosure with no realistic alternative. Tokens still never transit through the frontend or the page.

### Verdict

Approvable once the two requested items land: the `execCommand` copy fallback for HTTP users, and a Cypress pass on a real flow. Everything else is nits.

---

Pierre-Gilles:
Both requested items are done in `52cf484`, plus the nits.

**1. Copy button invisible for HTTP users** — you're right, and it was the worst kind of bug: the feature silently disappeared for exactly the audience the PR exists for. Removed the `window.isSecureContext` gate, the button always renders, and `copyRedirectUri()` now falls back to focus/select/`document.execCommand('copy')` when `navigator.clipboard` is absent or rejects — same shape as `TuyaGithubIssueSection.jsx`. Covered by a test that forces the fallback path (stubs `writeText` to reject, asserts `execCommand('copy')` runs and the "copied" feedback shows).

**2. Cypress on a real flow** — done, and it found nothing further, but it is now permanent coverage rather than a one-off pass. Two specs, in the existing Cypress job, no new tooling or CI step:

- `cypress/e2e/utils/OAuthUtils.cy.js` — the pure functions of `front/src/utils/oauth.js` (Cypress bundles specs, so a spec can import from `src` directly, which is why I did not introduce a separate unit runner). Wrapping and decoding, unicode states, provider parameters left untouched, missing and over-long states.
- `cypress/e2e/routes/integration/external-integration/ExternalIntegrationOAuth.cy.js` — the real configuration screen: real routing, real session, real components, real request to the server. Only the integration is stubbed, since it would need an installed Docker container. Asserts the `redirect_uri` actually sent, the wrapped state of the URL passed to `window.open`, what lands in `localStorage`, the refusal of a state-less authorize URL, and the clipboard-less copy.

Run locally against a real stack (server on `:1443` with a clean `gladys-cypress.db`, front built with the Cypress env on `:1444`): **27 specs, 134 tests, 131 passing, 3 pre-existing skips, 0 failures.** The new specs pass, and nothing else regressed. What I still cannot do here is a pass against a live Spotify app with a real installed integration — the preview Docker image on this PR is the way to close that last gap, and the redirect page is live now.

**Nits** — all applied: the copy handler receives the value instead of querying the DOM; the comment now names `assets/js/oauth-state.js` in `my.gladysassistant.com` as the source of the mirrored bounds; the preference comment says per browser and per origin (with the Gladys Plus consequence spelled out); and the unrelated `de.json` hunk is reverted — good catch, my JSON rewrite had un-escaped `„`/`“` in `contextTypes.device_feature`.

On `wrapAuthorizeUrl()` re-serializing the query: agreed on the risk, leaving it as is per your call. If an exotic provider ever chokes on it, the fix is a string-level swap of the `state` parameter only.

On the unsigned wrapped state: worth recording that we agree on the threat model. Signing it would not buy much — the signer and the verifier would be the same instance, and an attacker can always craft their own state and send the victim a link. What actually protects the flow is that the code is delivered to an origin the visitor sees and confirms, and that the inner state is verified by the integration.

---

PR #2780, opened 2026-08-05
Add an account_link config field for providers that are not OAuth2

callemand:
### Description

Not every service a cloud integration has to sign into is an OAuth2 one, and `oauth2` is currently the only config field able to open a URL in the user's browser.

Porting a Roborock integration to **Xiaomi Home** made the mismatch concrete. The sign-in is a QR page the user approves in the Xiaomi Home app; Xiaomi then redirects to **its own** STS endpoint, never back to Gladys, and the integration learns about the approval by long-polling the provider. Checked against the live API — the URL carries `ticket, dc, sid, ts` and nothing else: no `client_id`, no `redirect_uri`, no authorization code, no `state`, no token endpoint.

Borrowing the `oauth2` field there puts two untruths on screen:

- a **redirect URI to declare in a developer application** that does not exist, for a round trip that never happens;
- since #2769 made the anti-CSRF state mandatory, _"This integration did not provide a usable security state… It cannot be connected until its developer fixes it"_ — unless the integration fabricates a placeholder state, which is what I ended up doing.

So this declares what it actually is, rather than patching `oauth2` with an exception.

**`account_link`** shares everything genuinely common with `oauth2`: a Connect button, no stored value, credentials kept off-schema, the same `oauth.get-authorize-url` relay and the same `connection_status` driving the badge. It differs exactly where the flow does:

| | `oauth2` | `account_link` |
|---|---|---|
| `redirect_uri` | required | ignored (there is none) |
| anti-CSRF `state` | mandatory | not required — no round trip to protect, nothing to verify |
| callback route | used | never involved |
| rendering | redirect-URI block + instance-redirect option | Connect button and status only |

**No SDK change is needed**: the existing WebSocket message is reused, so an integration keeps its single `onOAuthAuthorizeUrl` handler. `redirect_uri` is simply absent from the payload for this type.

Two details that matter in practice:

- **`noreferrer` on top of `noopener`** for this type. The address of the instance — often a private LAN one — has no business leaking to a provider that will never talk to it, and Xiaomi rejects a sign-in URL opened with a cross-site `Referer` outright (isolated: `lpLogin/result?code=10012`, while `Sec-Fetch-Site` alone is fine). `oauth2` behaviour is left untouched here.
- the URL an integration returns is checked to be **http(s) before being opened** (`assertOpenableUrl`). `noopener,noreferrer` does nothing against an active scheme: `window.open('javascript:…')` runs in a document inheriting this origin. `http` is deliberately allowed — a provider approved on the local network, a device pairing page, has no certificate to offer.

Like `oauth2`, the field is refused in a per-user `contact_schema` (linking a provider account is integration-scoped, never per user) and refuses `default`, `placeholder` and any direct value.

Verified end to end on a real Roborock S6 through a Xiaomi Home account: the redirect-URI block is gone, the placeholder state is no longer needed, and the account links as before.

### Checklist

**On the checklist, precisely:**

`npm run coverage` was run. The four changed server files come out at **100% statements / lines / functions**; the few uncovered branches in them sit on lines this PR does not touch (the `= {}` parameter default and the `service.manifest &&` fallback). 6 new tests cover the relay without a `redirect_uri`, the `redirect_uri` still being required for an `oauth2` field, the type being refused in a `contact_schema`, and `default` / direct values / front exposure being refused. Full suite: **2883 passing**, with the same 3 failures as pristine `master` in this environment (`install_service_dependencies`, `jsonBodyMiddleware`, `HomeKitService`) — compared run for run on both branches.

`npm run eslint` and `npm run prettier-check` pass on front and server.

**Cypress was not run**, although the UI changed — I have no Cypress-capable setup here. The existing `ExternalIntegrationOAuth.cy.js` specs exercise the `oauth2` path, which this PR leaves functionally unchanged, but that deserves a real run before merge. No spec was added for the new type; happy to write one if you point me at how you'd want it fixtured.

Spec updated in `docs/specs/external-integrations.md`.

<sub>An earlier revision of this PR also rendered the sign-in URL as a QR code under the button. It is gone: the Xiaomi Home app turns out to have no scanner, so the caption was inviting users to do something they cannot, and a picture appearing on a configuration screen for that is noise.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added non-OAuth account linking through authorization URLs, including QR code support.
  * Added secure HTTP(S) authorization URL validation.
  * Added English, German, and French localization for authorization URL errors.
  * Added connection status messaging for external integrations.

* **Bug Fixes**
  * Improved OAuth error handling while preserving standard OAuth2 redirect behavior.
  * Prevented account-link and OAuth credentials from being exposed or assigned directly.
  * Restricted account-link fields from per-user contact schemas and direct defaults.


callemand:
Thanks — two of the three fixed in fac24666, one skipped with a reason.

**Fixed — active URL schemes (Major).** Valid, and specific to this new path. `noopener,noreferrer` indeed does nothing against `javascript:`, which would run in a document inheriting this origin. Added `assertOpenableUrl` in `front/src/utils/oauth.js`, applied before `window.open`:

| URL | |
|---|---|
| `https://eu.account.xiaomi.com/longPolling/login?ticket=lp_42` | accepted |
| `http://192.168.1.50/pair` | accepted |
| `javascript:alert('xss')` | rejected |
| `data:text/html,<script>…` | rejected |
| `ftp://host/x`, `""`, `not-a-url` | rejected |

`http` is deliberately kept: a provider approved on the local network — a device pairing page — has no certificate to offer, and that is a legitimate `account_link` case. A refusal now surfaces a dedicated message (`oauthInvalidUrlError`, fr/en/de) rather than the generic "check that the integration is running", which would send the user chasing the wrong problem.

**Fixed — Save eligibility (Minor).** Correct, that was an omission on my side: `hasSavableField` excluded `section` and `oauth2` but not `account_link`, so a schema made only of account fields showed Save and posted an empty configuration. It now excludes both, through the shared `ACCOUNT_FIELD_TYPES`.

**Skipped — opening the tab before the request (Major).** The concern is real but it is **not introduced by this PR**, and fixing it here would change the `oauth2` flow this PR deliberately leaves alone. On `master`, `connectOAuth` already awaits `httpClient.post(...)` before calling `window.open`, so the popup has been outside the user-gesture task since #2769 — my change to `onOAuthConnect` only awaits a promise that was previously left floating, which does not move `window.open` relative to the gesture:

```js
// master, unchanged by this PR
const { authorize_url: authorizeUrl } = await this.props.httpClient.post(...);
// …
window.open(urlToOpen, '_blank', 'noopener');
```

Pre-opening a blank tab and navigating it afterwards is the right fix, but it affects both field types and deserves its own change — happy to open a follow-up PR or an issue, whichever you prefer. For what it is worth, the tab does open in practice on the instance I tested this on (a real Xiaomi Home account, one desktop browser); I have not checked Safari, where the heuristics are reportedly stricter, so I would not claim the current behaviour is safe everywhere.

front/src/routes/integration/all/external-integration/config-page/ConfigSchemaForm.jsx
```diff
@@ -343,7 +342,6 @@ const ConfigSchemaForm = ({
   saveConfigStatus,
   updateConfigValue,
   saveConfig,
-  connectionStatus,
   oauthStatus,
   oauthInvalidState,
```
Pierre-Gilles, line 346: **Bug: the new `oauthInvalidUrl` error message can never render.**

`ConfigSchemaForm` neither destructures nor forwards the new `oauthInvalidUrl` prop, so `this.props.oauthInvalidUrl` inside `ConfigField` is always `undefined`.

Repro: an integration returns an `authorize_url` like `javascript:alert(1)` or `not a url` → `assertOpenableUrl` throws `EXTERNAL_INTEGRATION_OAUTH_INVALID_URL` → `index.js` sets `state.oauthInvalidUrl = true` → `ConfigTab` passes `oauthInvalidUrl={props.oauthInvalidUrl}` to `ConfigSchemaForm`, which silently drops it. The user then sees the generic `oauthConnectError` ("check that the integration is running") — a misleading message — instead of the `oauthInvalidUrlError` string added in all three locales for exactly this case.

Fix: add `oauthInvalidUrl` to this destructure and pass it down to `<ConfigField>` (alongside `oauthInvalidState`).

---

server/lib/external-integration/externalIntegration.getOAuthAuthorizeUrl.js
```diff
@@ -23,14 +29,14 @@ async function getOAuthAuthorizeUrl(selector, { key, redirect_uri: redirectUri }
...
-  if (!field || field.type !== 'oauth2') {
-    throw new BadParameters(`config.${key}: not an oauth2 field`);
+  if (!field || !ACCOUNT_FIELD_TYPES.includes(field.type)) {
+    throw new BadParameters(`config.${key}: not an oauth2 or account_link field`);
+  }
+  if (field.type === 'oauth2' && (typeof redirectUri !== 'string' || redirectUri.length === 0)) {
+    throw new BadParameters('redirect_uri: must be a non-empty string');
   }
   const result = await this.sendCommand(service, WEBSOCKET_MESSAGE_TYPES.EXTERNAL_INTEGRATION.OAUTH_GET_AUTHORIZE_URL, {
     key,
```
Pierre-Gilles, line 42: **Contract deviation: `redirect_uri` is still relayed for `account_link` fields.**

The payload on the next line always includes `redirect_uri: redirectUri`, so a caller-supplied value is relayed verbatim to the integration even for an `account_link` field — while the spec updated in this same PR says the WS message's `redirect_uri` "is absent for an `account_link` field".

Scenario: an admin API client POSTs `{key: 'xiaomi_account', redirect_uri: 'https://x'}` to `/oauth/authorize_url` → the `sendCommand` payload includes `redirect_uri: 'https://x'` → an integration keeping a single handler (as the spec recommends) that distinguishes the two flows by `redirect_uri` presence treats the `account_link` request as an oauth2 one and builds the wrong URL.

Fix:
```js
redirect_uri: field.type === 'oauth2' ? redirectUri : undefined,
```

---

front/src/routes/integration/all/external-integration/config-page/ConfigTab.jsx
```diff
@@ -118,9 +150,10 @@ const ConfigTab = props => {
                     saveConfigStatus={props.saveConfigStatus}
                     updateConfigValue={props.updateConfigValue}
                     saveConfig={props.saveConfig}
-                    connectionStatus={get(integration, 'connection_status')}
+                    connectionStatus={connectionStatus}
```
Pierre-Gilles, line 153: **Dead prop.** The cleanup commit removed `connectionStatus` from `ConfigSchemaForm`'s parameter list and from `ConfigField`, but this producer was left behind — the value passed here is silently discarded.

```suggestion
```

---

server/lib/external-integration/externalIntegration.getConfigForFront.js
```diff
@@ -33,9 +33,10 @@ async function getConfigForFront(selector) {
       if (hasValue) {
         configuredSecrets.push(field.key);
       }
-    } else if (field.type === 'oauth2') {
-      // nothing is ever stored under an oauth2 key (tokens live off-schema):
-      // the front renders a Connect button, fed by connection_status
+    } else if (field.type === 'oauth2' || field.type === 'account_link') {
```
Pierre-Gilles, line 36: **Nit: use the `ACCOUNT_FIELD_TYPES` constant this PR introduces for exactly this pair.**

This branch re-lists the members inline while `getOAuthAuthorizeUrl.js`, `validateManifest.js` and the frontend all use the constant. A future third account type (or a rename) would have to be found and updated here separately, or `getConfigForFront` would start returning stored values for it.

```js
} else if (ACCOUNT_FIELD_TYPES.includes(field.type)) {
```
(with `const { ACCOUNT_FIELD_TYPES } = require('./constants');` at the top)

---

front/src/routes/integration/all/external-integration/config-page/index.js
```diff
@@ -447,9 +448,14 @@ class ExternalIntegrationConfigPage extends Component {
...
   connectOAuth = async field => {
-    this.setState({ oauthStatus: RequestStatus.Getting, oauthInvalidState: false });
+    this.setState({ oauthStatus: RequestStatus.Getting, oauthInvalidState: false, oauthInvalidUrl: false });
     const { selector } = this.props;
     const callbackPath = getOAuthCallbackPath(selector);
+    // an account_link provider never comes back to Gladys: the user approves it
+    // elsewhere and the integration notices on its own side. There is no
+    // redirect URI to declare, no state to carry across a round trip, and no
+    // callback tab to prime — the URL is opened as it was built.
+    const usesRedirect = field.type !== 'account_link';
```
Pierre-Gilles, line 458: **Nit: two opposite-polarity encodings of the same distinction.**

`connectOAuth` derives `usesRedirect` as `field.type !== 'account_link'` while `ConfigSchemaForm` derives it as `field.type === 'oauth2'`. For the two current types the predicates agree, but they classify any future `ACCOUNT_FIELD_TYPES` entry oppositely by default: this side would treat it as redirect-based and demand an anti-CSRF state, while `ConfigSchemaForm` would hide the redirect-URI block — a divergence invisible until the flow breaks.

A single shared predicate (e.g. `field.type === 'oauth2'` in both places, or a `USES_REDIRECT` set next to `ACCOUNT_FIELD_TYPES` in the shared constants) removes the trap.

---

server/lib/external-integration/manifest.schema.json
Pierre-Gilles, line 1: **Stale description on the `default` property** (line ~231, outside this diff, hence the file-level comment): it still says default is "not allowed for secret and oauth2 fields", although `validateConfigFieldDefault` now also rejects it for `account_link` and `section` fields.

A third-party integration developer reading this schema — which `validateManifest.js` calls "the exact mirror" of the code — would conclude `default` is legal on their `account_link` field, ship a manifest with it, and only discover the 422 (`config_schema[0].default: not allowed for account_link fields`) at install time.

---

Pierre-Gilles (commented):
Review of the `account_link` config field type (17 files). 6 findings posted inline, by severity:

1. **Functional bug** — `ConfigSchemaForm` drops the new `oauthInvalidUrl` prop, so the dedicated invalid-URL error message (added in all three locales) can never render.
2. **Spec/contract deviation** — the server still relays a caller-supplied `redirect_uri` to the integration for `account_link` fields, while the spec updated in this PR says it must be absent.
3–6. **Cleanup/consistency** — dead `connectionStatus` prop still passed by `ConfigTab`; inline type pair instead of the new `ACCOUNT_FIELD_TYPES` constant in `getConfigForFront`; two opposite-polarity `usesRedirect` predicates; stale `default` description in `manifest.schema.json`.

Also verified and found sound: the new `assertOpenableUrl` guard (it also closes the pre-existing `javascript:` URL hole for the oauth2 instance-redirect path), the conditional server-side `redirect_uri` validation, manifest validation for the new type (`placeholder`/`default`/`options`/`contact_schema` correctly rejected), the `not`/`enum` JSON-schema construct, the badge/message move from `ConfigField` to `ConfigTab`, localStorage priming correctly skipped for `account_link`, complete de/en/fr locales, and `relayOAuthCallback` correctly staying oauth2-only.

---

Pierre-Gilles:
@callemand Je ne sais pas si tu as vu les retours sur la PR + les conflits? :)

Pierre-Gilles:
@callemand Je vois que tu as force-push, il y a encore des conflits sur la branche

Pierre-Gilles:
Merged in https://github.com/GladysAssistant/Gladys/pull/2879

PR #2783, opened 2026-08-06
Add a per-feature setpoint step, driven by the device

callemand:
### Description

The setpoint step shown in the UI is decided per category (`SETPOINT_STEP_BY_CATEGORY` in `SetpointDeviceFeature`):
air conditioning steps by a whole degree, so a unit that exposes a half-degree setpoint (Mitsubishi, Daikin…) cannot
be driven to it. No integration has any way to express what its own device actually accepts.

This adds a nullable `step` column on `t_device_feature`, so a feature can carry the resolution its physical device
supports. It takes precedence, and the category defaults stay exactly as they are behind it:

```js
feature.step || SETPOINT_STEP_BY_CATEGORY[category] || 0.5
```

A feature that declares no `step` behaves exactly as before.

- migration + model: nullable `DOUBLE` `step`, refused when `<= 0` — a step of 0 would freeze the `+`/`-` buttons and
  a negative one would invert them, so `NULL` stays the only way to say "nothing declared"
- `setDiscoveredDevices`: `step` validated as a finite number > 0, so an integration can publish it
- `getDiscoveredDevices`: `step` joins the feature signature — without it a unit that starts declaring a half-degree
  step never flags `structure_changed`, the Discovery screen never offers "Update", and the step never reaches an
  already-created device
- `device.addFeature`: `step` is updatable; the nullable fields are normalized on both sides before comparing, so a
  stored `null` against an omitted key no longer makes every no-op payload look like an update (that branch called
  `.get()` on the `[affectedCount]` of a static `update()` and threw)
- `room.get` / `room.getBySelector`: served through a shared `ROOM_DEVICE_FEATURES_ATTRIBUTES` — the two lists were
  near-duplicates and had already drifted, and `room.get` now serves `last_value_string` as well
- `SetpointDeviceFeature`: consumes `feature.step`, and rounds each `+`/`-` on the finest grid of its two operands so
  a decimal step does not drift in floating point (`20.1 + 0.1`) and a `20.5` set from the physical remote is not
  snapped to a whole degree

**Scope**: the step is declared by the integration that knows the hardware, not entered by the user — it is a
capability of the device, not a preference. There is deliberately no `step` field in the device edition forms; an
internal integration sets it when it builds its features, an external one publishes it in its discovery.

The living spec (`docs/specs/external-integrations.md`) documents the field with the discovery rules. The SDK side
(typings + docs, no code change — features are published verbatim) is GladysAssistant/integration-sdk-js#23.

## Forum

Forum: https://community.gladysassistant.com/t/integration-externe-melcloud-home/10420

### Checklist


callemand:
Thanks @coderabbitai — two of the three findings were real and are fixed in bd8e25c:

- **Decimal steps drifting in floating point.** Right, and it only became reachable now that a device can publish its own step: `0.5` is exact in binary but `0.1` is not, so `20 + 0.1 + 0.1` saved `20.200000000000003`. Both `add` and `substract` now round to the precision the step itself carries.
- **`step = 0` accepted by the model, rejected by discovery.** Aligned: the model refuses `<= 0` too, so `NULL` stays the only way to say "no step declared". Covered by a new `device.addFeature` test.

I left the third one (`down` as a no-op) as is: no migration in this repository removes a column on `down` — of the 68 migrations, only the `createTable` ones do anything (`dropTable`), and every `addColumn` migration, including the most recent ones, has an empty `down`. Adding a `removeColumn` here would make this the single exception to that convention, on SQLite where the operation rebuilds the table. Happy to add it if a maintainer prefers.

callemand:
Second round, thanks @cursor / @coderabbitai.

**Fixed — the living spec.** Right, the discovery contract is specified in `docs/specs/external-integrations.md` and I had left `step` out of it (712a1ee). It now sits with the `category`/`type`/`unit` rules of `POST /discovered_device`: finite number > 0 or `400`, absent/`null` meaning the device declares nothing and the UI keeps its category default. I did not add it to the JSON example — that example publishes a read-only temperature *sensor*, where a setpoint step would be misleading.

**Left as is, with reasons:**

- **Scientific notation in `STEP_DECIMALS`** (`1e-7` → 0 decimals → rounds the increment away). Real in the abstract, unreachable here: JS only switches to exponent notation below `1e-6`, and this component renders temperature and current setpoints, whose `min`/`max` are whole degrees and amps. A device declaring a step of `1e-7` is not a device. @cursor read it the same way ("no need to chase scientific-notation cases for this UI"). One line to add if a maintainer disagrees.
- **`??` instead of `||`.** `||` is deliberate here: it also catches a `0` that predates the validation, where `??` would let it through and freeze the buttons. It matches the line right below it (`DEFAULT_VALUE_BY_CATEGORY[...] || 0`) too.
- **`areObjectsEqual` seeing `null` vs `undefined` as different in `addFeature`.** Correct, and as @cursor notes it is pre-existing for the other nullable fields (`unit`). Out of scope for this PR — worth its own change if the updates turn out to be noisy.

CI is red for reasons unrelated to the diff: GitHub Actions has been in a major outage since 15:22 UTC, every failed job here died on `Set up job` or was cancelled after 15-35 min waiting for a runner. Cypress and Front build did pass. Locally, `cd server && npm run coverage` is green apart from three failures that reproduce identically on `master`.

server/lib/device/device.addFeature.js
```diff
@@ -3,7 +3,7 @@ const { buildUniqueSelector } = require('../../utils/addSelector');
 const { areObjectsEqual, pick } = require('../../utils/objects');
 const db = require('../../models');
 
-const UPDATABLE_FIELDS = ['category', 'type', 'unit', 'min', 'max'];
+const UPDATABLE_FIELDS = ['category', 'type', 'unit', 'min', 'max', 'step'];
```
Pierre-Gilles, line 6: Adding `step` here routes every existing feature into the update branch, which is broken. After the migration every stored feature has `step: null`, while all current callers send payloads without a `step` key; `areObjectsEqual` compares with `===`, so `null !== undefined` makes a previously-equal feature "unequal" and it falls into the `else if` branch — where `db.DeviceFeature.update()` (static Sequelize update) returns `[affectedCount]`, so `updatedFeature.get({ plain: true })` throws `updatedFeature.get is not a function`. And since `pick()` yields `step: undefined`, which Sequelize ignores, the mismatch never converges: every subsequent call re-attempts the update.

The broken update branch is latent and pre-existing, but this line makes the normal no-op payload of every current integration hit it. Normalizing `undefined` → `null` before comparing (or a null-tolerant equality for optional fields) would keep the no-op path a no-op.

---

front/src/components/boxs/device-in-room/device-features/SetpointDeviceFeature.jsx
```diff
@@ -39,14 +42,14 @@ const SetpointDeviceFeature = ({ children, ...props }) => {
     const prevValue = isNullOrUndefined(props.deviceFeature.last_value)
       ? DEFAULT_VALUE_IN_CASE_EMPTY
       : props.deviceFeature.last_value;
-    updateValue(prevValue + SETPOINT_STEP);
+    updateValue(Number((prevValue + SETPOINT_STEP).toFixed(STEP_DECIMALS)));
```
Pierre-Gilles: This does more than the float-drift correction the comment above describes: it snaps the *current value* to the step grid. For a category whose default step is 1 (air conditioning, switch, EV charge) with no declared step, `STEP_DECIMALS` is 0, so a `last_value` of 20.5 — set from the physical remote — yields `Number((21.5).toFixed(0)) = 22` on `+` (old behavior: 21.5) and 20 on `-`. The user's half-degree setpoint is silently destroyed.

Deriving the precision from both operands (e.g. `Math.max(decimals(prevValue), STEP_DECIMALS)`) would fix the drift without snapping values the device itself reported.

---

front/src/components/boxs/device-in-room/device-features/SetpointDeviceFeature.jsx
```diff
@@ -24,8 +24,11 @@ const DEFAULT_VALUE_BY_CATEGORY = {
 };
 
 const SetpointDeviceFeature = ({ children, ...props }) => {
-  const SETPOINT_STEP = SETPOINT_STEP_BY_CATEGORY[props.deviceFeature.category] || 0.5;
+  const SETPOINT_STEP = props.deviceFeature.step || SETPOINT_STEP_BY_CATEGORY[props.deviceFeature.category] || 0.5;
   const DEFAULT_VALUE_IN_CASE_EMPTY = DEFAULT_VALUE_BY_CATEGORY[props.deviceFeature.category] || 0;
+  // a device can declare a step like 0.1, which drifts in binary floating point
+  // (20 + 0.1 + 0.1 = 20.200000000000003): stay on the grid the step defines
+  const STEP_DECIMALS = (`${SETPOINT_STEP}`.split('.')[1] || '').length;
```
Pierre-Gilles: Edge case: server-side validation accepts any finite step > 0, but a step ≤ 1e-7 stringifies in scientific notation (`` `${1e-7}` `` → `'1e-7'`), which contains no `'.'`, so `STEP_DECIMALS` is 0 and `add()` posts `Number((20 + 1e-7).toFixed(0)) = 20` — the buttons become a no-op. Either bound the accepted range in `setDiscoveredDevices`/the model (a sub-micro setpoint step has no physical meaning anyway), or derive the precision in a notation-proof way.

---

docs/specs/external-integrations.md
```diff
@@ -572,7 +572,7 @@ Actions are **integration-scoped**: a single rendering place, the Configuration
...
+→ `200 { "success": true, "count": 1 }`. Rules: prefixed `external_id`s (device **and** features), `category`/`type`/`unit` within Gladys's standard lists (`DEVICE_FEATURE_CATEGORIES`/`TYPES`/`UNITS`), max 2000 devices (discovery is fleet-wide: a network integration publishes one entry per client on the network, ~1000 on a large setup), `400` otherwise. Optional device-level field **`poll_frequency`** (values from the existing `DEVICE_POLL_FREQUENCIES`: 1 s to 60 s) to receive `device.poll` from the core scheduler once the device is created. Optional feature-level field **`step`** (finite number > 0, `400` otherwise), the resolution the **physical device** accepts for a setpoint — `0.5` for an air conditioner steppable by half a degree, which the dashboard's `+`/`-` buttons then honor. Absent or `null` means "nothing declared" and the UI keeps its per-category default: the step is the device's business, and no category default can be right for every unit of that category.
```
Pierre-Gilles: "Absent or `null` means 'nothing declared'" doesn't hold on the update path: in `device.create`'s update branch, `featureToUpdate = { ...feature }` with no `step` key is ignored by Sequelize's `instance.update()`, so an integration that stops declaring a step (key absent) and re-publishes keeps the old stored value forever — only an explicit `step: null` clears it. Either normalize absence to `null` on update, or amend this sentence to say that clearing a previously declared step requires an explicit `null`.

---

server/test/lib/external-integration/externalIntegration.setDiscoveredDevices.test.js
```diff
@@ -119,6 +119,22 @@ describe('externalIntegration.setDiscoveredDevices', () => {
...
+  it('should reject a non-positive or non-numeric step', async () => {
+    const zeroStep = buildDiscoveredDevice(service.selector);
+    zeroStep.features[0].step = 0;
+    await expectBadParameters([zeroStep], 'features[0].step: must be a positive number');
+    const stringStep = buildDiscoveredDevice(service.selector);
+    stringStep.features[0].step = '0.5';
+    await expectBadParameters([stringStep], 'features[0].step: must be a positive number');
+  });
+
+  it('should accept a valid step', async () => {
```
Pierre-Gilles: This only asserts the returned count — nothing in the suite reads `feature.step` back from `getDiscoveredDevices`, the created device, or `room.get`. A future refactor that drops `step` during normalization (a field whitelist, for instance) would still pass everything, and the end-to-end plumbing is precisely what this PR exists for. Worth asserting the round-trip here.

---

server/lib/room/room.get.js
```diff
@@ -19,6 +19,7 @@ const DEVICE_FEATURES_ATTRIBUTES = [
   'unit',
   'min',
   'max',
+  'step',
```
Pierre-Gilles: Not introduced by this PR, but it just demonstrated the trap: `DEVICE_FEATURES_ATTRIBUTES` is a near-duplicate of the list in `room.getBySelector.js` (which already drifts — it has `last_value_string`, this one doesn't), and every new feature column must be remembered in both files or one endpoint silently serves features missing the field. Extracting a shared constant would remove the trap for the next column.

---

Pierre-Gilles (commented):
The direction is right: a setpoint step is a capability of the physical device, and it belongs on the feature next to `min`/`max`/`unit` rather than in a per-category table in the front — with the nullable-column + fallback approach keeping every existing integration's behavior intact. That said, the review found one delivery gap in a file the PR doesn't touch, plus a few edge-of-contract issues (inline comments):

**`featureSignature` in `externalIntegration.getDiscoveredDevices.js` doesn't include `step`** — this is the most impactful finding and it's in a file outside the diff. The signature hashes `external_id/category/type/unit/min/max` only, so an integration that re-publishes a feature with a *different step* (or adds one where none existed) never sets `structure_changed`: the Discovery screen never shows the Update button, and the new step can never reach `t_device_feature` for an already-created device. Since a device-driven step is the whole point of the PR, `step` should be part of the structure signature.

Inline, from most to least severe:

- `device.addFeature`: adding `step` to `UPDATABLE_FIELDS` routes every existing no-`step` payload into the update branch (DB `null` vs payload `undefined`), which throws on `updatedFeature.get` — see comment.
- `SetpointDeviceFeature.jsx`: `toFixed(STEP_DECIMALS)` snaps fractional setpoints to whole numbers for integer-step categories (20.5 + 1 → 22), a regression beyond the float-drift fix; and steps ≤ 1e-7 (scientific notation) make the buttons a no-op.
- Spec vs `device.create`: an absent `step` can never un-declare a stored step on update, contradicting the "absent or null" sentence added to the spec.
- Tests: the "valid step" test never asserts the step actually round-trips.
- `room.get`/`room.getBySelector`: the duplicated attribute lists this PR had to patch twice deserve a shared constant.

Single-pass review; no adversarial verification pass was run on these findings.

---

callemand:
Thanks @Pierre-Gilles — all six findings held up when I traced them, including the two that were my own regressions. Fixed in 539b75e.

**`featureSignature` missing `step`** — you were right that this was the one that mattered: `step` was reaching `t_device_feature` and the dashboard, but never an *already-created* device, which is the only case the feature exists for. `step` is now part of the signature. The regression test (`should flag a step that appears on an already-created feature`) fails with the line removed and passes with it — I checked both ways rather than trusting it.

**`addFeature` routing every no-op into the broken branch** — confirmed end to end: stored `null` vs omitted key, `===` in `areObjectsEqual`, then `.get()` called on the `[affectedCount]` of a *static* `update()`. Normalizing the nullable fields on **both** sides of the comparison keeps a no-op a no-op (your suggestion; normalizing only the payload left the pre-existing `should update one feature but not the name/selector` test entering the branch, since its RAM fixture has no `unit`/`step` keys either), and the branch now reads the row back instead of calling `.get()` on a counter. Two tests cover the no-op and the real update.

**`toFixed(STEP_DECIMALS)` snapping 20.5 → 22** — my regression from the previous round, and the failure mode is exactly as you described. The sum is now rounded on the finest grid of its two operands, so a 20.5 reported by the device survives a whole-degree step. Verified on ten cases including yours (`20.5 + 1 = 21.5`, `20.1 + 0.1 = 20.2`).

**Scientific notation** — I had waved this one off last round on the grounds that a sub-micro setpoint is not a physical device. Your framing is better: the contract accepts it, so the hole is reachable by contract, not by hardware. `decimalsOf` now reads the exponent (`1e-7` → 7 decimals) rather than bounding the range server-side, which keeps the validation permissive.

**Spec vs `device.create`** — correct, `instance.update()` only writes the keys it is given. I amended the sentence rather than normalizing absence to `null` in `device.create`: that path is shared with `unit`/`min`/`max` and changing it for all of them is its own change, not this PR.

**Test asserting only the count** — fair, and the weakest part of the original diff. It now asserts the round-trip through `getDiscoveredDevices`, plus the "no step declared" case, plus `step` served by both `room` endpoints.

**Duplicated attribute lists** — extracted as `ROOM_DEVICE_FEATURES_ATTRIBUTES` in `utils/deviceQueryIncludes.js`, the module both files already imported. Note the side effect, in the additive direction: `room.get` now serves `last_value_string` too, which only `getBySelector` did — the drift you pointed at.

`cd server && npm run coverage`: 2883 passing. The only uncovered lines in the touched files are pre-existing (the two `throw`s of `addFeature`, `createdFeatures || []`, the `param.name` guard) — nothing on a changed line. Three failures reproduce identically on `master` on my machine (`install_service_dependencies`, `jsonBodyMiddleware` `EPIPE`, HomeKit `EADDRINUSE`).

PR #2787, opened 2026-08-07
Add port name declarations for manifest placeholder resolution

Pierre-Gilles:
### Description

This PR adds support for optional `name` fields on declared container ports, enabling manifest section texts to reference assigned host ports via `{{port:<name>}}` placeholders that are resolved by the frontend.

**Key changes:**

- **Backend validation**: Added `SUB_CONTAINER_PORT_NAME_REGEX` constant (`[a-z0-9_]{2,20}`) to validate port names. Port names must be unique across the entire manifest (not per-container) since placeholders carry no container prefix.
- **Placeholder validation**: New `validateSectionPortPlaceholders()` function ensures all `{{port:<name>}}` references in section texts (config_schema, action fields, contact_schema) reference declared port names. Unknown references are rejected at validation time to prevent unresolved tokens on screen.
- **Port state tracking**: The `getSubContainersState()` endpoint now includes the `name` field in port responses, allowing the frontend to map declared names to assigned host ports.
- **Frontend resolution**: New `resolveManifestPlaceholders()` utility resolves both `{{gladys_host}}` (browser hostname) and `{{port:<name>}}` (assigned host port) placeholders in section texts. Tokens with unassigned ports remain as-is for debugging.
- **Schema updates**: Updated manifest JSON schema to document the new optional `name` field on ports.

**Use case**: Enables integrations like OCPP to display dynamic URLs in configuration sections (e.g., "Point your charge point to `ws://{{gladys_host}}:{{port:ocpp}}`") that resolve to the actual assigned host port.

### Checklist

- [x] Tests pass: Added comprehensive unit tests for port name validation, placeholder validation, and resolution
- [x] Linter and prettier pass
- [x] No breaking changes: The `name` field is optional; existing manifests without port names continue to work

https://claude.ai/code/session_01LAM9omi7idv8G4wyasE4gv

## Summary by CodeRabbit

* **New Features**
  * Added support for named container ports in integrations.
  * Added `{{gladys_host}}` and `{{port:<name>}}` placeholders in integration labels, descriptions, configuration fields, and actions.
  * Integration details now display port names for easier browser access.
* **Bug Fixes**
  * Added validation for invalid, duplicate, unknown, or restricted port placeholder references.
  * Unavailable port references remain safely unresolved instead of causing errors.
* **Tests**
  * Expanded coverage for named ports, placeholder resolution, and validation rules.


Pierre-Gilles:
Thanks for the review — the three follow-ups, in order:

**1. Store schema sync (`manifest.schema.json`).** Correct and worth flagging loudly: port objects are `additionalProperties: false`, so until the canonical schema in `GladysAssistant/integration-store` gains the same `name` property, the indexer rejects any published manifest declaring one — which is exactly the OCPP case this feature exists for. Nothing to change in this repo (the spec already names the store as the canonical owner with a vendored copy here), but a companion PR there is a prerequisite before store installs can use `{{port:<name>}}`. Installs by `repo_url` and dev mode are unaffected, since they only go through the server's `validateManifest`.

**2. "indexer and server" wording.** Keeping it deliberately. That phrasing is not a claim about JSON Schema's expressiveness — it is the contract the vendored copy already states in `validateManifest.js`: *"a manifest accepted by the indexer must always install here, and vice versa"*. The spec makes the same claim for other rules JSON Schema cannot express either (manifest-wide `containers[].name` uniqueness, `source`/`options` mutual exclusion). Softening it to "the server only" would tell the store repo it may skip the check, and then a manifest the indexer accepts would fail to install — the exact divergence the contract exists to prevent.

**3. Stale `placeholderPorts`.** Fair, and the spec was the thing that was wrong — fixed in fbd249f. The Configuration screen fetches the detail on mount, and `status-changed` patches the status only, so a host port allocated under an already-open screen is not picked up before a reload. The spec now says that instead of "resolves as soon as the port is allocated". I did not add a re-fetch: ports are allocated at the sub-container's first start, which in practice precedes the user opening the screen, and the token left visibly unresolved is the intended honest fallback.

---

Pierre-Gilles:
Both correctness findings were real and are fixed in 7f41f96.

**Inherited keys in the port map.** Confirmed reachable: `constructor` and `__proto__` both match the allowed port-name pattern `[a-z0-9_]{2,20}`. On a plain object an unassigned `{{port:constructor}}` rendered `function Object() { [native code] }` instead of staying unresolved, and `portsByName['__proto__'] = 42115` never created an entry, so a port genuinely named that could never resolve. Now `Object.create(null)` for the map plus an own-property check on read. Verified across the four cases (unassigned reserved name stays raw, assigned reserved name resolves, normal name unaffected, empty map stays raw).

**`{{port:<name>}}` in `contact_schema`.** Also real — the reduced non-admin view carries no `containers`, so the token resolved for an admin and stayed raw for everyone else. I took the second of your two options rather than the first: **validation now rejects it** with `contact_schema[i].description.en: {{port:x}} is not available in the per-user contact schema`. Widening the reduced view to carry assigned host ports would trade a deliberate privacy boundary (C.5 excludes `containers`, `docker_image` and `webhooks` from what a non-admin sees) for a cosmetic gain, and role-dependent rendering of the same declarative block is worse than saying the feature is not available there. `{{gladys_host}}` stays allowed in `contact_schema` — the browser resolves it identically for every role. The frontend no longer threads port assignments into that card, since they can no longer occur.

**Test coverage.** Added the two `contact_schema` cases you asked for (declared-but-refused reference, and an accepted `{{gladys_host}}`-only section), on a send-only communication manifest so the `contact_schema` branch is genuinely exercised. Spec updated in the same commit. Full external-integration suite: 516 passing, eslint and prettier clean on server and front.

---

PR #2788, opened 2026-08-07
fix(external-integration): preserve line breaks in integration-published text

Pierre-Gilles:
### Description

An external integration currently has no way to display a multi-line message. Two things cumulate:

1. the core renders the text an integration publishes as an **escaped text node** — deliberate, and unchanged here: the container is unaudited third-party code, injecting it as markup would be an XSS in Gladys' own admin UI;
2. HTML **collapses `\n` into a space** by default.

So an integration returning a structured result from an action gets one run-on line. The workaround in the wild is padding the message with bullet characters (`•`) to keep it readable.

This renders those blocks with `white-space: pre-wrap` — the idiom already used for the container logs (`logs-page/style.css`) and the chat bubbles (`routes/chat/style.css`) — so a `\n` in the message becomes a line break on screen. Three details come with it:

- `overflow-wrap: anywhere`, so a long unbroken token (a URL, an id) wraps instead of overflowing the card. `pre-wrap` alone does not break it;
- a height cap with scroll on action results only: nothing bounds their length, the integration decides what it returns, and a verbose one would otherwise push the rest of the screen down;
- `tabIndex` on those scroll containers — a bare `<div>` with `overflow-y: auto` is not focusable in Chromium, so a keyboard-only user could not scroll a long result.

Applied to **every** free-form text the core renders on an integration's behalf, so a developer does not have to guess which field honors line breaks:

| Where | File |
| --- | --- |
| Action result (success and error) | `config-page/ActionsCard.jsx` |
| Action description | `config-page/ActionsCard.jsx` |
| `config_schema` field and `section` descriptions | `config-page/ConfigSchemaForm.jsx` |
| `connection_status` message (config screen) | `config-page/ConfigSchemaForm.jsx` |
| `connection_status` message (supervision screen) | `supervision-page/SupervisionCard.jsx` |
| Manifest description (install screen) | `install-page/index.js` |

Left out on purpose: the catalog cards, where the manifest description is a one-line tagline in a grid and uneven heights would be a regression; and the degraded-transport tooltip, which is a native `title` attribute already honoring `\n` without CSS. Both exclusions are now named in the spec so a later "apply everywhere" pass does not undo them.

No protocol change, no escaping change — still plain text, still no HTML or Markdown interpreted. `docs/specs/external-integrations.md` now states the contract explicitly so the question does not come back.

## Forum

Forum: https://community.gladysassistant.com/t/sauts-de-ligne-lors-de-laffichage-dinformation-dans-une-integration-externe/10462

### Checklist

- [x] Tests pass: `cd server && npm run coverage` (Codecov requires 100% coverage on changed lines) and Cypress (`npm run cypress:run`) if the UI changed — no server code touched; the change is CSS-only on the frontend, no behavior a Cypress spec asserts on
- [x] Linter and prettier pass on both front and server (`npm run eslint`, `npm run prettier`) — prettier verified locally against the major version the repo pins (1.x); eslint could not be run locally (front dependencies are not installed in this environment), but CI's Front test job covers it and is green
- [x] No undocumented breaking change — purely additive, an integration sending single-line messages renders exactly as before

https://claude.ai/code/session_01TAhGDNtnbF4rfbzJLmXZRh


Pierre-Gilles:
Review points addressed in `12fbbf2`. One of them I did differently than suggested, hence this note.

**Duplicated wrapping rules** (@cursor) — fixed, but not with `composes`. `composes` appears nowhere in this codebase today, and a silent failure there degrades to unstyled text rather than a build error, which I could not verify locally. Instead the two classes are applied side by side at the call site: `.integrationText` keeps `white-space` / `overflow-wrap` as the single source of truth, and `.resultScroll` now carries only the height cap. Same one-definition property, no new build-level mechanism.

**Keyboard access to the scrollable result** (@coderabbitai) — valid, fixed. A bare `<div>` with `overflow-y: auto` is not focusable in Chromium, so a keyboard-only user could not scroll a result longer than the cap. `tabIndex` on both result blocks.

**Spec over-promising** (@coderabbitai, @cursor — same point) — fixed. The contract paragraph now names the two deliberate exclusions, catalog cards and the degraded-transport `title` tooltip, so a later "apply everywhere" pass does not reintroduce uneven card heights.

---

PR #2932, opened 2026-08-17
Add an on-demand Claude autofix triggered by a /claude PR comment

Pierre-Gilles:
### Description

Writing **"Please fix PR conflicts"** on a pull request (as on #2925) reaches nothing today, and it is structural rather than a one-off bug.

The only Claude automation in this repository is `claude-scheduled-autofix.yml`, whose triggers are `schedule` and `workflow_dispatch` — **no `issue_comment`**, so no comment can ever start a session. The only comment-driven workflow is `cursor-automation-webhook.yml`, which recognises `/cursor review` and triggers Cursor, not Claude.

Even had the autofix been event-driven, its selection rules exclude that request three times over:

1. a PR in `mergeable_state: dirty` is **explicitly skipped** (`skip 'merge conflict'`) — precisely the case being asked about;
2. only `cursor[bot]` and `coderabbitai[bot]` comments count as actionable feedback (`REVIEW_BOT_LOGINS`);
3. the maintainer's own login is in `AUTOFIX_ACTOR_LOGINS`, so his comments are read as *"handled"* markers rather than as requests — a comment of his in a thread even marks that whole thread as done.

**What this PR adds:** `claude-on-demand-autofix.yml`, the missing on-demand channel. A comment whose line starts with the `/claude` command, followed by an instruction, fires a cloud session with that comment as its work order — "fix the merge conflicts with master", "the readInitialDeviceStates test fails on CI, have a look", "the alarm badges render green even when the fault is active", "why did you make the door feature read-only here?".

It is deliberately a **generic feedback channel**, not a conflict-resolution mode: merges, tests, UI details and plain questions all take the same path. It listens on the three places a review actually happens — the PR conversation, a comment on a line of the diff (the file and line become part of the request), and the body of a submitted review.

It reuses the routine, repository variable and secret of the cron, so there is **no new setup**: the payload now carries a `mode` field, and the routine prompt gained the matching procedure.

Differences from the cron, all deliberate:

- a **merge conflict**, a **draft** PR and the `claude:autofix-exhausted` / `claude:autofix-skip` labels no longer block — those labels stop the cron, and an explicit human request outranks the automation's own stop conditions;
- commits carry an `Autofix-Request: <comment id>` trailer instead of `Autofix-Pass: <n>`, so asking by hand **never eats into the per-PR `MAX_PASSES` budget**.

### Access control

Only the logins in `COMMAND_ALLOWED_LOGINS` (currently `Pierre-Gilles`) may run the command, matched case-insensitively.

An `author_association` check alone would **not** be equivalent, and the first version of this PR got that wrong: `MEMBER` is granted to every member of the GladysAssistant organisation, and **8 of this repository's 9 collaborators hold the `triage` role, which carries no push access**. Gating on the association would therefore have let them fire a session that pushes commits to `claude/` branches and spends the routine owner's claude.ai allowance — handing them, indirectly, a write access GitHub deliberately withholds. The association is still checked as a *second* barrier (and at job level, so a drive-by command from outside the org never allocates a runner), so an allow-listed login that leaves the organisation stops working on its own.

This is far stricter than `/cursor review`, which the PR author may also run — deliberately: triggering Cursor costs a Cursor run, whereas triggering Claude pushes commits under the routine owner's identity and draws down their subscription.

### Anti-loop

The command must open its own line, so quoting it never fires a session. Two guards then separate a session's *answer* from a new command, which matters because **sessions post under the routine owner's identity — a `User`, and the very login that types the command** — so neither the bot-author check nor the allow-list can tell them apart:

- every on-demand answer, from the workflow or from the session, must carry `<!-- Autofix-Request: <comment id> -->`, and the workflow skips any body whose *unquoted* text holds an `Autofix-` marker;
- the prompt forbids a session from starting a reply line with the command.

Quoted lines are excluded from the marker check on purpose, so GitHub's "Quote reply" cannot swallow a command written under a quoted answer. The workflow also listens for `/claude` only, never `@claude`, so re-enabling the GitHub App mode could not double-fire on the same comment.

When a request cannot be served — closed PR, fork PR, non-`claude/` branch, bare command, failed fire — the workflow **answers where the command was written**, replying in the review thread for an inline command. An explicit command that does nothing visible is exactly the failure mode this workflow exists to remove.

**Security:** like `cursor-automation-webhook.yml`, it never checks out or executes PR head code — it reads PR metadata through the API and POSTs to the routine API. The comment body only ever transits through shell variables and a file, never interpolated into code. In the routine prompt, `instruction` is the single field readable as a work order, and its authority stays bounded to that PR and its branch; everything else remains untrusted data.

### Deploying this (order no longer matters, but read this)

The prompt on the routine's page and the payload in the workflow deploy **separately and never atomically** — one lives in a web UI, the other in git. The first revision of this PR ignored that and instructed copying the prompt *before* merge, while making the prompt reject any payload without a `mode` field. Since the cron on `master` does not send one, that would have stopped every scheduled session on an unknown mode and **silently killed the live 3-hour cron**. Thanks to @cursor for catching it.

A missing `mode` is now read as `"scheduled"`, so the prompt is backward compatible with the payload in production and either order is safe. `/claude` simply stays inert until the prompt is copied to [claude.ai/code/routines](https://claude.ai/code/routines) from `.github/CLAUDE_AUTOFIX_ROUTINE_PROMPT.md`. The rule is now documented for the next mode added: teach the prompt the new payload first, deploy it, then ship the workflow that sends it.

### Checklist

- [x] Tests pass — no application code is touched (`.github/` only), so the server/front suites and Cypress are unaffected. The gate logic was instead exercised directly: both `run:` blocks were extracted from the YAML and run under a stubbed `gh`, **40/40 cases passing** — accepted commands (all three comment kinds, later-line and multi-line bodies, case-insensitivity), ignored ones (plain English, quoted command, mid-line command, bot author, marker-carrying session reply, `/claudette`), the quoted-marker regression (a command under a quoted answer still fires), access control (**each of the 8 triage collaborators refused despite carrying `MEMBER`**, case-insensitive match, no prefix match on a similar login, allow-listed login refused once no longer a member), refusal routing (inline command answered in its review thread, review body and conversation falling back correctly, every notice carrying the loop-guard marker), and payload plumbing (verbatim instruction, valid JSON, a body full of quotes/backticks/`$(…)` surviving as data).
- [x] Linter and prettier pass on both front and server — unchanged by this PR. `.github/*.md` is outside prettier's configured scope (`front`/`server` lint `.js`/`.jsx`/`.json` only); `CLAUDE_AUTOFIX.md` already failed `--check` on `master`, so it is left as it was rather than reformatted in this diff. Both workflow files parse as valid YAML and both `run:` blocks pass `bash -n`.
- [x] No undocumented breaking change — the cron's behaviour is unchanged apart from the added `mode: "scheduled"` payload field. `.github/CLAUDE_AUTOFIX.md` was restructured to cover both workflows, and its "Known limits" section documents the on-demand ones honestly: no budget cap (a human asking is the throttle), the per-PR concurrency group serialising the *fires* rather than the sessions (which outlive their job), and the fact that answering inside a review-bot thread marks it handled for the cron, since the cron keys that on the replying login rather than on a marker.

## Summary by CodeRabbit

- **New Features**
  - Added on-demand autofix requests through `/claude` commands in pull request comments and reviews.
  - Provides request validation, status reactions, launch notifications, and result summaries.
  - Serializes autofix requests per pull request to prevent conflicting runs.
  - Clearly separates scheduled autofix processing from on-demand requests.

- **Bug Fixes**
  - Prevents unsupported, unauthorized, malformed, or unsafe requests from starting automation.
  - Blocks bot loops and restricts execution to eligible pull requests and branches.

- **Documentation**
  - Expanded guidance for workflows, safeguards, limits, setup, and known constraints.


Pierre-Gilles:
All six findings verified against the code and fixed in 842c609. Two were real defects, not nits.

**1. The prompt would have killed the live cron** (`CLAUDE_AUTOFIX_ROUTINE_PROMPT.md`) — the important one. It rejected any payload without a `mode` field, and the cron on `master` does not send one. Following this PR's own instruction to copy the prompt *before* merge would have stopped every scheduled session on an unknown mode, silently. A missing `mode` now reads as `"scheduled"`. The underlying rule is now documented, since it will recur: prompt and payload deploy separately and never atomically, so the prompt must stay backward compatible with the payload already in production. The PR description has been corrected too — it was the thing giving the dangerous order.

**2. The loop guard did not guard the loop** (`claude-on-demand-autofix.yml`) — both halves of the diagnosis were right. Sessions post under the routine owner's identity, a `User`, so `COMMENT_AUTHOR_TYPE = Bot` never matched them; and `Autofix-Request` existed only as a *commit trailer*, never in a reply body, so the marker grep never matched either. A reply restating the request could fire another session, with no pass budget to stop it. The prompt now requires `<!-- Autofix-Request: <comment_id> -->` in the reply body and forbids starting a reply line with the command.

One addition beyond what was asked: the guard now ignores **quoted** lines. GitHub's "Quote reply" copies the marker into the quote, so the guard as specified would have silently swallowed a command typed underneath a quoted answer. There is a regression test for it.

**3. Replies now go where the command was written.** `refuse()` and the fire-failure notice reply in the review thread when the command came from a diff line, with a fallback to the PR conversation for review bodies and for threads GitHub refuses.

**4. Review bodies are no longer silent.** Reviews have no reaction endpoint, so 👀 and 🚀 were both no-ops there — that path now posts a short "on it" comment, and the doc's claim that reactions always acknowledge has been qualified.

**5. Unauthorized actors are filtered before a runner is allocated,** via an `author_association` test in the job-level `if`. The login allow-list stays in the script as the real access control, so it is not duplicated. Consequence, now documented: authors outside the organisation get no 👎, because they never reach the script.

One interaction worth flagging, which came out of fix 3 rather than from the review: answering inside a review thread marks that thread handled *for the cron*, since the cron keys "handled" on the replying login (`github-actions[bot]` is in `AUTOFIX_ACTOR_LOGINS`) rather than on a marker. So a bare command typed as an inline reply under a review-bot finding, then refused, drops that finding from the cron's selection. Self-correcting in practice, and recorded in "Known limits".

Gate coverage went from 34 to 40 cases; all pass.

---

PR #2951, opened 2026-08-18
fix(mdns): mediated discovery with multiple mDNS services

valentinhttr:
## Summary

- browse every mDNS service declared by an external integration during a mediated mDNS scan
- merge the raw results while preserving the existing host API and SDK contract
- document the multi-service behavior and add a regression test

## Why

The manifest already accepts several `network_discovery` entries with `type: "mdns"`, but `runNetworkDiscoveryScan` selected only the first matching entry. Integrations for devices exposing several DNS-SD services, such as Apple TV (`_airplay._tcp` and `_companion-link._tcp`), could therefore never receive the other declarations from the host-network core.

This is especially visible on routed networks: the bridge integration receives the first mediated announcement but cannot recover the missing services with its own direct mDNS query.

## Testing

- targeted network discovery suite: 26 passing
- `npm run prettier-check`
- ESLint on all modified server JavaScript files
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Network discovery now scans all mDNS services declared by an integration and combines results into one response.
  - mDNS discovery no longer requires separate service selection.
  - Results preserve declared service order and combine duplicate host addresses.
  - Duplicate service declarations are queried only once.
  - Other discovery types retain their existing behavior.
  - Only one discovery scan runs at a time.

- **Documentation**
  - Clarified mDNS scanning behavior and scan constraints in the API and integration specifications.


server/lib/external-integration/networkDiscovery/networkDiscovery.runScan.js
```diff
@@ -96,7 +97,10 @@ async function runNetworkDiscoveryScan(
       return await this.scanUdpActiveBroadcast({ port, payload, timeoutMs });
     }
     if (type === 'mdns') {
-      return await this.scanMdns({ service: capture.service, timeoutMs });
+      const results = await Promise.all(
+        captures.map((declaredCapture) => this.scanMdns({ service: declaredCapture.service, timeoutMs })),
+      );
+      return results.flat();
```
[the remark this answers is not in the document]
valentinhttr: Implemented in 1a6987cb. All declared mDNS services now share a single `multicast-dns` client while still sending one PTR query per service for compatibility with existing devices. The single-service `scanMdns({ service })` form remains supported, results preserve manifest order, and the regression suite covers both single- and multi-service discovery (27 targeted tests passing, with 100% line coverage across the network-discovery files).

server/lib/external-integration/networkDiscovery/networkDiscovery.scanMdns.js
```diff
@@ -70,10 +75,12 @@ async function scanMdns({ service, timeoutMs, mdnsOptions }) {
...
     mdns.destroy(resolve);
   });
-  return [...instances.values()].map((instance) => ({
-    ...instance,
-    addresses: instance.host ? addressesByHost.get(instance.host) || [] : [],
-  }));
+  return serviceNames.flatMap((serviceName) =>
+    [...instancesByServiceName.get(serviceName).values()].map((instance) => ({
+      ...instance,
+      addresses: instance.host ? addressesByHost.get(instance.host) || [] : [],
```
Pierre-Gilles: `addressesByHost` accumulates with `push` and never dedupes, and now that one socket receives one response per declared service, the *exact* case this PR targets duplicates the addresses.

A real Apple TV answers both PTR queries from the **same** host (`Apple-TV.local`), and each response carries that host's A record in `additionals`. Both records land in `addressesByHost.get('Apple-TV.local')`, so both merged instances come back with the address twice:

```json
[
  { "name": "Living Room._airplay._tcp.local",        "host": "Apple-TV.local",
    "addresses": ["192.168.1.50", "192.168.1.50"], "port": 7000,  "txt": [] },
  { "name": "Living Room._companion-link._tcp.local", "host": "Apple-TV.local",
    "addresses": ["192.168.1.50", "192.168.1.50"], "port": 49152, "txt": [] }
]
```

(I reproduced this by replaying the two packets through the new handler.)

The duplication is latent on `master` too — a responder re-announcing inside the scan window already produced it — but with a single service the extra A record was the exception; with N services on one host it is the rule. Integrations get `addresses` as a raw list and will iterate it, so this turns into duplicate connection attempts per device.

The test doesn't catch it because its two services resolve to two different hosts (`airplay.local` / `companion.local`), which isn't how a multi-service device announces itself. Suggest pointing both SRV records at a single target in the fixture.

Fix is a one-liner at collection time:

```js
} else if (record.type === 'A' || record.type === 'AAAA') {
  if (!addressesByHost.has(record.name)) {
    addressesByHost.set(record.name, new Set());
  }
  addressesByHost.get(record.name).add(record.data);
}
```

with `addresses: instance.host ? [...(addressesByHost.get(instance.host) || [])] : []` here.

---

server/lib/external-integration/networkDiscovery/networkDiscovery.scanMdns.js
```diff
@@ -3,22 +3,23 @@ const multicastDns = require('multicast-dns');
...
  * @returns {Promise<Array>} Resolve with [{ name, host, addresses, port, txt }].
  * @example
  * const results = await gladys.externalIntegration.scanMdns({ service: '_hue._tcp', timeoutMs: 5000 });
  */
-async function scanMdns({ service, timeoutMs, mdnsOptions }) {
-  const serviceName = `${service}.local`;
-  // instance name -> { name, host, addresses, port, txt }
-  const instances = new Map();
+async function scanMdns({ service, services, timeoutMs, mdnsOptions }) {
+  const serviceNames = (services || [service]).map((declaredService) => `${declaredService}.local`);
```
Pierre-Gilles: `serviceNames` can contain duplicates, and the new code no longer tolerates them.

Nothing rejects the same mDNS service declared twice: `manifest.schema.json` has no `uniqueItems` on `network_discovery`, and `validateNetworkDiscoveryEntry` only dedupes *ports within one entry*, never entries against each other. So this manifest validates today:

```json
"network_discovery": [
  { "type": "mdns", "service": "_hue._tcp" },
  { "type": "mdns", "service": "_hue._tcp" }
]
```

On `master` it was harmless — `find` took the first entry. With the merge, `serviceNames` is `['_hue._tcp.local', '_hue._tcp.local']`, so:

- `new Map(serviceNames.map(...))` collapses to one bucket (last wins),
- the `forEach` sends the same PTR query twice,
- and the closing `flatMap` walks `serviceNames`, so it iterates that one bucket twice — **every instance is returned twice**.

Confirmed by replay: a single Hue bridge comes back as two identical entries.

Cheapest fix is to dedupe here:

```js
const serviceNames = [...new Set((services || [service]).map((s) => `${s}.local`))];
```

Adding `"uniqueItems": true` to the `network_discovery` schema array (or a cross-entry check in `validateNetworkDiscoveryEntry`) would be the belt-and-braces version, but it only helps new manifests — already-installed ones keep their stored manifest, so the runtime dedupe is the one that matters.

---

server/lib/external-integration/networkDiscovery/networkDiscovery.runScan.js
```diff
@@ -53,10 +53,11 @@ async function runNetworkDiscoveryScan(
     );
   }
   const declaredCaptures = (service.manifest && service.manifest.network_discovery) || [];
-  const capture = declaredCaptures.find((entry) => entry.type === type);
-  if (!capture) {
+  const captures = declaredCaptures.filter((entry) => entry.type === type);
+  if (captures.length === 0) {
     throw new ForbiddenError(`network_discovery: capture type ${type} is not declared in the manifest`);
   }
+  const capture = captures[0];
```
Pierre-Gilles: Not asking you to widen this PR — just flagging that the same first-match bug still bites `udp-active-broadcast`, and there it produces a wrong **403**, not just a missing result.

`capture` is now `captures[0]`, and the authorization check below reads `capture.ports`:

```js
if (!capture.ports.includes(port)) {
  throw new ForbiddenError(`network_discovery: port ${port} is not declared in the manifest`);
}
```

With two `udp-active-broadcast` entries (allowed: `MAX_NETWORK_DISCOVERY_ENTRIES` is 5, and `MAX_UDP_BROADCAST_PORTS` is per entry, so a manifest can legitimately declare 10 ports across two entries)…

```json
[ { "type": "udp-active-broadcast", "ports": [9999] },
  { "type": "udp-active-broadcast", "ports": [20002] } ]
```

…a scan on port `20002` is rejected as undeclared even though the user approved it on the install screen. Same for `udp-broadcast`, where only the first entry's ports get listened on.

The union is the natural read of the manifest as an authorization contract, and it's small:

```js
const declaredPorts = captures.flatMap((entry) => entry.ports || []);
```

used for both the 403 check and the `scanUdpBroadcast({ ports })` call. Cursor already noted SSDP is in the same spot and fine to defer — this one seems worth at least an issue, since a 403 on an approved port is harder to diagnose in the field than a missing announcement.

Minor while here: for `type === 'mdns'`, `capture` is now dead — the branch uses `captures.map(...)`. Moving `const capture = captures[0]` into the non-mDNS paths (or naming it `firstCapture`) would make it obvious that first-match is deliberate for the remaining types rather than leftover.

---

Pierre-Gilles (commented):
Reviewed the diff plus the surrounding manifest validation, the install screen and the spec. The premise is right and the fix is in the correct place: `network_discovery` has always accepted several `mdns` entries, `runNetworkDiscoveryScan` only honoured the first, and the bridge had no way to recover the rest. Sharing one `multicast-dns` client across the declared services (after Cursor's earlier note) is the right shape — it keeps the socket cost bounded, and the per-service PTR queries stay compatible with responders that dislike multi-question packets.

A few things I checked that are fine:

- The `MDNS_SERVICE_REGEX` (`^_[a-z0-9-]+\._(tcp|udp)$`) rules out DNS-SD subtypes, so the `endsWith` bucketing in `scanMdns` can't mis-attribute an instance between two declared services — the suffix-overlap worry doesn't apply.
- `MAX_NETWORK_DISCOVERY_ENTRIES = 5` caps the fan-out at five PTR queries per scan; no new unbounded cost.
- `NetworkDiscoverySummary.jsx` already renders one line per entry, so two `mdns` declarations show as two lines on the install screen — the authorization contract stays truthful with no front change.
- Ordering by manifest rather than by network response is a good call, and the 25 ms delay in the new test is a genuine check of it rather than a coincidence.

Three findings inline, in rough order of how much they matter:

1. **Duplicate `addresses` on the very case this PR is for** — a real Apple TV answers both queries from one host, so the A record arrives twice and both merged instances come back with the address listed twice. The fixture uses two distinct hosts, which is why it passes. One-line `Set` fix.
2. **Duplicate mDNS declarations now duplicate every result** — nothing validates entry uniqueness, and the closing `flatMap` walks `serviceNames` rather than the buckets. Harmless on `master` (first-match), a regression here.
3. **`udp-active-broadcast` still first-match** — out of scope, but the same bug returns a 403 on a port the user actually approved, which I'd rather see tracked than lost.

(1) and (2) are both small and local; with those addressed this looks good to me. I reproduced both by replaying the record streams through the new handler rather than inferring them from reading.

---

server/lib/external-integration/networkDiscovery/networkDiscovery.scanMdns.js
valentinhttr: Confirmed and fixed in `5196506` — thanks, this is exactly the case the PR exists for and I had it backwards in the fixture.

`addressesByHost` now collects into a `Set` and is spread at return time:

```js
} else if (record.type === 'A' || record.type === 'AAAA') {
  if (!addressesByHost.has(record.name)) {
    addressesByHost.set(record.name, new Set());
  }
  addressesByHost.get(record.name).add(record.data);
}
```

```js
addresses: instance.host ? [...(addressesByHost.get(instance.host) || [])] : [],
```

I also took the fixture suggestion: both SRV records in `should browse multiple declared services through one mDNS session` now target a single `Apple-TV.local`, which is how a real multi-service device announces itself. That is what made the bug invisible — with two distinct hosts each bucket only ever saw one A record.

Checked the new fixture actually catches it: against the pre-fix `scanMdns` the test fails with `addresses: ['192.168.1.50', '192.168.1.50']` on both instances, and passes with the `Set`.

Good catch on the latent-on-`master` framing too — a re-announcing responder could already produce it with one service, it just moved from exception to certainty here.

---

server/lib/external-integration/networkDiscovery/networkDiscovery.scanMdns.js
valentinhttr: Right, and I agree it's a regression I introduced rather than a pre-existing one — `find` made it a non-issue on `master`. Fixed in `5196506` with the runtime dedupe you suggested:

```js
const serviceNames = [...new Set((services || [service]).map((declaredService) => `${declaredService}.local`))];
```

I checked your premise before applying it, since the fix is only worth having if a duplicate really can reach `scanMdns`: `manifest.schema.json` has `minItems`/`maxItems` on `network_discovery` but no `uniqueItems`, and `validateNetworkDiscoveryEntry` receives one entry at a time (`entry, index, errors`) — the `seenPorts` `Set` is scoped inside a single `udp-broadcast` entry, so nothing compares entries against each other. Two identical `mdns` entries validate.

Added a regression test (`should return one result per instance when a service is declared twice`) that scans `['_hue._tcp', '_hue._tcp']` and asserts both halves of what you described: `queryCount === 1` (one PTR query, not two) and a single result. Against the pre-fix code it fails on `expected 2 to equal 1`.

On the belt-and-braces part: I left the schema alone deliberately, for the reason you gave — installed integrations keep their stored manifest, so `uniqueItems` would not help anyone already affected, and the runtime dedupe makes the duplicate harmless rather than an install-time error. Happy to add it in a follow-up if you'd rather have the manifest rejected outright, but that's a behaviour change for existing integrations and felt like the wrong call to slip into this PR.

---

server/lib/external-integration/networkDiscovery/networkDiscovery.runScan.js
valentinhttr: Agreed on both the diagnosis and on keeping it out of this PR. I verified the scenario holds — `MAX_NETWORK_DISCOVERY_ENTRIES` is 5 and the `ports` cap in `validateNetworkDiscoveryEntry` is per entry, so two `udp-active-broadcast` entries totalling 10 ports validate fine, and the 403 then reads only `captures[0].ports`.

Tracked in #2970, with the manifest example, the union fix you suggested (`captures.flatMap((entry) => entry.ports || [])` for both the 403 check and `scanUdpBroadcast`), and a note that `ssdp` is in the same spot with lower impact. Left this PR scoped to mdns.

Took the minor point in `5196506`: `capture` is now `firstCapture`, with a comment saying first-match is deliberate for the remaining types. It reads much better next to the `captures.map(...)` in the mdns branch — it was genuinely ambiguous whether that variable was leftover.

---

Pierre-Gilles:
Merged in https://github.com/GladysAssistant/Gladys/pull/3002

PR #2953, opened 2026-08-18
System: advertise Gladys on the local network with mDNS

William-De71:
- Sends goodbye packets on shutdown (cache-flush bit correctly unset, per
  RFC 6762) so clients drop the record immediately instead of waiting for the
  TTL to expire.
- The hostname is stored in the `MDNS_HOSTNAME` system variable and sanitized
  server-side, falling back to `gladysassistant` if invalid.

Tested on a real network: `avahi-resolve -n gladysassistant.local` returns the
host IP, and the A/SRV/TXT records are visible from an independent mDNS client.

## Forum
https://community.gladysassistant.com/t/affichage-adresse-ip/9892

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added mDNS configuration to System Settings, including hostname validation, save status, errors, and the resulting `.local` URL.
  * Added English, French, and German translations for the new settings.
  * Gladys now advertises itself on the local network through mDNS.
  * System information now includes the detected local IP address and server port.
* **Bug Fixes**
  * Improved local IP detection by prioritizing suitable physical network interfaces and excluding virtual interfaces.


Pierre-Gilles:
Merged

PR #2980, opened 2026-08-21
Settings: turn the houses tab into a readable list of houses

Pierre-Gilles:
### Description

The houses tab rendered every house as a fully expanded form, stacked one after the other. With two houses the page was two ~1000px forms — two Leaflet maps included — with nothing but a card title to tell them apart, and no way to see at a glance how a house was set up.

**The tab is now a list.** Each house is one row showing its name, its room count, and whether it has a location and an alarm code. Only the house being edited opens its form (one at a time, and its map mounts with it). A row also says when a house was never saved or holds unsaved changes, so closing one no longer hides pending edits.

**The form is split into labelled sections** — identity, location, rooms, alarm — with the actions gathered in a footer: the save CTA on the left, the delete action pushed to the right behind an explicit confirmation naming the house. Saving shows a short confirmation, an empty name blocks the save instead of failing server-side, and rooms wrap as chips instead of sitting in a 4-column grid full of holes.

Around it:

- a new house is created with a localized name (the previously unused `defaultNewHouseName` key), suffixed so creating several in a row doesn't hit the unique-name constraint, and it opens right away;
- the empty page offers to create a first house instead of only saying "no houses found", and a search that matches nothing now reads differently from an install with no house;
- the per-house update status is keyed by house id everywhere, so one house's save no longer wipes the alert another one is showing (`deleteHouse` also stored an unapplied `$auto` literal in its error path);
- `saveHouse` and `deleteHouse` no longer apply their result to the state and row index captured before the request, so a house added, deleted or saved meanwhile can't be written to or spliced out by mistake;
- the room removal control and the alarm-code visibility toggle are native buttons with localized accessible names, reachable from the keyboard.

Only the settings tab is touched: `EditHouse` is used nowhere else, and the signup house step is unchanged.

The list row reads, left to right: house icon, name, then `9 rooms · Located · No alarm code`, with a "Not saved yet" / "Unsaved changes" badge and a chevron on the right. Screenshots of the list, the open panel, the new-house flow, the delete confirmation and the mobile layout were taken locally against demo mode.

### Checklist

- [x] Tests pass: the server is untouched (no coverage impact), and the Cypress job is green on CI. The page was also exercised by hand in demo mode: list, open/close, rename → unsaved badge, create → save → saved notice, failed delete → panel and error kept, keyboard access to the room and alarm-code buttons, the localized new-house name in a fr-FR browser, and the mobile layout.
- [x] Linter and prettier pass on the front (`npm run eslint`, `npm run prettier-check`), translations are complete (`npm run compare-translations`), and `npm run build` succeeds.


Pierre-Gilles:
Thanks — all the review findings were valid and are fixed in 80c34fe.

| Finding | Fix |
| --- | --- |
| `state.language` is never set, the locale lives in `state.user.language` | `getDefaultHouseName` reads `state.user.language`. Verified in a fr-FR browser: a new house is now named "Nouvelle Maison" (then "Nouvelle Maison 2"), where it used to be "New House" — the key this PR wires up now actually applies. |
| Page header summarizes 0 rooms as "1 room" | Shares the 0 / 1 / n split `RoomsSummary` already had. Two houses with every room removed now read "2 houses · No room". |
| Room removal / alarm-code toggle are click-only containers | Both are native `<button type="button">` with a localized accessible name (same shape as the gateway backup key toggle). Checked with the keyboard: focus + Enter removes a room and toggles the alarm code. |
| "+" button has no accessible name below `lg` | Localized `aria-label` at every width (`housesSettings.newButtonLabel`). |
| A failed deletion still collapses the row and drops its badge | The row closes only once the house is really gone, watched through its per-house status like a save. Checked against a failing delete: panel stays open, error alert visible, "Unsaved changes" badge kept. |
| Dirty markers cleared before the refetch result is known | They are cleared when a fetch actually replaces the collection (`Getting` → `Success`), so a failed search or re-order keeps them. |
| Async results applied to captured state and index | `saveHouse` and `deleteHouse` now read `store.getState()` after the request and locate their house by id, so a house added, deleted or saved meanwhile can no longer be written to or spliced out by mistake. |
| First-save success keyed under the temporary id | Worth noting the id does not actually change today: `POST /api/v1/house` sends the client-generated `id` and `db.House.create` persists it, so the response carries the same id. Hardened anyway — the save publishes its status under the returned id as well as the sent one, so the row stays in sync if that ever changes. |

Front lint, prettier, `compare-translations` and the build all pass.

---

PR #2989, opened 2026-08-23
Migration: group destination devices by integration and show room

sescandell:
## Summary

When migrating a device to another one, the migration popin listed candidate devices by name only. When several devices share the same name (a common case, e.g. multiple "Light" devices coming from different integrations or rooms), it was impossible to tell which one to pick from the dropdown.

This PR improves the readability of that choice by giving more context on each candidate device, so the user can confidently identify the right destination before starting a migration (which moves history, scenes and dashboards, then deletes the source device).

## Test plan

- [ ] Open the migration popin on a device with several same-named candidates in different rooms/integrations
- [ ] Verify the dropdown now lets each candidate be distinguished unambiguously
- [ ] Verify text search in the select still works
- [ ] Verify selecting a candidate still correctly triggers feature auto-mapping and the rest of the migration flow

## Summary by CodeRabbit

- **Enhancements**
  - Improved device search with accent- and case-insensitive matching.
  - Search now covers integration, room, and device names, requiring all entered terms to match.
  - Destination devices are grouped by integration, labeled with translated room and integration names, and sorted for easier browsing.
  - Selected devices now display room-aware labels.
  - Improved selector display and menu positioning for easier navigation.

- **Localization**
  - Added translated labels for unknown integrations in English, German, and French.


sescandell:
Thanks for the reviews — pushed two follow-up commits addressing the feedback:

- Devices without a resolvable service now group under a translated "Unknown integration" label instead of an empty group.
- The local accent/case-insensitive search helper was replaced with the existing `front/src/utils/normalizeSearchText.js` (also handles ligatures), removing the duplicated logic and the literal-combining-character regex.
- Groups are now keyed on the integration's technical slug but labelled/searched with the same human-readable name already shown on `/dashboard/devices` (`getDeviceIntegration` / `disambiguateIntegrationNames` from `front/src/routes/devices/integrationLinks.js`), so community integrations show their manifest name instead of `ext-owner-repo`, and built-in ones show their translated title instead of the raw service name.
- The destination select now portals its menu to `document.body` (same pattern as `SelectDeviceFeature`), so the taller grouped menu isn't clipped by the modal's scrollable overlay.
- `docs/specs/device-migration.md` B.5 updated to match the new picker behavior.

PR #3005, opened 2026-08-25
fix(security): reserve integration secrets to admins, and remove the shell from the backup chain

HowmationFr:
### Description

Two independent authorization and injection issues, both in code paths that handle integration credentials. They are grouped here because they share a threat model — an account with no administrative rights, or none at all, reaching secrets and commands it should never touch — but each is reviewable on its own commit.

---

#### 1. Service and global variables were readable and writable by any authenticated user

`GET|POST /api/v1/service/:service_name/variable/:variable_key` and `GET|POST /api/v1/variable/:variable_key` were only marked `authenticated`. Service variables hold the credentials of the integrations — the Telegram bot token, the MQTT broker password, the Netatmo OAuth tokens, the eWeLink and CalDAV passwords — and global variables hold the instance-wide Gladys Plus keys. Any user, a guest included, could read all of them, and overwrite them: pointing `MQTT_URL` at another broker exfiltrates the whole sensor traffic, and corrupting `GLADYS_GATEWAY_USERS_KEYS` breaks Gladys Plus.

Authorization now follows the scope of the variable rather than the route:

- `GET|POST /api/v1/variable/:variable_key` become `admin: true`. Per-user settings already go through `/api/v1/user/variable/:variable_key`, which can only ever touch the caller's own row and stays open to everyone.
- The two service-variable routes cannot be flagged at the route level: the same routes serve the per-user account pages (CalDAV, Nextcloud Talk, CallMeBot) through `userRelated`. The check moved into the controller and is derived from the resolved `userId`, so the authorization and the row being addressed can never diverge — a service-wide variable requires an admin, a user-scoped one does not. `variable.getValue` matches exactly on `(service_id, user_id)` with no fallback, so a user-scoped read cannot reach a service-wide row.

Both the local HTTP API and the Gladys Plus gateway go through these checks: `setupGateway` already honours the `admin` flag, and the controller check covers the gateway path too.

Some services expose the very same secrets through a route of their own, handing their configuration object straight to `res.json()`. Left open they would make the fix above pointless, so they are reserved to admins as well: mqtt `config` (broker password), netatmo `configuration` (client secret), zwavejs-ui `configuration` (broker password), nuki `config` (API key), and the tuya `configuration` write. Their front pages are device integrations, already hidden from non-admin users, so no screen loses a working call. node-red `configuration` is left as is: its controller already whitelists the two version fields and never returns the stored password.

Front side: the Telegram page is visible to every user — an admin configures the bot there, a habitant gets their own linking link. Reading the API key first would now abort the whole page load for a non-admin, so it is only fetched for admins, which is also the only role the form is rendered for.

---

#### 2. The name of a restored backup reached a shell

`POST /api/v1/gateway/backup/restore` builds every path it works on from the basename of the caller-supplied `file_url`, and that name was interpolated into a `/bin/sh` command:

    exec(`gzip -dc ${compressedBackupFilePath} > ${sqliteBackupFilePath}`)

`path.basename` stops a path traversal but keeps `$( )`, backticks, quotes and semicolons, so the file name was executed by the shell. That route and `POST /api/v1/gateway/backup-key` are both `authenticatedOrNotConfigured`, which means that while no user exists in database the whole chain is reachable without authenticating — the signup restore flow. On a configured instance an admin reaches it too, which turns an admin UI into a shell in the Gladys process.

The same caller-controlled string reached two more interpreters further down:

- `gateway.restoreBackup.js` built `sqlite3 STORAGE ".restore 'PATH'"` for a shell, `PATH` being the name of a file inside the restored archive. The existing archive check only rejected absolute paths, `..` and symlinks, not metacharacters.
- `IMPORT DATABASE 'FOLDER'` is a DuckDB statement, and DuckDB reads and writes files from SQL. `FOLDER` comes from the same archive.

Tracing the chain surfaced a fourth sink, on the backup side this time: `gateway.backup.js` ran `openssl enc -pass pass:KEY` through a shell, `KEY` being the backup passphrase, which `saveBackupKey` stores without any validation. A key containing a command substitution runs on every backup, including the scheduled ones.

**Removing the shell is the fix**, applied to the whole backup/restore chain rather than to the one reachable call:

- `gzip -dc IN > OUT` becomes `spawnToFile('gzip', [...], out)`, a new helper whose redirection is a write stream. `execFile` is not usable here: it buffers stdout, and a database is not bounded by the exec buffer, so large restores would break.
- `sqlite3 .backup` / `.restore`, and `tar` (through a new `cwd` option on `execFile`, replacing a `cd ... &&`), become argument-array calls.
- `openssl enc` becomes an argument-array call. Side effect: a passphrase containing a space no longer breaks the backup, since no shell splits it any more.
- Both DuckDB `IMPORT`/`EXPORT DATABASE` paths are escaped for the SQL string literal they sit in.

Defense in depth, in the new `utils/backupSafety.js`: a backup name and every archive entry must match `[A-Za-z0-9._-]`, checked **per path segment** so the Parquet folder keeps its nested files. `.` and `..` are rejected explicitly, since the alphabet allows the dot. The name is validated before the download, so a malformed URL is never fetched at all, and a rejected archive is never retried with the old restore strategy.

`post /api/v1/gateway/backup/restore` also gets `rateLimit: true`: it is reachable before authentication and makes the server download and unpack a remote file.

**The backup key is deliberately left unconstrained.** It is the passphrase existing backups were encrypted with, so refusing some characters — or refusing to overwrite an existing key — would lock users out of restoring their own data. With no shell left in the chain, the key does not need a charset.

---

#### Tests

Regression tests reproduce both issues rather than only asserting the new behaviour: non-admin reads and writes rejected on service-wide and global variables (over HTTP and through the Gladys Plus gateway), with the stored value checked to be untouched after a rejected write; the `userRelated` flow verified to still work for a non-admin; a real backup renamed `evil$(touch MARKER).enc`, asserting both the rejection and that the marker was never created; and an encrypted archive fixture whose single entry is named `x$(touch ...).db`.

#### Not addressed here

`file_url` is passed as-is to an `axios` GET, so the server will fetch any URL including internal ones — a blind SSRF, unauthenticated on a fresh instance. Fixing it needs to know which hosts Gladys Plus signs its backup URLs on.

Roughly a hundred other service routes (`discover`, `scan`, `permit_join`, `connect`…) are reachable by any authenticated user. That is action authorization rather than secret disclosure, and it deserves its own change.

### Checklist

- [x] Tests pass: full server suite green, with the failure set compared against the pre-change baseline to confirm no regression. Coverage is 100% on every changed line; the lines left uncovered in `gateway.downloadBackup.js` and `childProcess.js` are pre-existing code this PR does not touch. Cypress green.
- [x] No undocumented breaking change. Behaviour change to be aware of: a non-admin user now gets a 403 on the instance-wide settings that were previously writable by anyone (timezone, data retention, battery threshold, mDNS hostname) and on the integration setup screens.

## Summary by CodeRabbit

* **Security Improvements**
  * Restricted service configurations, credentials, shared variables, and administrative settings to administrators.
  * Restricted access to integrations with shared credentials or household-wide access.
  * Added rate limiting for unauthenticated backup restoration on unconfigured instances.
  * Hardened backup downloads, restores, and archives against unsafe names, paths, and command injection.
  * Improved Telegram integration loading so users can continue accessing their personal link when API-key retrieval fails.

* **Tests**
  * Added coverage for authorization controls, backup safety, and error handling.


front/src/routes/integration/all/telegram/actions.js
```diff
@@ -11,10 +12,15 @@ const actions = store => ({
...
     });
     try {
-      const variable = await state.httpClient.get('/api/v1/service/telegram/variable/TELEGRAM_API_KEY');
-      store.setState({
-        telegramApiKey: variable.value
-      });
+      // the bot API key is a service-wide secret: only an admin can read it,
+      // and only an admin is shown the form to change it. Every other user
+      // comes to this page for their own linking link, which is per-user.
+      if (state.user && state.user.role === USER_ROLE.ADMIN) {
```
Pierre-Gilles: On a hard page load this condition is always false, even for an admin. `getDefaultState()` seeds the store with `user: { language }` only, and `checkSession()` fills in the full `user` after `/api/v1/me` resolves — which is strictly after the first render, while `componentWillMount` in `index.js` dispatches `getTelegramApiKey` *during* that first render. So an admin refreshing (or deep-linking to) the Telegram page never loads the API key and sees an empty field. The form itself recovers because rendering is reactive, but this action is never re-run.

Two options, both already used in the codebase:
- fetch unconditionally and swallow the 403 (no client-side role knowledge needed), or
- re-trigger the load when `user` arrives, the way `front/src/routes/integration/index.js` reloads in `componentDidUpdate` when the user id changes.

---

server/api/controllers/variable.controller.js
```diff
@@ -1,5 +1,27 @@
...
+ * integrations (API keys, broker passwords, OAuth tokens), so reading or
+ * writing them is an administration gesture. Variables scoped to the calling
+ * user are left alone: a user can only ever touch their own row.
+ * @param {object} req - The Express request.
+ * @param {string} [userId] - The user the variable is scoped to, or null when it is service-wide.
+ * @returns {void}
+ * @example
+ * ensureAdminOnServiceWideVariable(req, null);
+ */
+function ensureAdminOnServiceWideVariable(req, userId) {
```
Pierre-Gilles, line 17: The check itself is right and well-placed (deriving it from the resolved `userId` means the authorization and the row being addressed can't diverge, and it covers the gateway path too). But one visible page didn't get the Telegram treatment: **Free Mobile** is a `communication` integration, so it stays visible to non-admin users (only the `device`/`weather` types and homekit are hidden from them), and its page reads `FREE_MOBILE_USERNAME` / `FREE_MOBILE_ACCESS_TOKEN` **without** `userRelated` on mount (`front/src/routes/integration/all/free-mobile/actions.js`). A habitant opening that page now gets 403s and lands in the error state.

Since there is nothing per-user on that page (unlike the Telegram linking link), the cleanest fix is probably to add `free-mobile` to `HIDDEN_INTEGRATIONS_FOR_NON_ADMIN_USERS` in `front/src/routes/integration/index.js`. As it stands, the PR's "no screen loses a working call" claim doesn't quite hold for this one.

---

server/api/routes.js
```diff
@@ -825,12 +829,18 @@ function getRoutes(gladys) {
...
+    // settings go through /api/v1/user/variable below, which stays open to
+    // every authenticated user.
     'post /api/v1/variable/:variable_key': {
       authenticated: true,
+      admin: true,
       controller: variableController.setValue,
     },
     'get /api/v1/variable/:variable_key': {
       authenticated: true,
+      admin: true,
```
Pierre-Gilles, line 843: Restricting these routes is correct, but two front surfaces still reachable by non-admins consume them and weren't adjusted:

- the **OpenAI** page (`communication` type, so visible to habitants): `WeeklyDigestSettings.jsx` reads and writes `/api/v1/variable/AI_WEEKLY_DIGEST_*` on mount;
- the **Settings → System / Backup / Gateway** tabs: `SettingsLayout.jsx` has no role gating, and those pages read/write the timezone, device history, mDNS hostname, backup key… through these routes.

The PR body documents the 403 behaviour change, so this may be intended — but as it stands habitants get error states on screens they can still navigate to. Either hide these tabs/pages for non-admins (here or in a follow-up), or gate the fetches like the Telegram page. At minimum the OpenAI weekly digest card is worth handling in this PR, since the integrations catalog still shows that page to them.

---

Pierre-Gilles (changes requested):
Strong PR overall — the server side is ready as far as I can tell, and I'm requesting changes only for the front-end fallout (see inline comments).

**What I verified and confirmed sound:**

- The controller-level authorization check is well designed: derived from the resolved `userId` rather than the route, so the check and the row being addressed can never diverge. `variable.getValue` matches exactly on `(service_id, user_id)` with no fallback, so a user-scoped read can't reach a service-wide row.
- The Gladys Plus gateway path is covered: `setupGateway` honours the `admin` flag, and the controller check applies to gateway calls too (and is tested).
- The claims in the description check out: node-red's `configuration` really does whitelist the two version fields; CalDAV, Nextcloud Talk and CallMeBot all pass `userRelated: true`; MQTT's `saveConfiguration` is mounted on `/connect`, which was already admin-only; the `device`/`weather` integrations and homekit are hidden from non-admins in the catalog. I also swept the remaining non-admin `config` routes (bluetooth, caldav, lan-manager) — no secrets in their payloads.
- The backup/restore chain: `spawnToFile` is correctly implemented (waits for both process exit and stream flush, all error paths handled), `backupSafety` validates per path segment and explicitly rejects `.`/`..`, absolute paths and tar's trailing slash are handled, and the DuckDB literal escaping is standard. The name is validated before the download, and a rejected archive never falls back to the old restore strategy.
- The tests are genuinely adversarial (a real backup renamed with `$(touch …)`, a booby-trapped archive fixture, gateway 403, stored value asserted untouched after a rejected write).

**Changes requested (inline):**

1. `front/src/routes/integration/all/telegram/actions.js` — on a hard page load the store's `user` isn't populated yet when `getTelegramApiKey` runs, so an admin refreshing the page never loads the API key.
2. The Free Mobile page stays visible to non-admins and now errors out (its variables are service-wide, fetched without `userRelated`).
3. The OpenAI weekly digest card and the Settings → System / Backup / Gateway tabs are still reachable by non-admins and now hit 403s — decide whether to hide them or gate the fetches (a follow-up is fine for the settings tabs, but the OpenAI card is worth handling here).

**Nit:** the description's list of service routes moved to admin doesn't mention zigbee2mqtt `setup`, which the diff (rightly) includes.

---

Pierre-Gilles:
Merged in https://github.com/GladysAssistant/Gladys/pull/3086

Thanks for your PR 🙌

PR #3030, opened 2026-08-28
fix: give the "Talk on a speaker" volume a Horizon slider (#3025)

### LINKED ISSUE #3025: The scene action "Talk on a speaker" volume is ugly in Horizon theme
The volume bar is not very clean in the Horizon theme, it should be adapted to be more "Horizon" : 

Image

Pierre-Gilles:
> **This pull request was produced by an automated routine (Claude Code) and has NOT been reviewed by a human.** Please review it carefully before merging.

Fixes #3025

### Description

**What changed and why**

The scene editor renders on the Horizon glass scene and paints every `.form-control` as a rounded white field (`front/src/routes/scene/edit-scene/style.css`). The "Talk on a speaker" volume control was built out of two `.form-control` elements, so on the glass it read as two stacked white boxes: a full-width **disabled** text field mirroring the value, and the `input[type=range]` trapped inside a second field box — which is the look reported in the issue.

The volume now follows the Horizon grammar already used by the other glass controls (the setpoint capsule, the light slider, the frosted pills):

- the value became a quiet rounded pill sitting on the label line, replacing the full-width disabled mirror field;
- the slider dropped `form-control custom-range` for its own skin: a soft rounded track (`rgba(28, 35, 52, 0.09)` with the same inset shadow as the light slider), an accent fill up to the current value, and a white round thumb with a soft shadow, plus a focus ring;
- the fill is a gradient stop driven by a `--volume-fill` custom property set inline from the value. A range input only fires `change` on release, so the fill and the pill follow the drag through a local draft state; the scene action itself is still only updated on release, exactly as before (no extra editor-wide re-renders while dragging).

Colors are light-only, as the rest of the Horizon CSS: dark mode comes from the app-wide inversion filter, and `style/dark-mode.css` explicitly excludes `input[type='range']` from the field repainting, so the new track inverts with the theme.

Files touched:
- `front/src/routes/scene/edit-scene/actions/PlayNotification.jsx`
- `front/src/routes/scene/edit-scene/actions/PlayNotification.css` (new, same per-action CSS-module convention as `CheckTime.css` / `DeviceSetValue.css`)

No server change, no new translation key (the pill renders `${volume}%`, matching the `{value}%` convention used elsewhere in the front).

### Checklist

- [x] Tests pass: no server code changed, so `npm run coverage` is not applicable to this diff; Cypress could **not** be run here (the Cypress binary cannot be downloaded in this sandbox) and no existing spec covers this action

**What was verified locally**

- `cd front && npm run prettier` then `npm run prettier-check` → "All matched files use Prettier code style!"
- `cd front && npm run eslint` → 0 errors (186 pre-existing warnings repo-wide; the one new warning on the changed file is `react/forbid-dom-props` for the inline `--volume-fill`, the same warning many existing components already carry)
- `cd front && npm run compare-translations` → no errors
- `cd front && npm run build` → built successfully; checked the emitted CSS to confirm the CSS-module class and the `--volume-track` / `--volume-fill` custom properties survive the build
- Server was not touched, so server tests were not run

**What a human must check before merge**

- **The actual rendering** — this was not verified visually: no browser is available in the sandbox, so the new slider was never seen on screen. Please open a scene with a "Talk on a speaker" action and confirm the track, fill, thumb and value pill look right in the Horizon theme, in **light and dark mode**, on desktop and on a phone.
- Cross-browser: the track/thumb skin uses `::-webkit-slider-*` and `::-moz-range-*`; please confirm Firefox and Safari render the same shape (in particular the thumb centering on the 8&nbsp;px track).
- Whether replacing the read-only value field with a pill is the wanted direction, or whether the value should stay editable.
- When the action has never been given a volume, the value pill is hidden and the fill sits at 50% to match the position the browser gives an untouched range. That mirrors the previous behaviour (the disabled field was empty), but it is a deliberate choice worth confirming — the action still saves `volume: undefined` until the slider is touched.
- The sibling `DeviceSetValue` action still uses `form-control custom-range` and so keeps the boxed look. It was left alone to stay inside the scope of this issue; it may deserve the same treatment in a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VinPPcP62ovdumA7r5n2Eu)_

## Summary by CodeRabbit

* **UI Improvements**
  * Enhanced the scene editor’s volume control with a custom slider, dynamic fill indicator, and percentage display.
  * Added clearer keyboard focus states and improved visual styling across supported browsers.
  * Volume changes now preview while dragging and are saved when the slider is released.

### DIFF
--- a/front/src/routes/scene/edit-scene/actions/PlayNotification.css
+++ b/front/src/routes/scene/edit-scene/actions/PlayNotification.css
@@ -1,0 +1,95 @@
+/* The Horizon volume control. The editor paints every .form-control as a rounded
+   white field on the glass, which turned the slider into a boxed field and the
+   read-only value into a second full-width box: the volume now reads as a value
+   pill on the label line, over a soft rounded track.
+   Light colors only: dark mode comes from the app-wide inversion filter, which
+   leaves range inputs alone (style/dark-mode.css). */
+
+.volumeLabel {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 0.5rem;
+}
+
+.volumeValue {
+  background-color: rgba(28, 35, 52, 0.07);
+  border-radius: 999px;
+  padding: 0.1rem 0.6rem;
+  font-size: 0.75rem;
+  font-weight: 600;
+  color: var(--gl-ink, #1c2334);
+}
+
+/* The slider is not a field: it drops the .form-control box and becomes a track
+   of its own. The accent fill is a gradient stop driven by --volume-fill, set
+   inline from the current value (custom properties reach the pseudo-elements). */
+.volumeRange {
+  display: block;
+  width: 100%;
+  height: 1.5rem;
+  padding: 0;
+  margin: 0;
+  background: transparent;
+  border: none;
+  cursor: pointer;
+  -webkit-appearance: none;
+  appearance: none;
+  --volume-track: linear-gradient(
+    to right,
+    var(--gl-accent, #3d6df0) 0%,
+    var(--gl-accent, #3d6df0) var(--volume-fill, 50%),
+    rgba(28, 35, 52, 0.09) var(--volume-fill, 50%),
+    rgba(28, 35, 52, 0.09) 100%
+  );
+}
+
+.volumeRange:focus {
+  outline: none;
+}
+
+/* Chrome, Safari, Edge */
+.volumeRange::-webkit-slider-runnable-track {
+  height: 8px;
+  border-radius: 999px;
+  background: var(--volume-track);
+  box-shadow: inset 0 1px 3px rgba(28, 35, 52, 0.12);
+}
+
+.volumeRange::-webkit-slider-thumb {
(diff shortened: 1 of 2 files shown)


Pierre-Gilles:
The three review comments are addressed in [#3038](https://github.com/GladysAssistant/Gladys/pull/3038), which targets this PR's branch (`claude/issue-3025`) — merging it brings the fixes here:

- the accent fill no longer sits at 50% while `volume` is still `undefined`: the track stays grey until a value is committed, thumb at the browser midpoint;
- `::-moz-range-progress` is neutralised so Firefox stops painting the native fill over the custom track, and `::-moz-range-track` spans the full width;
- `:focus-visible` gains a transparent outline so keyboard focus survives forced-colors mode, where `box-shadow` is stripped.

The WebKit thumb `margin-top: -6px` is `(8px − 20px) / 2` and correct for the 8px track, so it was left as is. The rendering still has not been checked visually — no browser in this sandbox.

---

PR #3066, opened 2026-08-30
Show room name alongside device name in the devices widget selector

sescandell:
### Description

In the dashboard "Devices" widget, when adding a feature through the dropdown, devices were grouped only by their name. When several devices share the same name (e.g. multiple "Lumiere"), it became impossible to tell which one to pick.

The group label now includes the room name alongside the device name (`Lumiere - Master Bedroom`) to remove the ambiguity.

### Checklist

## Summary by CodeRabbit

* **Enhancement**
  * Device options now display the associated room name when available, using the format “device - room.”
  * Devices without an assigned room continue to display their original names.

### DIFF
--- a/front/src/components/boxs/device-in-room/EditDevices.jsx
+++ b/front/src/components/boxs/device-in-room/EditDevices.jsx
@@ -107,8 +107,9 @@ class EditDevices extends Component {
           feature => !selectedDeviceFeaturesOptions.some(selected => selected.value === feature.value)
         );
         if (filteredDeviceFeatures.length > 0) {
+          const roomName = device.room && device.room.name;
           deviceOptions.push({
-            label: device.name,
+            label: roomName ? `${device.name} - ${roomName}` : device.name,
             options: filteredDeviceFeatures
           });
         }


Pierre-Gilles:
Thanks a lot for raising this, the problem is real: two devices sharing a name gave two identical group headers in the "add a feature" dropdown of the Devices widget, and there was no way to tell which one to pick.

We went with a slightly different approach in #3082 rather than appending the room to the header, for two reasons:

- Many users already put the room in the device name ("Lumière salon"), so a `Lumière salon - Salon` header adds noise for them, and headers were getting long in a menu that is already height-constrained.
- react-select only searches option labels, not group headers, so typing the room name still would not have narrowed the list.

#3082 reuses `SelectDeviceFeature`, the picker already used by scenes and the actions box. Features are now grouped by **room** (with a "No room" group at the end), every option keeps the device name in its label (`Lumière (Luminosité)`), and the room is part of the search text, so typing `lumiere chambre` narrows the list. Same-named devices in different rooms end up under different headers, which covers your case, and the stored box config is unchanged.

It is being merged now, so I'm closing this one. Feel free to reopen the discussion if your setup still hits a case that #3082 does not solve.

---

Issue #2999, opened 2026-08-24
Energy: support contracts with weekend / day-type pricing (EDF Zen Week-End, Engie Elec Week-end…)


guim31:
## Context

A whole family of French electricity offers bills **weekends and public holidays at the off-peak rate, at all hours**, on top of the usual daily peak/off-peak split. Examples:

- **EDF Zen Week-End** — option *Heures Creuses + WE*: 22.60 cts €/kWh peak on weekdays, 16.92 off-peak on weekdays, and 16.92 all day Saturday, Sunday and public holidays
- **Engie** Elec Week-end
- **OHM Énergie** Soir et Week-End
- **Enercoop** Flexibilité nuit & week-end
- **Électricité de Strasbourg** offre Week-end

These are not niche: they are the standard commercial answer to the ongoing Enedis/CRE off-peak slot reform.

## Problem

The current data model cannot express them. `hour_slots` is a set of half-hour slots (`"04:30,05:00,…"`) applied identically to **every day of the week** — there is no day-of-week dimension. And `ENERGY_CONTRACT_TYPES` (`server/utils/constants.js`) only allows `base`, `peak-off-peak` and `edf-tempo`.

Interestingly, the building block already half exists: Tempo uses `day_type` (`blue` / `white` / `red`) to vary the price by day. But `day_type` is assigned per-day by RTE and imported as data, whereas a weekend is deterministic from the calendar.

## Consequence today

The workaround is to import such a contract as a plain `peak-off-peak` and simply drop the weekend rebate. On a flat consumption profile that overestimates the bill by about **5.5 %**, and considerably more for households that shift their heavy usage to the weekend — which is precisely what these offers are designed to encourage. The error is systematic and always in the same direction, so it does not average out over time.

I have opened GladysAssistant/energy-contracts#16 adding EDF Zen Week-End with exactly that approximation, documented as such. It felt better than nothing, but it is clearly not the right long-term answer.

## Possible directions

Rough sketches, no strong opinion on which is best:

1. **Extend `day_type`** with calendar-derived values such as `weekend` / `holiday` / `weekday`, resolved at billing time rather than imported. Reuses the Tempo machinery and stays close to the existing schema.
2. **Add a new contract type** (e.g. `peak-off-peak-week-end`) with its own hour-slot handling, leaving `peak-off-peak` untouched.
3. **Make `hour_slots` day-aware**, which is the most general but also the most invasive change.

[...]

PR #3099, opened 2026-09-08
feat(energy): energy-calendar integration type and day-type contract

### LINKED ISSUE #1234: Tasmota: manage IR and RF sensors
According to https://community.gladysassistant.com/t/releve-de-bugs-divers/6271/122?u=alextrovato

Be able to handle IR and RF Tasmota message as sensor first.

guim31:
Energy contracts whose prices depend on the kind of day (EDF Zen Week-End, Engie Elec Week-end...) could not be modelled: hour_slots applied to every day alike and day_type was an ENUM of the Tempo colors. National calendars do not belong in the core (#2999): this adds the API letting an external integration provide them, and a generic contract type that uses it.

- new manifest type "energy-calendar": provider API relayed over WS external-integration.energy-calendar.get-day-types, payload normalized and bounded (spec B.19)
- new core manager gladys.energyCalendar (duck-typed provider loop)
- new "day-type" contract: prices keyed by a free day_type slug and optional hour_slots, calendar asked once per cost run
- t_energy_price.day_type widened from ENUM to a validated string (TEXT under SQLite, no migration)
- frontend: contract select, import suffixes, external integration screens, en/fr/de translations

Forum: https://community.gladysassistant.com/t/api-permettre-aux-integrations-externes-de-declarer-leurs-propres-types-de-contrat-energie/10704

### Checklist

- [ ] If a forum topic or GitHub issue exists, the description links it (`Forum: https://community.gladysassistant.com/t/...` or `Closes #...`)
- [ ] Tests pass: `cd server && npm run coverage` (Codecov requires 100% coverage on changed lines) and Cypress (`npm run cypress:run`) if the UI changed
- [ ] Linter and prettier pass on both front and server (`npm run eslint`, `npm run prettier`)
- [ ] No undocumented breaking change

## Summary by CodeRabbit

* **New Features**
  * Added energy-calendar integrations for weekdays, weekends, public holidays, and other day types.
  * Added day-type energy contracts with day-type and time-slot pricing.
  * Added provider fallback, date-range validation, and pricing fallback behavior.
  * Added support across the integration interface, SDK, and WebSocket API.
* **UI & Localization**
  * Added energy-calendar guidance and day-type contract fields in English, German, and French.
  * Energy-calendar integrations open directly to configuration without device screens.
* **Bug Fixes**
  * Improved handling of invalid or unavailable calendar data during cost calculation.


Pierre-Gilles:
Hi @guim31, thanks a lot for this PR, and for the spec-first approach: the B.21 section, the bounded payload, the fallback loop and moving the cost history cleanup after the prerequisites are all well thought out. It's really nice to see a contribution that follows the existing patterns this closely.

Before going further, I'd like to share where I want to take the energy feature, because it changes what the right abstraction is.

**The goal: any contract in the world, without a core PR.** Today adding a "complex" contract means adding a contract type to the core, a cost function, a frontend select entry, translations… This PR generalizes one family (prices keyed by day type + hour slots), which is a real step, but it still adds a hardcoded type to a closed list. Seasonal tariffs, dynamic/spot pricing (Tibber, Octopus Agile, EPEX-based offers), tiered pricing, or the next offer we haven't thought of would still need a core PR. And we'd end up with two parallel mechanisms for the same concept (Tempo with its gateway feed and 6 AM rule, day-type with its provider), which is the kind of debt that never gets cleaned up.

Looking at what tariffs actually look like worldwide, they boil down to two data-driven primitives the core should support natively, with no contract-specific code:

1. **Time-based pricing: one label (or price) per 30-minute slot.** A provider resolves each slot of a range to a label, or directly to a price, in one call per cost run (like your `getDayTypes`, but per slot instead of per day). The core stores prices keyed by label and multiplies. This covers peak/off-peak with seasons, weekends and public holidays (Ontario, Australia, Japan, California TOU), Tempo and critical-peak days, and dynamic hourly tariffs when the provider returns prices. `hour_slots`, the 6 AM Tempo rule and the per-contract cost functions disappear from the core; Tempo becomes a provider like any other, and weekday/weekend can stay resolvable natively since they're deterministic.

2. **Volume-based pricing: tiers on cumulative consumption within a billing cycle.** The default model in the US (California baseline), Canada (BC Hydro, Ontario tiered), Japan, Korea, India, China and Latin America: the first N kWh of the cycle at one price, the next at another. Thresholds and the cycle start date are data; the core tracks the cumulative consumption itself, which it can do since the cost run already walks the states in order. Tiers combine with labels, because some tariffs (California) do both at once.

Demand charges (billing on the monthly kW peak) and monthly net metering are a third family I'd leave for later.

On top of that, a provider should declare its vocabulary in the manifest so the price editor offers a select instead of a free text input, and Gladys should warn when a contract references a provider that isn't installed.

Your PR already contains a lot of what this needs (the provider loop, the normalization, the proxy capability, the "load everything before destroying history" fix), so the work isn't lost, and I'll reuse it.

I'd rather not merge an intermediate `day-type` contract type that we'd deprecate a few months later, so I'm going to close this PR and take the general version on myself. I'll write up the target design on the forum topic first, so the interface is public before the code, and I'd be glad to have your review on it.

Thanks again, this was a genuinely useful push on a topic I wanted to open up.

---

PR #3101, opened 2026-09-10
feat(alarm): three arming modes — Presence, Night, Away

cicoub13:
## Why

A house alarm is rarely all-or-nothing: people protect their home differently depending on whether they are away, home during the day, or asleep. That is the trio most alarm panels on the market offer, and it is also what HomeKit expects.

Gladys only had two levels, `armed` and `partially-armed`. Anyone who wanted a night mode repurposed the partial one for it and was then left without a presence mode. And because Gladys had no night mode, the HomeKit integration did not expose one either — `NIGHT_ARM` was deliberately excluded from `SUPPORTED_TARGET_STATES`, so the Home app showed no Night button.

Panic also sat in the same list as the arming modes, which conflated two different things: an arming mode is a level of protection you choose, panic is something you do.

## What changes

**The enumeration.** `ALARM_MODES` becomes `disarmed`, `presence-armed`, `night-armed`, `away-armed`, `triggered`. The first four are arming modes, gathered in the new `ALARM_SETTABLE_MODES_LIST`; `triggered` is deliberately left out of it, because it is a state the alarm enters by itself and never one anybody asks for. `ALARM_MODES_LIST` (all five) stays what the model, the Joi scene schema and `CHECK_ALARM_MODE` validate against — a scene may legitimately test whether the alarm is going off.

**Panic becomes an action.** New scene action `alarm.trigger-panic` (house only, no mode), a button of its own set apart in the widget, and `house.panic()` now puts the house in `triggered` whatever mode it was armed in. The `alarm.panic` scene trigger is unchanged.

**One `arm` for three modes.** The three arming modes share the same code path — same delay, same tablet locking, same conflict check — and differ only by the pair of events they announce themselves with. `house.partialArm.js` is gone.

**One delay, for all three.** `alarm_delay_before_arming` now applies to the three arming modes alike. This is the one behavior change of the lot: partial arming used to be immediate.

**HomeKit.** The states map one for one, the Night button appears in the Home app, and the deliberate `NIGHT_ARM` exclusion goes away.

| Gladys | HomeKit |
|---|---|
| `presence-armed` | `STAY_ARM` |
| `night-armed` | `NIGHT_ARM` |
| `away-armed` | `AWAY_ARM` |
| `disarmed` | `DISARMED` |
| `triggered` | `ALARM_TRIGGERED` |

**Routes.** `presence_arm`, `night_arm` and `away_arm` replace `arm` and `partial_arm`, with **no compatibility alias**.

## Migration

Renaming values stored in `t_house.alarm_mode` and hard-coded in everyone's scenes costs a data migration, so this one is paid once — the new spec records that the enumeration is **append-only** from here on.

`t_house.alarm_mode` is a plain `TEXT` column (Sequelize maps `ENUM` to `TEXT` on SQLite), so no table recreation is needed. The migration rewrites the houses, then walks `t_scene.actions` **recursively**, including the actions nested in `if` / `then` / `else` blocks, and rewrites the scenes in raw SQL so the Joi validator of `t_scene` cannot reject a scene over unrelated legacy content. A scene setting the alarm to `panic` becomes an `alarm.trigger-panic` action.

## Widget

Three mode tiles, a full-width disarm button, and panic under a rule as a quiet outline button. Presence introduces the only new hue; away keeps the blue the single "arm" tile wore, and night takes the ink the partial tile used to — it suits it better. A triggered alarm turns the whole card solid: it is the one loud state.

## Out of scope

Nothing here decides *when* the alarm goes off: no device is attached to the alarm, `triggered` has a single cause (the panic button), and there is no entry delay. Telling an intrusion apart from a deliberate press, naming the sensor that fired, and multi-mode scene conditions are all separate work.

## Also in this diff

`server/test/services/homekit/controllers/homekit.controller.test.js` is folded into `api/`, where it belonged since that directory was renamed. The two files were **not** duplicates despite appearances: they split the controller's three routes between them (`reload`/`reset` on one side, `device` on the other), so deleting the stray one outright would have dropped coverage.

## Checks run locally

- `server`: prettier, eslint (0 errors), full test suite, and 100 % line/function/branch coverage on every file touched
- `front`: prettier, eslint, `compare-translations`, `build`
- Cypress: 65/65 specs passing
- The gateway is unaffected: it relays websocket messages end-to-end encrypted and never reads their type — verified against the `gladys-gateway` repo, which has no reference to the alarm at all

Forum: https://community.gladysassistant.com/t/systeme-dalarme-gladys/10808

## Summary by CodeRabbit

* **New Features**
  * Added Presence, Night, and Away alarm arming modes.
  * Added separate Disarm and Panic controls; panic now displays a Triggered state.
  * Added mode-specific scene triggers and a dedicated panic scene action.
  * Added Night mode support for HomeKit.
* **Documentation**
  * Added specifications covering alarm modes, events, APIs, countdown behavior, and HomeKit mapping.
* **Updates**
  * Updated alarm widgets, dashboard states, translations, demo data, and scene configuration for the new modes.
  * Updated alarm notifications and controls to reflect the selected arming mode.


cicoub13:
Thanks — five of the seven are fixed in 09a9b809, and one is a question I would rather a maintainer answered than decide myself.

**Fixed**

1. **Overlapping delayed `arm()`.** A pending arming is now cleared before a new one is scheduled, so Presence-then-Night leaves one timer instead of two racing to write their own mode. `armHouse` also deletes its own entry when the delay is over, so a house that finished arming is no longer reported as "arming in progress" to `disarm` — that part was leaky before this PR too. Covered by two tests: one starts Away then Night and asserts only Night is written and Presence never announced itself, the other checks the map is empty once the delay elapsed.

2. **`panic()` during a countdown.** Same clear-and-delete block as `disarm()`. The test panics mid-delay and asserts the house is still `triggered` well past the moment the arming would have fired.

3. **Cancel during a mode switch.** Went with disabling the other two tiles while the house is armed: the way from one mode to another goes through Disarm. It needs no new server operation, and it removes the ambiguity rather than papering over it — on a wall tablet "Cancel" must never quietly mean "turn the alarm off". The spec says so explicitly now.

4. **ARMING not filtered by house.** The handler now ignores a payload whose `house` is not the widget's.

6. **Partial to Presence.** Called out in the spec's B.7: partial was semantically Stay — arming part of the house while its occupants live in it — which is what presence means. Some people stretched it into a night mode for want of one, and Gladys never recorded *why* partial was used, so the two populations cannot be told apart. Worth a line in the release notes, agreed.

7. **Half-migrated instance.** A scene whose JSON cannot be parsed is logged and skipped instead of rejecting the `Promise.all` after the houses have already been rewritten. The test covers a corrupt scene sitting next to a sound one, and asserts the sound one still goes through.

**Left open for a maintainer**

5. **Aliases for the two removed routes.** You are right that this reaches further than I first assessed. I had checked that the gateway relays websocket messages end-to-end encrypted and never reads their type, and concluded it was unaffected — but it also tunnels REST verbatim (`gladys-api-call` reaches `gateway.handleNewMessage.js`, which dispatches `options.url` into the local router), so any remote client still posting the old paths gets a 404.

@Pierre-Gilles, this one is your call: a pair of deprecated aliases kept until the next major, matching the migration's mapping, or the clean break with a coordinated mobile and docs release? Happy to add the aliases in a follow-up commit if you prefer the safer rollout — it is four lines.

**Also worth confirming (product)**

The exit delay now applies to the former partial path too, so a Presence or Night arm from inside the house waits the same countdown as walking out. That was a deliberate answer to "one delay for all modes", but it is the one behavior change in the diff, and I would rather it be an explicit decision than a side effect.

Checks after the fixes: server prettier and eslint (0 errors), 100 % line, branch and function coverage on `house.arm.js`, `house.panic.js`, `house.disarm.js` and `house.disarmWithCode.js`; front prettier, eslint, `compare-translations` and build; Cypress 65/65.

PR #3115, opened 2026-09-21
feat(scene): add a "Gladys conversation only" delivery option to the message and AI scene actions

cicoub13:
### Description

Installing a communication integration immediately changed what every existing
scene did. `forwardToChannels` broadcasts to every service exposing
`message.sendToUser`, so a "send message" action written years ago started
sending SMS without the scene being touched. The channel selector could name
one channel, but could not say "no channel at all": on an instance whose only
channel was SMS, there was no way to write in the Gladys conversation alone.

The `service` property of the message actions gains a third case, the sentinel
`MESSAGE_GLADYS_ONLY_SERVICE`:

| `action.service` | Delivery |
|---|---|
| absent | Gladys conversation + every channel configured by the user (unchanged) |
| a service name | Gladys conversation + that channel only (unchanged) |
| `MESSAGE_GLADYS_ONLY_SERVICE` | Gladys conversation only (new) |

`sendToUser` returns right after the database insert and the `message.new`
websocket event, which already happened before any channel was called: a
message is always in the Gladys conversation, and a channel is only ever an
addition. Existing scenes are not rewritten — no `service` still means
broadcast.

The same choice reaches the AI. The "ask the AI" action carries `service` into
its `source: 'AI'` message, and `message.reply` applies the three cases to the
answer instead of always broadcasting it, so a scene that only stores the
answer in a variable no longer texts the user. That action also gets the
channel selector the send-message actions already had — it had none at all.

**Changes**

- `server/utils/constants.js`: `MESSAGE_GLADYS_ONLY_SERVICE` sentinel.
- `server/lib/message/message.sendToUser.js`: skip `forwardToChannels` for the sentinel.
- `server/lib/message/message.reply.js`: apply the three cases to a `source: 'AI'` answer.
- `server/lib/scene/scene.actions.js`: `AI.ASK` carries `service` into its message (read locally by `reply`, never sent to the chat API).
- `front`: new "Gladys conversation only" option in `MessageServiceSelector`, selector added to `AskAI.jsx`, fr/en/de strings.
- `docs/specs/external-integrations/capabilities/communication-type.md`: the delivery choice documented in the same diff, as the living spec requires.

### Related request

Forum: https://community.gladysassistant.com/t/impossible-denvoyer-des-messages-a-gladys/10851

### Checklist

- [x] If a forum topic or GitHub issue exists, the description links it
- [x] Tests pass: `cd server && npm run coverage` — 6667 passing; the 11 failures are pre-existing in this dev container (verified identical on the base commit) and come from the sandbox, not the diff. Patch coverage is 100% on the changed lines. Cypress could not run locally (binary download blocked in the container), it runs in CI
- [x] Linter and prettier pass on both front and server, plus `compare-translations` and the front `npm run build`
- [x] No undocumented breaking change — the default (no `service`) keeps broadcasting

## Summary by CodeRabbit

* **New Features**
  * Scene messages can now be delivered to all configured services, one selected service, or only the Gladys conversation.
  * Added service selection for AI-generated replies.
  * Messages remain visible in the Gladys conversation regardless of delivery selection.
  * Added German, English, and French labels explaining these delivery options.
  * Added support for configuring message delivery through scene integrations.


Pierre-Gilles:
Merged in https://github.com/GladysAssistant/Gladys/pull/3118

